PR #1, opened 2015-04-08
gluster: fix tcmu_get_attribute error check

ddiss:
This fixes the wce tcmu_get_attribute() error check, which currently has
a cut-n-paste bug.


agrover:
thanks! :+1:

PR #2, opened 2015-06-04
Fix compiler warning and remove generated files on 'make clean'

tasleson:
A number of different ways to fix this.  Some may object with the use of a macro.  If not liking swizzle it another way, but this is currently a bug.


agrover:
thanks! :+1:

PR #8, opened 2015-10-02
QEMU QCoW version 1 handler

cleech:
Adapted from the QEMU code.  Handles qcow version 1 with compression
(read only, copy-on-write expanded) and read-only backing images in
either qcow or raw format.  Encryption is intentionally not supported,
as it's a flawed design that has been deprecated in QEMU.

My plans are to extend this to qcow2 next, but qcow was a good small
module to test porting from QEMU to tcmu-runner.

Signed-off-by: Chris Leech <cleech@redhat.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,16 @@ set_target_properties(handler_glfs
   PREFIX ""
   )
 
+# Stuff for building the qcow handler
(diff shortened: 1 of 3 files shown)


agrover:
woohoo thanks! :+1:

PR #9, opened 2015-10-12
Corrected the "handler" path

os12:
- 64-bit Debian systems just have `/usr/lib/` and so the runner binary should look there
 - added minimal platform detection


agrover:
Much thanks!! :smiley_cat:

os12:
Thanks for the quick review, Andy!

PR #10, opened 2015-10-13
Added a missing CMake file

os12:
- ba41c3eeba583ce6fa was incomplete
 - these *.cmake files are ignored for some reason


agrover:
I've redone CMakeLists.txt a bit so this isn't needed anymore, but thanks.

Issue #12, opened 2015-10-14
rfe: async handler API


agrover:
We should enable those handlers that wish to, to complete requests out of order.

os12:
Right, this needed for serious embedding. That is, I would like to take the request processing code from tcmu-runner (see #13), link it to an external application (as it manages its own threads) and then process the requests in an async (ie non-blocking) way. Note, the notification itself should work with `epoll()` today, yet the request processing is still blocking.

Looking at `tcmu_handle_device_events()`/`handle_one_command()`, the plugin should be able to complete/fail a request as well as mark it as "pending". It should then be possible to complete a pending request (possibly out of order).

I can take care of the user-mode work, including any and all API/embedding extensions and issues. Yet, this feature would require kernel support as the behavior of the circular queue would change. Also, it would be good to map the individual iovecs for each request, rather than copy the data back and forth.

os12:
Hi Andy, the recent work from @mkhon submitted in #21 is for this issue. Maxim extended the Runner's "file" handler with async options, but left the "consumer" example untouched. I've just had a look - it should be trivial to correct/extend. One of us will look into that.

Could you clarify the handler binding rules please:
  1. Can multiple consumer-like processes bind to distinct TCMU devices?
  1. What is the best way to implement device selection within each process? It looks like I must implement the `added` handler to accept/skip based on the value of `tcmu_get_dev_cfgstring(dev)`.

Thanks!

agrover:
Hi Oleg,

1. the idea was that different, unrelated, processes could bind to distinct TCMU devices, as differentiated by part of the cfgstring called "subtype". 
1. each process should only get added() for the subtype it registered as supporting, so it should not need to skip or accept. It will likely need the remainder of the cfgstring to know how to configure itself to handle that particular device.

os12:
Right, I can easily put a discriminant into subtype. Thanks! (I plan to 
have multiple devices of a similar type handled by distinct processes 
for stability reasons. So, the daemons need to choose the device.)

os12:
Update:
  * The code landed in #26 (from @mkhon)
  * There is a notification issue between the kernel- and user-mode pieces. A PR is coming...
  * The kernel-mode code needs quite a bit of work:
    * The first cut of the implementation is here: https://github.com/rancher/linux/tree/v4.2-tcmu-async
    * There are issues with ring sizes, we are cleaning that up

PR #13, opened 2015-10-15
Factored out SCSI request processing

os12:
Today, tcmu-runner owns the worker threads, thus forcing the users' code into the
plugin structure. This certainly works for minimal examples, but prevents any kind
of embedding into larger 3rd-party applications.

Production-grade applications generally have their own threading structure and want
to deal with socket/file/tcmu notifications in an async way. As such, I just need a
minimal library that handles device opening and command parsing.

There are several API sets now:
 - the API each plugin must implement (unchanged, see tcmu_handler)
 - the API available to the plugin (api.c)
 - the API the host process must implement (debug functions today)
 - the minimal API tcmu_lib will provide (a new concept, builds on api.c)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 tcmu-runner
+consumer
 *.o
 *.a
 *.so
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 tcmu-runner
 *.o
+*.a
 *.so
 *generated.[ch]
 Makefile
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,12 @@ set_target_properties(handler_file
   PREFIX ""
   )
 
+# The minimal library consumer
+add_executable(consumer
+  consumer.c
+  )
+target_link_libraries(consumer tcmu)
+
 # Stuff for building the glfs handler
 add_library(handler_glfs
   SHARED
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,11 +31,15 @@ find_library(KMOD kmod)
 find_library(GFAPI gfapi)
 
 # Stuff for building the main binary
+add_library(tcmu
+  api.c
+  libtcmu.c
+  )
 add_executable(tcmu-runner
   main.c
-  api.c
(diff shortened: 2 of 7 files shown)


os12:
Hi Andy, thank you for the quick review!

I've corrected the issues you have pointed out to date, rewritten the patches and pushed again.

os12:
BTW, here are a few notes from the PR's first generation, for the record. The rest of the comments have been addressed and patches have been rewritten.

### Static/shared library

Indeed, shipping a `.so` file is possible.

However, the large C++ applications I have worked with build their entire world of dependencies from scratch to avoid API/ABI issues (imagine mixing compiler versions that ship different STL versions). As such, producing a static lib is perfectly fine and simplifies binary distribution a little. (Also, AFAIK, the C ABI on Linux is pretty stable anyway, so building out of the tree should work too).

Case in point, I've added tcmu-runner as a git module and build the static lib together with the hosting app.

### Logging support

The hosting app must implement the following API:

    void dbgp(const char *fmt, ...);
    void errp(const char *fmt, ...);

Consider a bigger application with its own log system: log files, rotation, levels, paths, redirection etc. Suppressing or dumping information to stderr/stdout is not useful in such a system.

Specifically, in my system, these C-style calls are routed to the main C++-style log subsystem with an appropriate path and severity.

agrover:
Ok, I guess as long as both static and shared are possible there's no issue.

About dbgp/errp, I see your point, I just don't like the the main application having to export specific symbols. But I think this is fine for now and we can merge this and explore different possibilities with this as a base.

Thanks for your work! :+1:

os12:
Great! Thanks, Andy!

P.S. the C-style debug function pointers can be passed into the `init_tcmu_lib()` function. We'll need to look into the API extensions as part of #12.

agrover:
Cool, that's what I was thinking.

I'm going to add you as a collaborator. Please continue to use PRs for significant changes, and I will start doing the same.

os12:
Cool, thanks!

BTW, opening PRs is the nicest way irrespective of the rights, as there 
a nice web trail (with discussion) and people who watch the repo get 
notified of changes.

PR #16, opened 2015-11-24
Shared library support

agrover:
The goal for this branch is to flip things around so TCMU can be used as a library by existing applications such as QEMU or others that have their own event loop. This will require that libtcmu give the client the file descriptors to wait on. This allows the client to have control over the whole spectrum of parallelization options, from every device being handled by a single thread, to each device being given its own thread, or something in between.

libtcmu would handle parsing netlink and also registering as a handler with tcmu-runner over DBus, as well as handling the low-level parts of the TCMU udev interface and supplying basic SCSI command emulation routines that clients can use if desired.

EDIT: libtcmu currently does no DBus (it's all in tcmu-runner). In the future it will make some DBus calls but will still otherwise function if DBus is not present (such as in a containerized environment.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@ tcmu-runner
 consumer
 *.o
 *.a
-*.so
+handler_*.so
+libtcmu.so*
 *generated.[ch]
 Makefile
 version.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,31 +30,51 @@ find_library(DL dl)
 find_library(KMOD kmod)
 find_library(GFAPI gfapi)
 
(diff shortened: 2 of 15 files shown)


os12:
Very cool, Andy!

What about the dependencies? Do things still work without DBus, when devices are configured manually?

agrover:
Thanks, Oleg. Dependencies at the moment just are libnl3 for netlink. I do want DBus support not only in tcmu-runner but also libtcmu. The two C DBus options right now are Glib's GDBus and systemd's sd-bus, if one of those is an easier dependency for you to accept than the other. If not, we at least can have a conditional build flag to exclude DBus support.

PR #21, opened 2015-12-13
Async io API proposal

mkhon:
This is the proposal for API change to support async handlers.
- async handler is supposed to always return TCMU_ASYNC_HANDLED
- async handler is supposed to use tcmulib_async_command_init() to copy the command
upon command completion
- async handler is supposed to use tcmulib_async_command_complete()

Some more comments:
- Probably tcmulib_async_command_init/tcmulib_async_command_init() should be moved to libtcmu_common.h as it seems to be the only header containing the API for handlers
- It looks like you tried to avoid using structs in tcmu-runner.h API (e.g. in handle_cmd()) but using structs allows to extend API without breaking both API (on the source level) and ABI so that compiled handlers are backward compatible, i.e. can be used with tcmu-runner of a newer API version (if this was important at all). For example I needed to pass cmd_id down to all the handler invocation so that it can be passed back when the command is completed
- The use of 'struct tcmulib_cmd' can be pushed down to the other libtcmu_common.h functions (like tcmu_emulate_* functions) but I wanted to have the minimal patch for the start. This can be done later.

### DIFF
--- a/api.c
+++ b/api.c
@@ -159,29 +159,41 @@ long long tcmu_get_device_size(struct tcmu_device *dev)
 	return size;
 }
 
-static inline int get_cdb_length(uint8_t *cdb)
+#define VARIABLE_LENGTH_CMD   0x7f
+
+/* defined in T10 SCSI Primary Commands-2 (SPC2) */
+struct scsi_varlen_cdb_hdr {
+	uint8_t opcode;         /* opcode always == VARIABLE_LENGTH_CMD */
+	uint8_t control;
+	uint8_t misc[5];
+	uint8_t additional_cdb_length;  /* total cdb length - 8 */
+	uint16_t service_action;
+	/* service specific data follows */
+};
+
+/* Command group 3 is reserved and should never be used.  */
+static const unsigned char scsi_command_size_tbl[8] =
+{
+	 6, 10, 10, 12,
+	 16, 12, 10, 10
+};
+
+#define COMMAND_SIZE(opcode) scsi_command_size_tbl[((opcode) >> 5) & 7]
+
+int tcmu_get_cdb_length(uint8_t *cdb)
 {
 	uint8_t opcode = cdb[0];
 
-	// See spc-4 4.2.5.1 operation code
-	//
-	if (opcode <= 0x1f)
-		return 6;
-	else if (opcode <= 0x5f)
-		return 10;
-	else if (opcode >= 0x80 && opcode <= 0x9f)
-		return 16;
-	else if (opcode >= 0xa0 && opcode <= 0xbf)
-		return 12;
-	else
-		return -EINVAL;
+	return (opcode == VARIABLE_LENGTH_CMD) ?
+		((struct scsi_varlen_cdb_hdr *) cdb)->additional_cdb_length + 8 :
+		COMMAND_SIZE(opcode);
 }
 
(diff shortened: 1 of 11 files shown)


agrover:
Hi @mkhon. Looks really good. Let's get this in!

One last thing -- I think we already have some scsi cmd size stuff over in api.c, could that be shared, if it was improved to handle variable-length SCSI commands?

os12:
Hi Andy, have you looked at the async *read* path? 

Here is my understanding of the sequence:
 - `tcmulib_async_command_init()` copies `iovec` structures and the caller can push the copied command to a worker thread (and keep it pending)
 - the copied command has an array of the original `iovec` structures (that still point to the command ring, I believe)
 - how do we ensure that the corresponding ring slot remains valid?
 - would out-of-order completions mess things up?

Could you clarify that please? It seems that the API should let the caller allocate a command of a certain size, then fill it in and the complete... What am I missing?

Thanks!

agrover:
Even for reading, we can still provide a buffer to the handler, because we know the read cmd's data length. If the caller allocated memory and filled it in with read data, then tcmulib would still have to copy from that buffer into the data area, so the kernel could access it. Using commands like readv() it is possible to read directly into the spot where TCMU needs it without an intermediate copy (well, without *another* copy).

The way I envisioned OoO working was we would use the next cmd_entry in the ring (which originally was for a different command), change cmd_id and fill it in cmd.rsp, and return that -- and then when the first command completed we would use the OoO cmd'd cmd_entry. Every entry on the cmd ring is large enough for completion status for any command, not just the original requesting cmd.

But one issue that I think you're touching on is that if the first command is still using the in-ring iovec[], then we can't take it to return the OoO cmd's status. I think if supporting async we need to make a copy of everything in the cmd.req (the iovec[] and cdb) for all commands, and then the first entry is free to be used to return the OoO cmd's status. I think the PR currently just copies these bits for expecting-to-be-async cmds, which is not enough.

mkhon:
Oleg,

iovec points to the data ring, not command ring, so it can be copied as is.

Andy,

Can you clarify what exactly is not enough? Currently command ring only contains (in addition to the req):
- cdb (which is deep-copied)
- iovec (which points to the data ring, so we need to copy only iovec and not the data)

Max

agrover:
If every single command calls async_command_init (or none do) then we're fine. If only some do, then that's bad. 

Maybe we should just have get_next_command do the malloc + copy of the cdb and iovec[] that async_command_init is doing?

mkhon:
Andy, yes, currently the assumption is that the handler (if it is going to operate in async mode) should always use async_command_init().

Yes, I can modify the API to always copy and always free (on completion) - just thought that additional malloc/copy is not what you want when you do non-async processing. But I agree that this would make the API more safe (and simple). I'll do that.

mkhon:
Andy, Oleg,

I did all the suggested changes. Please note that async handler now needs to call tcmulib_processing_complete().

mkhon:
Andy,

I haven't noticed that tcmu-runner is Apache-licensed. I returned back old CDB length getter (and added variable command length support on top of it).

agrover:
OK great. Please rebase to a clean set of commits and I will pull it in.

os12:
I'll open a rebased PR in a minute...

mkhon:
I rebased and submitted new pull request #26

PR #23, opened 2015-12-16
Corrected debug-level printing

os12:
- there was a clash between errp() functions
   + these overloads are legal in C++
   + yet might be considered ODR in C (assuming the language uses the term)
 - the printing function must be stashed away
   + and passed in


agrover:
great, thanks!

PR #25, opened 2016-01-05
Async API

os12:
This brings async work from @mkhon originally submitted in  #21. This wraps up #12.


mkhon:
I haven't noticed this. Please use #26 instead - it has more proper commit message and proper attribution.

PR #26, opened 2016-01-05
Async IO API proposal:

mkhon:
- the handler can now return TCMU_ASYNC_HANDLED
- tcmulib_get_next_command() now always copies iovec and cdb from the command ring
to the tcmulib_cmd. The returned command can be safely used in async handler.
- rename get_cdb_length() to tcmu_get_cdb_length() and add support for variable-length CDB.


agrover:
Much thanks, @mkhon!

PR #27, opened 2016-01-10
Fix a race condition result in userspace blocking to get any command

os12:
There is a window between finish the loop search for next command
(tcmulib_get_next_command() and mark the command as received (read uio fd in
tcmulib_processing_complete). If there are commands sent from kernel space
during this time window, userspace would already ack it but unable to pick it
up, which result in hang of userspace process, since polling the uio fd would be
blocked until (if lucky) next command arrives.

This patch introduce tcmulib_processing_start(), which would ack the receiving
of command before tcmulib_get_next_command(), so tcmulib_get_next_command()
wouldn't miss any command from kernel space. Though it may process more commands
than acked, it won't be a problem since we would just let loop runs one more
time to ack them.

The patch is from @yasker. See #12.

### DIFF
--- a/consumer.c
+++ b/consumer.c
@@ -212,6 +212,7 @@ int main(int argc, char **argv)
 				struct tcmulib_cmd *cmd;
 				struct tcmu_device *dev = tcmu_dev_array[i];
 
+				tcmulib_processing_start(dev);
 				while ((cmd = tcmulib_get_next_command(dev)) != NULL) {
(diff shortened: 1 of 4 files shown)


yasker:
Hi @os12 

This patch is for another branch and I haven't tested it against master. I am planning to open a PR on Monday, after I test it.

os12:
OK, the very same patch is resubmitted in #30 by the original author.

PR #28, opened 2016-01-11
Sample async file handler

mkhon:
- Sample async file handler. Build with -DASYNC_FILE_HANDLER to build async file handler.
- Use pread/pwrite instead of lseek+read/write.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,18 @@ set_target_properties(handler_file
   PREFIX ""
   )
 
+# Stuff for building the async file handler
+add_library(handler_file_async
+  SHARED
+  file_example.c
+  )
+set_target_properties(handler_file_async
+  PROPERTIES
+  PREFIX ""
+  COMPILE_FLAGS "-DASYNC_FILE_HANDLER"
+  )
+target_link_libraries(handler_file_async ${PTHREAD})
+
 # The minimal library consumer
 add_executable(consumer
   consumer.c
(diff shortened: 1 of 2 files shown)


mkhon:
I submitted separate pull request for pread/pwrite (see #29). Will merge/rebase it once it gets in.

agrover:
it might be clearer if we have the async file handler example be separate from the sync file handler example, what do you think?

os12:
How about building two `.so` files from the same source? Both are file handlers, both conform to the Runner's API...

P.S. is this even possible with CMake?

agrover:
I was really focusing on file_example.c as a code sample for handler writers rather than something that anyone would ever use. For actual use, the in-kernel fileio backstore should be strictly better. An async example would also be great, and I was thinking easier to understand if a developer could compare the async vs sync implementations side-by-side. The only issue would be we'd need to keep both of them up to date if other API changes happened.

mkhon:
Andy,

This code change does not have any practical use except testing async handler code paths and was not intended as a replacement for in-kernel file backend in any way. There's a lot of code shared between sync and non-async so I would suggest either having an ifdef (like now) or split them into common source file and two handler stubs. What do you think?

agrover:
ok, ifdefs are fine.

mkhon:
I updated the pull request with the latest code and changes as discussed above.

agrover:
Thanks!

PR #31, opened 2016-01-25
qcow2 image file support

cleech:
Basic qcow2 support, handles the refcount tables.  Also fixes up some issues when request iovecs are not being sector aligned.  Might be some issues to work out when faced with snapshots still.


agrover:
Great! I think people will really like this and find it very useful!

PR #36, opened 2016-03-15
libtcmu.h: Don't include scsi_defs.h

famz:
A couple fixes to let me correctly link to libtcmu with the following dummy application (configure detection snippet):

#include <stdio.h>
#include <libtcmu.h>

int main(int argc, char **argv)
{
  struct tcmulib_context *ctx = tcmulib_initialize(NULL, 0, NULL);
  return ctx != NULL;
}


famz:
Thanks for the quick merge, Andy!

PR #38, opened 2016-03-16
tcmu-runner: Make g_runner_handlers an array of pointers

famz:
on_check_config receives a user_data pointer, which points to some
element into g_runner_handlers, the array of struct tcmur_handler. This
is not a problem now because the array never resizes (reallocates) once
built, but that is a very fragile assumption and is prune to memory
corruption in the far future.

Instead, save the pointers. All existing callers of
tcmur_register_handler are passing addresses of a static global
variable, so there is no use-after-free risk.  Future callers must be
more careful of the life cycle of handler because they are no longer
copied by value.

### DIFF
--- a/main.c
+++ b/main.c
@@ -50,7 +50,7 @@
 static char *handler_path = DEFAULT_HANDLER_PATH;
 static bool debug = false;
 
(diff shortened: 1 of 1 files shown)


agrover:
yes, rather fragile right now. Thanks!

PR #41, opened 2016-03-20
build: use libkmod pkgconfig and add with-glfs and with-qcow parameters

ddiss:


agrover:
Looks good, thanks! :+1:

agrover:
Maybe also document new build parameters in the readme?

ddiss:
Thanks for the merge. Sure, I'll work on a readme patch to document the new params.

PR #42, opened 2016-03-21
README: document cmake with-glfs and with-qcow params

ddiss:


agrover:
thanks again!

PR #44, opened 2016-03-22
cmake: Optional build dependency on glusterfs

famz:
Don't build glfs handler if gluster api headers are not found. Print a
warning in cmake.


famz:
Dup with pull #42

Issue #15, opened 2015-10-20
Fix DBus for multiple independent handlers


agrover:
Now that we're going down the libtcmu path for supporting handlers other than tcmu-runner, if we're going to present a coherent picture of the available handlers to the user then they need to cooperatively register in one place. We can use Telepathy "mission-control-5" as a model for how to accomplish this.

os12:
Hey Andy, I think the embedded cases (ie large applications linking `libtcmu.a`) are orthogonal to the tcmu-runner's system/dbus registration and interfaces. That is a system would have the tcmu-runner with its plugins or a larger application with libtcmu linked.

Here is how I see the embedded case:
  * the application wants to reuse SCSI command processing and the user/kernel protocol implementation
  * the device registration process is tricky/awkward/proprietary 
  * the application would have little interest in responding DBus messages 
    * it would want to minimize the dependencies

agrover:
Yes it's not relevant to the embedded case. It's relevant to the distro use case, and I think we can accommodate both.

famz:
Does it make sense to handle the dbus communications in libtcmu?

agrover:
I think so.. it's just a matter of figuring out how to do it, given that multiple processes could be using libtcmu, and only one entity can back a DBus service. I looked into how Telepathy does this, but got lost in all the code.

famz:
I think we can extent tcmu-runner's org.kernel.TCMUService1 interface and let it be the mid-man coordinator between multiple handlers and targetcli.

Currently it only implements a CheckConfig method to verify the cfgstring, and provides object introspection for enumerating available subtypes, for targetcli. If we extent this service for libtcmu users as well, targetcli user interface is extended automatically. Specifically, we can add a "RegisterHandler" method (family) alongside CheckConfig, which accepts subtype string from libtcmu.

One more related feature I'd like to discuss is file descriptor passing.

In a typical virtualization setup, QEMU doesn't open files by itself, instead it gets FDs passed from libvirt, which opens image files on its behalf, via unix socket control messages. This is so because of privilege limitation for security concerns. In TCMU, QEMU process would need to open /sys/class/uio and /sys/device/tcm_user, which is not desired. A solution is let tcmu-runner who gets DBus messages do this, and pass it back to libtcmu through a unix socket. I think this is doable.

[...]

Issue #43, opened 2016-03-21
Change libtcmu to support multiple fds behind a single fd to client


agrover:
The solution to #15 is going to involve each libtcmu instance exporting DBus bus names. Therefore we need to support servicing a DBus API in addition to netlink. This is going to need two fds, but we only want the client to have to poll on one master fd for us. We can hide this implementation detail by giving the client an epoll fd that will signal if any of libtcmu's fds in the epoll interest list need servicing.

famz:
Can we count on the libtcmu client to use glib main loop (or handle glib fds equally)? That is the easiest solution that doesn't require particular catering in libtcmu.

FWIW this is already true for QEMU's main loop.

agrover:
No I don't think we can require clients to use glib main loop. libtcmu will be using DBus and needs glib for gdbus but this is an implementation detail, not part of the libtcmu API.

I'm not sure what you mean by "handle glib fds equally"?

famz:
OK. The epoll fd patch looks good to me! I'll try to integrate with my DBus branch.

"handle glib fds equally" means a custom event loop can call g_main_context_query() and handle the returned fds, without calling g_main_loop_run(), to integrate with glib. That is the current way in QEMU.

PR #46, opened 2016-03-22
RFC: DBus service in libtcmu as a prototype for issue #15

famz:
Andy, could you please take a look at the approach in this branch and see if you agree with the idea?

* tcmu-runner exports a "RegisterHandler" method in a new interface in a new "handler manager" object.

* libtcmu calls the method to register themselves with supported subtype, at the same time it owns a bus_name [1] that is derived from subtype, in the TCMUService1 interface.

* tcmu-runner connects to the subtype's bus_name to validate it, then add the subtype to the list.

* When targetcli queries, tcmu-runner will return all the registered subtypes and cfgstring descriptions.

* When targetcli calls CheckConfig, tcmu-runner will proxy it to the libtcmu's DBus service [1], and return the value to targetcli.

Note: above targetcli integration is lightly tested. Handling requests with the new libtcmu is not. Apologies for the cluttering printf's in the commits.


agrover:
Yes this looks good, except as I mentioned in #43 we need to *not* rely on glib in the client. (i.e. look at consumer.c and get it working like synthesizer.c but no (direct) linking to glib allowed). Can we use g_io_channel_unix_get_fd() to get something we can stick in an epoll set, give the epoll fd to the client to poll on, and then call a glib function to process dbus events when the fd becomes ready?

Here's a changeset that starts using epoll: https://github.com/agrover/tcmu-runner/tree/dev-epoll . We would just need to further modify the code to add the dbus fd to the epoll set and call the right thing in tcmulib_master_fd_ready().

Thoughts on this? Can we use gdbus without the glib eventloop?

But overall yes I think the significant work you have here is the right direction. I'm very excited by this progress!

agrover:
BTW FWIW systemd's sd-bus library (http://0pointer.de/blog/the-new-sd-bus-api-of-systemd.html) can be used this way, using sd_bus_get_fd() and sd_bus_process(), so that's a last-choice fallback alternative.

famz:
Yes we can. libtcmu should manually build the GDBusConnection for bus name watching and method calls, out of a floating GIOStream that is not managed by glib main context, and then attach it to epollfd. I'll look into it.

famz:
UPDATE: Bummer.. So I read the glib source code to figure out what is needed in our custom event loop, to drive the asynchronous g_bus_watch_name and g_bus_own_name API families (including their _*on_connection variants). The news is not so good: there is no easy way to find which fds we should to poll, nor are there any "fd ready handlers" in glib we can call to let it make progress.

That is because gdbus uses a complex thread pool Worker & Task queue that is tightly forged together with the glib event loop framework, using things like idle GSource to notify completion. This means we have to call g_main_context_check() and g_main_context_dispatch() to make progress, instead of FDs. The IO stream is actually handled in a worker thread.

So the only way I can think of  is creating our own DBus communication thread and call g_main_context_push_thread_default() there. IIUC, this should be enough to move everything about g_dbus_* to that thrread, where we run a dedicated glib event loop. Then we will need synchronization between the thread where the tcmulib context is created and handler callbacks are called. IMO that will be a major compromise of simplicity that we should think twice between going there.

agrover:
Ok, rats. In that case we may need to switch to sd-bus.

agrover:
Are you still working on this or would it be safe for me to try converting your gdbus implementation to sd-bus?

famz:
I haven't worked on sd-bus. Please go ahead!

famz:
Andy, any update on the sd-bus? Is there anything I can help?

agrover:
Sorry. I'm not able to work on this until late next week, Friday Apr 22. Is this blocking other progress you'd like to make?

famz:
No, just think libtcmu in this branch can be independently linked to sd-bus, which can be worked on separately from tcmu-runner. So maybe I can update my branch to do it. And later we can look at switching the daemon side. What do you thinkl

agrover:
I've been working on switching to sd-bus and it doesn't look quite as capable as I'd thought. I think right now we may want to accept using DBus-enabled libtcmu requires client glib event loop. Right now potential libtcmu users either don't want DBus integration at all, or are glib-eventloop based, so converting to sd-bus doesn't actually gain us much. It's still an option to add sd-bus support later, if it is needed.

@famz do you mind updating this PR to resolve any conflicts and any other issues so we can proceed with reviewing it for merging?

famz:
Yes, I can get back to work on this PR.

@agrover Could you elaborate the issues with sd-bus?

agrover:
Yeah, I think it just needs to mature a bit, get some more API docs and examples. The only way to really figure out how to use it at this point is to study its implementation in systemd git, and I honestly just got tired of doing that and lost motivation since sd-bus support was a self-generated requirement, not from a potential or current libtcmu user.

So this isn't a "no" just a "not yet" for enabling libtcmu to use sd-bus. If somebody needs it (or posts a PR) then that will change things. So I would say let's proceed with gdbus-based support for now, since that's something QEMU can certainly use.

famz:
Andy, I've rebased and cleaned up the branch. Please have a look and let me know if there are issues.

famz:
ping?

tcmu-synthesizer.c
```diff
@@ -0,0 +1,250 @@
...
+	case WRITE_10:
+	case WRITE_12:
+	case WRITE_16:
+		return SAM_STAT_GOOD;
+
+	default:
+		syn_debug("unknown command %x\n", cdb[0]);
+		return TCMU_NOT_HANDLED;
+	}
+}
```
rootfs: wonder why compiler doesn't warn function has no return

tcmu-synthesizer.c
```diff
@@ -0,0 +1,250 @@
...
+
+static int syn_handle_cmd(struct tcmu_device *dev, uint8_t *cdb,
+			  struct iovec *iovec, size_t iov_cnt,
+			  uint8_t *sense)
+{
+	uint8_t cmd;
+
+	cmd = cdb[0];
+	syn_debug("syn handle cmd %d\n", cmd);
```
rootfs: `%d` -> `%x`, to be consistent

agrover:
sorry @famz! Taking a look now. Thanks also @rootfs.

tcmu-synthesizer.c
famz: Doesn't **each** branch have a return? I think that's accepted by the compiler.

tcmu-synthesizer.c
famz: Sounds good!

tcmu-synthesizer.c
rootfs: right, that works

agrover:
Thanks again @famz, and very sorry again about the delay! Notifications fixed.

PR #47, opened 2016-03-23
tcmu-runner: Simplify on_check_config DBus return value

famz:
The commit is extracted from the RFC of yesterday. It is independent so we can just merge it.


agrover:
Thanks!

Issue #62, opened 2016-08-20
Build fails if cmake is run outside of the source directory


ddiss:
When invoked from a new subdirectory within the source tree (e.g. build/), cmake writes the autogenerated files (tcmuhandler-generated.c, tcmuhandler-generated.h and version.h) into the current directory. The subsequent build then fails, because the autogenerated files aren't located during compilation:

<pre>
tcmu-runner (master)> mkdir -p build
tcmu-runner (master)> cd !$
cd build
tcmu-runner/build (master)> ls
tcmu-runner/build (master)> cmake -Dwith-glfs=false ../
-- The C compiler identification is GNU 4.8.5
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.28") 
-- checking for module 'gio-unix-2.0'
--   found gio-unix-2.0, version 2.44.1
-- checking for module 'libkmod'
--   found libkmod, version 17
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.8") 
-- Configuring done
-- Generating done
-- Build files have been written to: build
tcmu-runner/build (master)> make
[  3%] Generating tcmuhandler-generated.c, tcmuhandler-generated.h
Scanning dependencies of target tcmu
[  7%] Building C object CMakeFiles/tcmu.dir/api.c.o
[ 11%] Building C object CMakeFiles/tcmu.dir/libtcmu.c.o
[ 15%] Building C object CMakeFiles/tcmu.dir/libtcmu-register.c.o
libtcmu-register.c:24:35: fatal error: tcmuhandler-generated.h: No such file or directory
 #include "tcmuhandler-generated.h"
                                   ^
compilation terminated.
CMakeFiles/tcmu.dir/build.make:117: recipe for target 'CMakeFiles/tcmu.dir/libtcmu-register.c.o' failed
make[2]: *** [CMakeFiles/tcmu.dir/libtcmu-register.c.o] Error 1
CMakeFiles/Makefile2:215: recipe for target 'CMakeFiles/tcmu.dir/all' failed
make[1]: *** [CMakeFiles/tcmu.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
</pre>

[...]

PR #63, opened 2016-08-20
build: fix builds done outside of source directory

ddiss:
cmake currently writes tcmuhandler-generated.* and version.h into the
build directory, which sees compilation fail if cmake is invoked from
outside the source directory (build_dir != source_dir).

Bug: https://github.com/open-iscsi/tcmu-runner/issues/62


agrover:
thanks! :+1:

PR #65, opened 2016-08-25
build: fix tcmu-runner_HANDLER_PATH

pkalever:
Currently, tcmu-runner fails with below error
$ ./tcmu-runner
couldn't open handlers

This is because the 'handler_path' is DEFAULT_HANDLER_PATH, which is
\#define DEFAULT_HANDLER_PATH "@tcmu-runner_HANDLER_PATH@"
which inturn set to '${CMAKE_INSTALL_LIBDIR}/tcmu-runner' this on
expanding gives 'lib64/tcmu-runner' excluding the CMAKE_INSTALL_PREFIX

As a result the tcmu-runner binary can't locate the handlers unless an explicit
--handler-path parameter is specified.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -std=c99")
 
 include(GNUInstallDirs)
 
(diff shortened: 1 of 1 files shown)


pkalever:
please-review: @agrover @vbellur @ddiss @famz

PR #66, opened 2016-08-31
build: fix RPATH settings for shared libraries

pkalever:
Currently,
$ tcmu-runner
tcmu-runner: error while loading shared libraries: libtcmu.so.1: cannot
open shared object file: No such file or directory

This patch set CMAKE_INSTALL_RPATH to required LIBRARY PATH,
as a result there will be no issues while loading shared libraries.

Signed-off-by: Prasanna Kumar Kalever prasanna.kalever@redhat.com

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ set_target_properties(tcmu
   PROPERTIES
   SOVERSION "1"
   )
+SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
 target_include_directories(tcmu
   PUBLIC ${LIBNL_INCLUDE_DIR}
   PUBLIC ${GLIB_INCLUDE_DIRS}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,8 +194,8 @@ install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
 install(FILES org.kernel.TCMUService1.service
   DESTINATION /usr/share/dbus-1/system-services)
 install(FILES tcmu-runner.conf DESTINATION /etc/dbus-1/system.d)
-if (SUPPORT_SYSTEMD)
+if (IS_DIRECTORY /usr/lib/systemd/system)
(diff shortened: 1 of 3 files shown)


pkalever:
please-review: @agrover @vbellur @famz

agrover:
:+1:

CMakeLists.txt
```diff
@@ -194,8 +194,8 @@ install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
 install(FILES org.kernel.TCMUService1.service
   DESTINATION /usr/share/dbus-1/system-services)
 install(FILES tcmu-runner.conf DESTINATION /etc/dbus-1/system.d)
-if (SUPPORT_SYSTEMD)
```
agrover: this is because you can invoke cmake as `cmake -DSUPPORT_SYSTEMD-ON` if you want systemd support. I'm not sure we can assume this from whether systemd is installed or not.

org.kernel.TCMUService1.service
```diff
@@ -3,6 +3,6 @@
 
 [D-BUS Service]
 Name=org.kernel.TCMUService1
-Exec=/usr/bin/tcmu-runner
+Exec=/usr/local/bin/tcmu-runner
```
agrover: why would we want to make this change?

agrover:
the second changeset has its own PR, did it get pushed to this one accidentally maybe?

org.kernel.TCMUService1.service
pkalever: do you mean changes in  tcmu-runner.service file?
Because make install do install tcmu-runner binary on /usr/local/bin/ path

Is that what you are asking?

pkalever:
Hey Yeah..
This is done by mistake, I use a script rfc.sh for pushing commits automatically (fetches/pulls the remote and push resolving conflicts)
I need to check with my local scripts, sorry about that. (May be it had push HEAD^^)

agrover:
Is this PR still waiting on review? thanks.

pkalever:
Please feel free to close them.

I shall send them again with a separate PR if needed.
As any way I cannot map to this PR, after I corrupted my fork :) 

Thanks!

PR #67, opened 2016-09-05
Don't call handler->check_config if NULL

famz:
The documentation of tcmulib_handler.check_config says "this function is
optional", handlers don't implement this should be handled gracefully.


agrover:
good point, thanks.

PR #68, opened 2016-09-08
build: Fix loading tcmu-runner unit using systemd

pkalever:
Currenlty,
$ systemctl start tcmu-runner.service
Failed to start tcmu-runner.service: Unit tcmu-runner.service not
found.

This is because in CMakeLists.txt 'SUPPORT_SYSTEMD' is not defined, hence
tcmu-runner.service is not installed

This patch also fixes the tcmu-runner binary path (installed from source)
in systemd and dbus service files

Signed-off-by: Prasanna Kumar Kalever prasanna.kalever@redhat.com


pkalever:
please-review: @agrover @vbellur

agrover:
👎 don't want to install into /usr/local. Don't want to make compiling for systemd dependent on if the current system supports systemd.

PR #70, opened 2016-11-02
build: fix failure on ubuntu

liuy:
On my ubuntu 14.04 I got following error on the fresh building:

/home/yliu/tcmu-runner/qcow.c: In function ‘qcow2_block_alloc’:
/home/yliu/tcmu-runner/qcow.c:1008:27: error: ‘FALLOC_FL_ZERO_RANGE’ undeclared (first use in this function)
    ret = fallocate(s->fd, FALLOC_FL_ZERO_RANGE, cluster, s->cluster_size);
                           ^
/home/yliu/tcmu-runner/qcow.c:1008:27: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [CMakeFiles/handler_qcow.dir/qcow.c.o] Error 1
make[1]: *** [CMakeFiles/handler_qcow.dir/all] Error 2

This is because we don't pass on HAVE_LINUX_FALLOC, which can be automatically
detected by cmake module function, to compiler.

So the reasonable fix is to add the include-file-check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ project (tcmu-runner VERSION 1.1.3 LANGUAGES C)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -std=c99")
 
 include(GNUInstallDirs)
+include(CheckIncludeFile)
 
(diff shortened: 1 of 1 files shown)


agrover:
much thanks! :+1:

PR #74, opened 2016-11-16
build: add cscope support

liuy:
This will create a new target in the Makefile, allowing devvers to enjoy cscope
utility just by typing 'make cscope'.

'make clean' will also clean up the intermediate cscope files.


agrover:
thanks!

PR #78, opened 2016-11-25
Introduce generic_handle_cmd() and add ceph rbd handler

liuy:
Currently storage engine plugins tightly couple with the iscsi command handling
which not only duplicate lines of code but also make plugin writing painful.

we can avoid it by adding one more indirction, a generic_handle_cmd(), which is
optional and can coexist with current command handling logic. With this added,
the command process can be split into two category:

1. one can register .handle_cmd to take the full control of command handling as
   before
2. or else one can register .{read, write, flush} to only handle storage IO
   stuff.

and then iscsi command handling can sit on top of the storage layer, meaning
that we can handle most of the commands without involving plugins, for e.g
COMPARE AND WRITE is implemented just by generic_handle_cmd().

What make it complex is how we deal with async storage APIs. We could start with
async APIs and build sync ones on top of them to have storage plugin only
implement async APIs. But initially I plan to start with sync APIs instead of
async ones because

1. it's simple and easy to verify the correctness.
2. it's easy to add async api later once upper iscsi layer is matured.
3. Pre-optimization is root of eveil.

Following patches will implement a rbd handler from scratch and adopt glfs
handler to generic_handle_cmd, leaving file and qcow handlers alone.

Todo:
1. add unmap(discard) operation
2. handle more commands (WRITE SAME, WRITE VERIFY, etc.)
3. redesign with or add async APIs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ set(tcmu-runner_HANDLER_PATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/tc
 
 option(with-glfs "build Gluster glfs handler" true)
 option(with-qcow "build qcow handler" true)
+option(with-rbd "build Ceph rbd handler" true)
 
 find_library(LIBNL_LIB nl-3)
 find_library(LIBNL_GENL_LIB nl-genl-3)
@@ -148,6 +149,24 @@ add_executable(consumer
   consumer.c
   )
 target_link_libraries(consumer tcmu)
+
+if (with-rbd)
+	find_library(LIBRBD rbd)
+
+	# Stuff for building the rbd handler
+	add_library(handler_rbd
+	  SHARED
+	  rbd.c
+	  )
+	set_target_properties(handler_rbd
+	  PROPERTIES
+	  PREFIX ""
+	  )
+	target_link_libraries(handler_rbd
+	  ${LIBRBD}
+	  )
+	install(TARGETS handler_rbd DESTINATION ${CMAKE_INSTALL_LIBDIR}/tcmu-runner)
+endif (with-rbd)
 
 if (with-glfs)
 	find_library(GFAPI gfapi)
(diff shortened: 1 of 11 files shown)


liuy:
@agrover any comments?

mikechristie:
The ceph parts look ok to me.

Some comments for later versions when the basics are merged.

For compare and write, we actually do not want it common. It is easier for ceph to pass the command to the OSD and have it do its normal locking/serialization as long as the request is within an object which seems like the common case for the major user ESX. We then do not have to take an extra target side lock, when doing async in the future, or cluster type of lock for HA targets. So for future versions, when we get async support back, we do not have to worry about locking in tcmu for ceph at least.

Here is the compare and write PR for ceph. It needs work for librbd support still:

https://github.com/ceph/ceph/pull/8911

For write same, we have something similar where the OSD does most of the work:

librbd support
https://github.com/ceph/ceph/pull/10019
lower level librados
https://github.com/ceph/ceph/pull/8568

liuy:
@mikechristie Thanks for your comments! Sounds great ceph is becoming more and more iscsi-aware by implementing the low level iscsi semantics APIs. Once these APIs are added, I think it is easy to add pass-through function into generic_cmd_handle(). For e.g, we can register command set in the plugin, then  generic_cmd_handle() check if the plugin have registered command set in the first place, and if yes, simply pass through the command that is one of the registered command set to the plugin, otherwise it goes to generic switch cases handling. For some other non-iscsi aware storage like glusterfs, sheepdog, we can simply register nothing and have generic_cmd_handle() handle the iscsi stuff.

agrover:
hi @liuy sorry for the delay.

I think this looks really good. Please change all references in code or commitlogs from iscsi to SCSI, since that is more accurately what we're working with here.

We also need to document that there are now two ways to implement a tcmu-runner plugin handler. The current description of options is at the bottom of README.md. Add text to the README.md similar to the explanation you give above in this PR?

Finally, it would be great to emit a warning message in tcmur_register_handler if both handle_cmd and any of the new function pointers are set, which should never normally happen.

Much thanks!

liuy:
@agrover I've addressed your comments and update the master branch :)

agrover:
thanks!

PR #81, opened 2016-12-08
Improvements to glfs handler 

pkalever:
Mostly refactored the code and implement a glfs object cache.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <scsi/scsi.h>
 #include <glusterfs/api/glfs.h>
+#include "darray.h"
 
 #include "tcmu-runner.h"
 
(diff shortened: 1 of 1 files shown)


pkalever:
Please review:
@agrover 
@vbellur
@pranithk

agrover:
big 👍 from me, love the detailed commitlogs too.

No action for 24 hours to give other reviewers a chance.

liuy:
@agrover A off-topic question. I noticed glfs problem is related to the .check_config and open, which almost share the portion of the same logic. [Precisely, .check_config could be implemented by combination of .open and .close]. I am wondering why we need .check_config and when we must implement it for plugin drivers and. It is optional by the code, so when we not have to implement it. So possibly glfs might not need to implement it to avoid introducing a complicated cache trick. Thanks!

agrover:
@liuy I was concerned that there might be handlers for which doing an open/close might not correctly verify that the configstring was correct. I think open/close works for handlers that adapt an API, but for a handler that emulated e.g. a tape jukebox with files, it might want to do something different, I just don't know.

It _is_ awkward. I'm definitely open to less awkward approaches.

PR #82, opened 2016-12-19
Add logger and system config features support

superlibj:
This patch series add logger and system config support.

The logger feature:
For now the log messages will output to system logger. And the non-block feature will be supported in the later series.

The system config feature:
For now it will support 3 type options: int, string and boolean. If the options has been changed in config file, the tcmu-runner, consumer and tcmu-synthesizer daemons should be restarted.
And the dynamic reloading feature will be added later.

Thanks.
BRs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ add_library(tcmu
   libtcmu-register.c
   tcmuhandler-generated.c
   libtcmu_log.c
+  libtcmu_config.c
   )
 set_target_properties(tcmu
   PROPERTIES
@@ -64,6 +65,7 @@ add_library(tcmu_static
   libtcmu-register.c
   tcmuhandler-generated.c
   libtcmu_log.c
+  libtcmu_config.c
   )
 target_include_directories(tcmu_static
   PUBLIC ${LIBNL_INCLUDE_DIR}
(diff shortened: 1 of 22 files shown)


superlibj:
@agrover and everyone

please help review, thanks.

agrover:
thoughts @pkalever @mikechristie ?

agrover:
This is valuable but not directly related to tcmu-runner's purpose. I'm wondering if https://github.com/rsyslog/liblogging might be worth looking into. Do we envision needing further logging improvements in the future? If so, maybe we should just rely on a lib from the start.

Also, we have a need for getting performance information out of tcmu-runner -- iops, throughput, latency etc. This is related but not the same as logging. Any ideas on how we could best achieve this, either together or separate from logging?

liuy:
liblogging seems inactive (build fixes recently) and looks complicated. What we need is a ring-buffer (in any situations, like logger process hangs or crashed, do not block tcmu daemon) based prink-like solution(can choose to output to std, syslog, file), which might be severel hundred of lines only. Rely on a library is not so charming if we can implement it by less than 1000 lines. (Most probjects has its own logger, partly because no good log library satisfy all the needs)

Accounting might be different story, because it need modify internal data structures while logger doesn't. So I'd suggest separate accounting implementation from logging.

superlibj:
This is the base for TCMU logging, planing to use the least code to support as many features as possible, such as rsyslog, file and so on. In the near future, the ring-buffer will be added for log-safer. For many other projects, they are all using separate implementions of themselives from lib logging to satisfy what they need. 

When all the features has been implemented, the perfomance information of tcmu-runner will be provided.

agrover:
looks good. Please rename tcmu_config_destory to tcmu_config_destroy and I'll merge it.

Thanks much!!!

mikechristie:
It looks ok to me.

I am not sure about perf counter stuff yet.

For config support, have you guys used libs like libconfig?

superlibj:
@agrover okay。

superlibj:
@mikechristie not yet，the config feature will be enhance later, and if needed and possible i will support that.

agrover:
thanks again!

PR #83, opened 2016-12-21
Support CMake 2.8 instead of CMake 3.0

dillaman:
RHEL 7 only supports CMake 2.8, so ensure tcmu-runner can
be built within a RHEL 7 build environment.


agrover:
thanks! 👍

PR #84, opened 2017-01-01
libtcmu: add dev and ring buffer cmd_tail update macro APIs

superlibj:
This patch add two cmd_tail update APIs to simplify the code.


agrover:
👍

PR #86, opened 2017-01-05
tcmu-runner: make a cdb debug info to the same line

superlibj:


mikechristie:
Patch looks ok.

What about  12 and 16 cdb length handling? It does not have to be done in this patch since the original code seemed to only handle up to 10.

PR #87, opened 2017-01-06
tcmu-runner: add possible 12 and 16 bytes CDB debuginfo support.

superlibj:

### DIFF
--- a/main.c
+++ b/main.c
@@ -286,14 +286,41 @@ static int generic_handle_cmd(struct tcmu_device *dev,
 	}
 }
 
-#define TCMU_CDB_MAX_LEN 10 /* TODO: is 10 bytes for now */
+#define TCMU_CDB_MAX_LEN 16 /* 16 bytes for now */
 static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
 {
-	int i, n;
+	int i, n, bytes;
 	char buf[3 * TCMU_CDB_MAX_LEN + 1];
-	bool short_cdb = cmd->cdb[0] <= 0x1f;
+	uint8_t group_code = cmd->cdb[0] >> 5;
 
-	for (i = 0, n = 0; i < (short_cdb ? 6 : 10); i++) {
+	switch (group_code) {
+	case 0: /*000b for 6 bytes commands */
+		bytes = 6;
+		break;
+	case 1: /*001b for 10 bytes commands */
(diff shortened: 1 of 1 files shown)


superlibj:
@agrover Are 12 and 16 bytes CDB possible here ?

agrover:
The initiator might send, yes. 

Btw in default case I think we should set to 6. Otherwise bytes is uninitialized.

superlibj:
Do you mean only default case or together with 3,6,,7 case ?

agrover:
I think it's best to still print something here for all possible CDBs, so we can accurately reflect all cmds we are getting. I don't think we can assume much about vendor-specific CDBs, but surely they're at least 6 bytes, so how about we just use bytes=6 for those cases too? @mikechristie ?

mikechristie:
I think the min is 6 for fixed length.

I think we can ignore variable and extended length CDBs until someone adds support for them.

superlibj:
From the "SCSI Commands Reference Manual.pdf"， the fixed CDB has at least 6 bytes, and the variable CDB has 7 bytes + ADDITIONAL CDB LENGTH (n–7).

The vendor-specific CDBs we just print the first 6 bytes for now. And the varible CDBs will print all of them.
Is that Okay ?

mikechristie:
I think this is a ok.

For future references, you can find more SCSI spec on t10.org. If your company is not a member you can still read the drafts which are helpful. They will have a lot of info that the "SCSI command Reference Manual" is missing.

agrover:
Much thanks!

superlibj:
@mikechristie @agrover Yes and thanks verrry much for you info about the SCSI.

liuy:
@agrover Unfortunately, my company is not a member of t10.org.  Any pointers that we could read those references? Drafts indeed are dry and lack of useful information. Most of the time, lines of codes are much easier to understand for us what an opcode is on earth :(

mikechristie:
Go to http://t10.org/scsi-3.htm. Scroll to the  bottom of the page to "DOCUMENT LINKS".  Click the "here" link. You can only read the drafts for the newest version.

For example for this problem you could open SPC http://www.t10.org/cgi-bin/ac.pl?t=f&f=spc5r13.pdf and in section 4 is where CDBs are described. It is basically what you see in the Seagate manual you referenced but it has more info due to being newer. For example, the Seagate manual I think is missing ALUA commands.

PR #88, opened 2017-01-10
tcmu-runner:Fix the possible the memory overflow.

superlibj:


superlibj:
@agrover 
Sorry for last patch, for the variable CDBs the buf may overfow.

agrover:
getting a little crazy but ok 😃

PR #89, opened 2017-01-13
asynchronous rbd handler

vshankar:
These set of changes aims to conver RBD handlers read/write/flush operation asynchronous by making use of asynchronous librbd APIs. This involves changes to `generic_handle_cmd()` to handle `TCMU_ASYNC_HANDLED` return value by store callout methods. Furthermore, `compare_and_write` has been added as a store callout method though keeping the implementation same (read + write as of now, but rbd would make use of `cmpext` OSD op when ready).

### DIFF
--- a/libtcmu_aio.c
+++ b/libtcmu_aio.c
@@ -1,0 +1,298 @@
+/*
+ * Copyright 2017, Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#define _GNU_SOURCE
+#include <errno.h>
+#include <assert.h>
+#include <stdint.h>
+#include <pthread.h>
+
+#include "libtcmu.h"
+#include "libtcmu_priv.h"
+#include "libtcmu_aio.h"
+#include "tcmu-runner.h"
+
+void track_aio_request_start(struct tcmu_device *dev)
+{
+	struct tcmu_track_aio *aio_track = &dev->track_queue;
+
+	pthread_cleanup_push(_cleanup_spin_lock, (void *)&aio_track->track_lock);
+	pthread_spin_lock(&aio_track->track_lock);
+
+	++aio_track->tracked_aio_ops;
+
+	pthread_spin_unlock(&aio_track->track_lock);
+	pthread_cleanup_pop(0);
+}
+
+void track_aio_request_finish(struct tcmu_device *dev, int *is_idle)
+{
+	struct tcmu_track_aio *aio_track = &dev->track_queue;
+
+	pthread_cleanup_push(_cleanup_spin_lock, (void *)&aio_track->track_lock);
+	pthread_spin_lock(&aio_track->track_lock);
+
+	assert(aio_track->tracked_aio_ops > 0);
+
+	--aio_track->tracked_aio_ops;
+	if (is_idle) {
+		*is_idle = (aio_track->tracked_aio_ops == 0) ? 1 : 0;
+	}
+
+	pthread_spin_unlock(&aio_track->track_lock);
+	pthread_cleanup_pop(0);
+}
+
+static void tcmu_callout_finished(struct tcmu_device *dev,
+				  struct tcmulib_cmd *cmd, int ret)
+{
+	cmd->callout_cbk(dev, cmd, ret);
+}
+
+/*
+ * This deals with return values from sync and async store handlers.
+ * Possible return values:
+ *
+ * 0: success from sync store handler
+ * -errno: failure from [a]sync store handler
+ * TCMU_ASYNC_HANDLED: command async handled
+ * TCMU_NOT_HANDLED: unhandled command (from command passthru)
+ */
+static int call_stub_exec(struct tcmu_device *dev,
+			  struct tcmulib_cmd *cmd,
+			  struct tcmu_call_stub *stub, bool is_async)
+{
+	ssize_t ret;
+	ssize_t requested;
+	int err = SAM_STAT_GOOD;
+	uint8_t *sense_buf = cmd->sense_buf;
+
+	switch(stub->sop) {
+	case TCMU_STORE_OP_READ:
+	case TCMU_STORE_OP_WRITE:
+		requested = tcmu_iovec_length(stub->u.rw.iov, stub->u.rw.iov_cnt);
+		ret  = stub->u.rw.exec(dev, stub->u.rw.iov,
+				       stub->u.rw.iov_cnt, stub->u.rw.off);
+		if (!is_async) {
+			if (ret != requested)
+				ret = -EIO;
+			else
+				ret = 0;
+		}
+
+		break;
+	case TCMU_STORE_OP_FLUSH:
+		ret = (ssize_t) stub->u.flush.exec(dev);
+		break;
+	case TCMU_STORE_OP_HANDLE_CMD:
+		ret = (ssize_t) stub->u.handle_cmd.exec(dev, cmd);
+		break;
+	default:
+		tcmu_err("unhandled store operation\n");
+		assert(0 == "unhandled store operation");
+	}
+
+	if (ret < 0) {
+		if (ret == TCMU_ASYNC_HANDLED || ret == TCMU_NOT_HANDLED)
+			err = ret;
+		else
+			err = errno_to_sam_status(ret, sense_buf);
+	}
+
+	if (!is_async)
+		tcmu_callout_finished(dev, cmd, err);
+	return err;
+}
+
+static void _cleanup_io_work(void *arg)
+{
+	free(arg);
+}
+
+static void *io_work_queue(void *arg)
+{
+	struct tcmu_device *dev = arg;
+	struct tcmu_io_queue *io_wq = &dev->work_queue;
+
+	while (1) {
+		struct tcmu_io_entry *io_entry;
+		struct tcmulib_cmd *cmd;
+
+		pthread_cleanup_push(_cleanup_mutex_lock, &io_wq->io_lock);
+		pthread_mutex_lock(&io_wq->io_lock);
+
+		while (list_empty(&io_wq->io_queue)) {
+			pthread_cond_wait(&io_wq->io_cond,
+					  &io_wq->io_lock);
+		}
+
+		io_entry = list_first_entry(&io_wq->io_queue,
+					    struct tcmu_io_entry, entry);
+		list_del(&io_entry->entry);
+
+		pthread_mutex_unlock(&io_wq->io_lock);
+		pthread_cleanup_pop(0);
+
+		/* kick start I/O request */
+		cmd = io_entry->cmd;
+		pthread_cleanup_push(_cleanup_io_work, io_entry);
+
+		(void) call_stub_exec(io_entry->dev, cmd, &io_entry->stub,
+				      false);
+		pthread_cleanup_pop(1); /* cleanup io_entry */
+	}
+
+	return NULL;
+}
+
+static int aio_schedule(struct tcmu_device *dev,
+			struct tcmulib_cmd *cmd,
+			struct tcmu_call_stub *stub)
+{
+	struct tcmu_io_entry *io_entry;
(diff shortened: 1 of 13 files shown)


dillaman:
@vshankar I have a feeling this won't be thread-safe since there are no locks between the command processing thread [1] and and the librbd aio callback [2]. Also note that you really only want to invoke ```tcmulib_processing_complete``` when idle since this is what wakes up the kernel. 

One possible solution to this would be to follow the fio-style approach and keep a list of in-flight librbd aio completions via the new ```rbd_aio_cb``` structure, add a way to wakeup the main thread's poll [3], and a new callout for completing commands that the main thread would invoke as part of its loop so that it can determine if ```tcmulib_processing_complete``` should be invoked.

[1] https://github.com/open-iscsi/tcmu-runner/blob/d0f347022f87936da10f9fea47c568075ef793d4/main.c#L354
[2] https://github.com/open-iscsi/tcmu-runner/pull/89/commits/f341398a8d564dca8ed90f4476c3fd4e8d7640d2#diff-612499a21226d10f78f40c2c58a896f2R272
[3] https://github.com/open-iscsi/tcmu-runner/blob/d0f347022f87936da10f9fea47c568075ef793d4/main.c#L368

vshankar:
> @vshankar I have a feeling this won't be thread-safe since there are no locks between the command processing thread [1] and and the librbd aio callback [2].

vshankar:
[Ignore the close/reopen PR; some magic keybinding that closed the PR]

> @vshankar I have a feeling this won't be thread-safe since there are no locks between the command processing thread [1] and and the librbd aio callback [2]. Also note that you really only want to invoke tcmulib_processing_complete when idle since this is what wakes up the kernel.

async versions of read/write/flush return `TCMU_ASYNC_HANDLED` with the callbacks completing the request by calling `tcmulib_{command,processing}_complete()`. The command processing thread does not complete/finish the request in this case.

> 
> One possible solution to this would be to follow the fio-style approach and keep a list of in-flight librbd aio completions via the new rbd_aio_cb structure, add a way to wakeup the main thread's poll [3], and a new callout for completing commands that the main thread would invoke as part of its loop so that it can determine if tcmulib_processing_complete should be invoked.

I will think more about this. Wanted to get a working version which can be used to run some tests.

dillaman:
@vshankar 

> async versions of read/write/flush return TCMU_ASYNC_HANDLED with the callbacks completing the request by calling tcmulib_{command,processing}_complete(). The command processing thread does not complete/finish the request in this case.

Correct, the event thread won't process this command -- but it might concurrently process another event that is not async and race w/ the completion (or two AIO callbacks might race each other). There is currently no thread protection within ```tcmulib_command_complete``` so you will most likely corrupt the contents of the ring buffer.

vshankar:
> Correct, the event thread won't process this command -- but it might concurrently process another event that is not async and race w/ the completion (or two AIO callbacks might race each other). There is currently no thread protection within tcmulib_command_complete so you will most likely corrupt the contents of the ring buffer.

Ah! didn't realize that command completion is not thread-safe, will fix that too. Thanks!

libtcmu-generic.c
```diff
@@ -0,0 +1,84 @@
...
+
+#include <stdlib.h>
+#include <errno.h>
+#include <scsi/scsi.h>
+
+#include "scsi_defs.h"
+#include "libtcmu_log.h"
+#include "libtcmu-generic.h"
+
+int generic_compare_and_write(struct tcmu_device *dev,
```
mikechristie: Hey,

We have been putting emulated commands in api.c with names like tmcu_emulate_$command_name(). I think you would want to put this command there too.

vshankar: Thanks. Will move this helper there.

liuy:
There is no need to touch compare_and_write if ceph is not ready.  As per #78 ,  it's better for us to provide a passthrough mechanism instead because not all backends are SCSI aware. I don't think it is good idea to make TCMU too much ceph centric. And compare_and_write should be another pr, in this pr, you'd better concentrate on the AIO support.

As for AIO support, I don't think it is good idea to combine async and sync semantics in a single function. I guess it would be better to add aio_{read, write, flush} instead and allow upper layer (generic_cmd_handling) to be not confused with what mode(async or sync) {read, write, flush} is.

liuy:
BTW, could you please give a performance number about async and sync ceph path?

vshankar:
> There is no need to touch compare_and_write if ceph is not ready. As per #78 , it's better for us to provide a passthrough mechanism instead because not all backends are SCSI aware. I don't think it is good idea to make TCMU too much ceph centric. And compare_and_write should be another pr, in this pr, you'd better concentrate on the AIO support.

I'll remove `comapre_and_write` from this PR. Also, to not make tcmu ceph centric, `generic_handle_cmd` can invoke handler specific methods for registered commands (such as `compare_and_write`) if available or emulate it as it does now. Is this what you mean by having plugin based command set (mentioned in PR #78)?

> 
> As for AIO support, I don't think it is good idea to combine async and sync semantics in a single function. I guess it would be better to add aio_{read, write, flush} instead and allow upper layer (generic_cmd_handling) to be not confused with what mode(async or sync) {read, write, flush} is.

That's was one of the options when starting this out, but the changes to have a single function support sync/async semantics looked pretty neat to me especially when the command processing thread[1] does care (via return value) if the command is handled  [a]synchronously.

[1]: https://github.com/open-iscsi/tcmu-runner/blob/master/main.c#L355

liuy:
> I'll remove comapre_and_write from this PR. Also, to not make tcmu ceph centric, generic_handle_cmd can invoke handler specific methods for registered commands (such as compare_and_write) if available or emulate it as it does now. Is this what you mean by having plugin based command set (mentioned in PR #78)?

I'm going to make a pr about 'command passthru' and let code explain :)

> That's was one of the options when starting this out, but the changes to have a single function support sync/async semantics looked pretty neat to me especially when the command processing thread1 does care (via return value) if the command is handled [a]synchronously.

Yes, there are many options and very hard to decide which one is the best until we have a full picture. That is the very reason I didn't include async APIs in #78 . The problem is that, some commands like compare_and_write might need sync read & write. There might be more of this kind command in the future. So not so ture of that you mentioned.

Building sync APIs on top of async ones is a common practice. So the store engine might only provide async ones and sync APIs can emulated in the upper generic layer if needed.

liuy:
@vshankar I've make a pr #91 With passthru support, we can handle any portion of the command set without adding more and more non-storage handlers in the structure.

vshankar:
Updating the PR with some enhancements -- probably not ready for full review as comments by @liuy needs to be addressed.

vshankar:
> Yes, there are many options and very hard to decide which one is the best until we have a full picture. That is the very reason I didn't include async APIs in #78 . The problem is that, some commands like compare_and_write might need sync read & write. There might be more of this kind command in the future. So not so ture of that you mentioned.
> 
> Building sync APIs on top of async ones is a common practice. So the store engine might only provide async ones and sync APIs can emulated in the upper generic layer if needed.

This seems to be a nice way forward. What do others have to think about it? @agrover @mikechristie @dillaman

mikechristie:
I prefer the async based approach.

We have 3 handler types right now:
1. tcmulib_handler - This type does whatever it wants because it handles all processing. We do not worry about this type.

2. tcmur_handler - handle_cmd based.

3. tcmur_handler - callout per operation (read/write/flush) type based.

We are saying #2 and #3 should be async now right? We can then remove some if/elses and worry about only one code path type. If the handler does not have a native async API, tcmu-runner should provide the threading helpers so we stop duplicating that type of code in the handlers.

Do we need sync based APIs? For compare and write and write verify, it seems like it would be simple to kick off the second part of the command from the completion handler of the first.

list.h
```diff
@@ -0,0 +1,252 @@
+#ifndef _LINUX_LIST_H
+#define _LINUX_LIST_H
```
BC-SDS: I guess we have a license problem here. Linux's GPLv2... Linux kernel's rbtree and list, bit operations are really useful, dunno how we can tackle with license collison

mikechristie: Is /usr/include/sys/queue.h ok to use in a apache 2 project? It has a list at least.

vshankar: @BC-SDS Thanks for reviewing. I was planning to use list implementation from ccan (https://github.com/rustyrussell/ccan/tree/master/ccan/list). Had linux kernel list already checked out, so used that. How about using from ccan? Its BSD-MIT license.

liuy: BC-SDS is one of my account. Didn't notice I logged in as BC-SDS. I hope we can find a licence compatible library for a broader data structure, like list, rbtree, bit operations, etc.

vshankar:
> We are saying #2 and #3 should be async now right? We can then remove some if/elses and worry about only one code path type. If the handler does not have a native async API, tcmu-runner should provide the threading helpers so we stop duplicating that type of code in the handlers.

Yeh -- store handlers could hint if its capable of performing asynchronous IO. Sync handlers can be kicked off via worker threads (> 1)?

> 
> Do we need sync based APIs? For compare and write and write verify, it seems like it would be simple to kick off the second part of the command from the completion handler of the first.

sync at the tcmulib level -- seems not. We can emulate async behavior even even if the handlers are synchronous.

vshankar:
Would depend on #98

vshankar:
This is not rebased with #98 yet as it might be helpful for some performance runs.

mikechristie:
For function naming prefixes let's do the following:

- If a function is exported to a handler, add at least a "tcmu_" prefix. For example, errno_to_sam_status should be tcmu_errno_to_sam_status. I think probably you meant to name tcmu_command_passthrough something like call_store_passthrough.
- If a function is used by handler that does its own event loop then the function gets the tcmulib_ prefix. tcmulib_track_aio_request_start would just be track_aio_request_start. None of the code you are modifying/adding in these patches is this type of handler.

libtcmu_aio.c
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	return rc;
+}
+
+static void call_stub_exec_sync(struct tcmu_device *dev,
+				struct tcmulib_cmd *cmd,
+				struct tcmu_call_stub *stub)
+{
+	ssize_t ret;
```
mikechristie: Do we need both of these because you are having the store callouts return different values? I do not think we need to. Bryant and I have converted qcow and file* to use the store callouts and to work similarly.

You could have them all return normal old ints. -EXYZ == error. 0 means it was successfully completed or queued for the async case. It seems we fail if the entire IO cannot be completed so we do not need to return the number of bytes read/written, but we could return the number of bytes in the sync case if we needed it.

Or.

You can have them all return SCSI and TCMU values. The handler calls tcmu_set_sense_data on failure.

It does not matter to me. I like handlers not having to deal with SCSI details, but the second option is more flexibly. Any preference?

libtcmu_store.c
```diff
@@ -19,101 +19,628 @@
...
+			  struct iovec *iovec, size_t iov_cnt, off_t off)
+{
+	int ret;
+	struct tcmu_call_stub stub;
+	struct tcmulib_cmd *readcmd;
+	uint8_t *sense = tcmulib_cmd->sense_buf;
+	ssize_t half = (tcmu_iovec_length(iovec, iov_cnt)) / 2;
+
+	ret = errno_to_sam_status(-ENOMEM, sense);
+
```
mikechristie: The execution of this entire command has to be atomic, so you need a mutex.

libtcmu_store.c
```diff
@@ -19,101 +19,628 @@
...
+		return;
+	}
+
+	/* passthrough command was not handled - fallback to generic handling */
+	ret = call_store_handler(dev, store, tcmulib_cmd, cmd);
+	tcmu_command_finish(dev, tcmulib_cmd, ret,
+			    (ret != TCMU_ASYNC_HANDLED) ? true : false);
+}
+
+static int tcmu_command_passthrough(struct tcmu_device *dev,
```
mikechristie: Did you mean to call it call_command_passthrough() like your other functions.

libtcmu_store.c
vshankar: Ummm.. looks like `file_optimal` too does not lock around the compound operations. Also, this would mean introducing locks in `read`, `write` and `flush` paths.

libtcmu_store.c
vshankar: hmm, yeh. will change.

libtcmu_store.c
mikechristie: Only CAW requires a lock. It is defined to be atomic. It ESX it is the basis for its atomic test and set command.

I am not sure why it requires locks around all those calls. Just take a device level mutex at the beginning of the caw call and then drop it at completion/failure.

vshankar: A separate `write` could sneak in between a the `read` and `write` operation in CAW. Wouldn't that break the atomicity of the CAW operation?

vshankar:
@mikechristie addressed comments in PR #98 and this pr is rebased over that.

mikechristie:
Thanks Venky.

PR #90, opened 2017-01-16
Add write_verify support

bgly:
This commit adds WRITE VERIFY for qcow and file

### DIFF
--- a/api.c
+++ b/api.c
@@ -887,7 +887,7 @@ int tcmu_emulate_mode_select(
 }
 
 int tcmu_emulate_start_stop(struct tcmu_device *dev, uint8_t *cdb,
-			                                uint8_t *sense)
+			    uint8_t *sense)
 {
 	if ((cdb[4] >> 4) & 0xf)
 		return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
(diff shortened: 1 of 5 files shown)


bgly:
@agrover Review please

mikechristie:
Could the do_verify_op and the write part of the implementations be made generic? It could look similar to vshankar's generic_compare_and_write where it takes the handler specific read/write functions as arguments.

bgly:
I added a few more patches to this patch set, that addresses some other problems while testing bonnie++ performance test on user:file/user:file_async.

bgly:
I have tested installing fedora using user:fbo onto a user:file backstore. Then I also manually created a file system using user:file_async/user:file backstore and ran bonnie++ performance tests on this patch without any failures.

mikechristie:
I think the patches look ok.

Just 2 questions/comments.

1. Why was sync cache only implemented for the async file handler? Just wondering because it looked like you could support imm=0 for the sync case still.

2. Andy, is the policy going forward to do the generic_handle_cmd approach? It looks like some of those handlers have the out of bounds issue found and fixed in file_example here. Do we need the handle_cmd model? It looks like we could convert qcow and once we settle the debate on async vs sync we could convert file_example.

bgly:
Yeah, I was just testing file_async so I bounded it towards that but it will also work for file sync. I uploaded a new patch

file_example.c
```diff
@@ -360,33 +398,39 @@ static int file_handle_cmd(
...
+		unsigned int to_copy;
+		int i = 0;
 
 		remaining = length;
 
-		while (remaining) {
-			unsigned int to_copy;
-
-			to_copy = (remaining > iovec->iov_len) ? iovec->iov_len : remaining;
+		tcmu_info("Enter write: %m\n");
```
mikechristie: Hey.

Were the tcmu_info calls here and below supposed to be tcmu_dbg or deleted for the final merge?

main.c
```diff
@@ -230,6 +230,9 @@ static int generic_handle_cmd(struct tcmu_device *dev,
 						   ASC_READ_ERROR, NULL);
 		} else
 			return SAM_STAT_GOOD;
+	case WRITE_VERIFY:
+		return store->write_and_verify(dev, tcmulib_cmd, iovec,
+					       iov_cnt, offset);
```
mikechristie: For this you mean to call the function you added right?

tcmu_emulate_write_verify()

?

tcmu-runner.h
```diff
@@ -72,6 +72,8 @@ struct tcmur_handler {
 	ssize_t (*write)(struct tcmu_device *, struct iovec *, size_t, off_t);
 	ssize_t (*read)(struct tcmu_device *, struct iovec *, size_t, off_t);
 	int (*flush)(struct tcmu_device *);
+	int (*write_and_verify)(struct tcmu_device *, struct tcmulib_cmd *,
+				struct iovec *, size_t, off_t);
```
mikechristie: We can just drop this.

main.c
bgly: Yes, fixed

tcmu-runner.h
bgly: Done

file_example.c
bgly: Deleted for final merge

bgly: Deleted for final merge

api.c
```diff
@@ -905,3 +905,72 @@ int tcmu_emulate_start_stop(struct tcmu_device *dev, uint8_t *cdb,
...
+
+		if (rc != SAM_STAT_GOOD) {
+			tcmu_err("write_verify failed: %m\n");
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+
+		tcmu_seek_in_iovec(iovec, ret);
+		remaining -= ret;
+		}
+	}
```
mikechristie: Something got messed up in one of your uploads. That bracket was supposed to be before the tcmu_seek_in_iovec call right?

}
tcmu_seek_in_iovec(iovec, ret); 
remaining -= ret;

Do not worry about re-uploading. I will fix it up when I merge.

mikechristie:
Merged through command line with these fixes:

https://github.com/open-iscsi/tcmu-runner/commit/5934dca954ad58400b574deb50528d2d69ba470b

PR #91, opened 2017-01-19
runner: add command passthrough support for generic_handle_cmd

liuy:
COMPARE_AND_WRITE in rbd.c is just an example (which will be overloaded once
ceph part is ready) about how we can pass down chose command to the specific
store.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -237,14 +237,19 @@ static int generic_handle_cmd(struct tcmu_device *dev,
 	uint32_t block_size = tcmu_get_dev_block_size(dev);
 	uint64_t num_lbas = tcmu_get_dev_num_lbas(dev);
 	uint8_t cmd;
-	ssize_t ret, l = tcmu_iovec_length(iovec, iov_cnt);
+	ssize_t ret = TCMU_NOT_HANDLED, l = tcmu_iovec_length(iovec, iov_cnt);
 	off_t offset = block_size * tcmu_get_lba(cdb);
 	struct iovec iov;
 	size_t half = l / 2;
 	uint32_t cmp_offset;
 
-	cmd = cdb[0];
+	if (store->handle_cmd)
+		ret = store->handle_cmd(dev, tcmulib_cmd);
 
+	if (ret != TCMU_NOT_HANDLED)
+		return ret;
+
+	cmd = cdb[0];
 	switch (cmd) {
 	case INQUIRY:
 		return tcmu_emulate_inquiry(dev, cdb, iovec, iov_cnt, sense);
@@ -463,7 +468,8 @@ static int invokecmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	struct tcmur_handler *r_handler = handler->hm_private;
 
 	int (*func)(struct tcmu_device *, struct tcmulib_cmd *) =
-		(r_handler->handle_cmd) ? : generic_handle_cmd;
+		(r_handler->write || r_handler->read || r_handler->flush) ?
+		generic_handle_cmd : r_handler->handle_cmd;
 	return func(dev, cmd);
 }
 
--- a/main.c
+++ b/main.c
@@ -67,12 +67,6 @@ static struct tcmur_handler *find_handler_by_subtype(gchar *subtype)
 
 int tcmur_register_handler(struct tcmur_handler *handler)
 {
-	if (handler->handle_cmd &&
-	    (handler->read || handler->write || handler->flush)) {
-		tcmu_err("Skip bad handler: %s\n", handler->name);
-		return -1;
-	}
-
 	darray_append(g_runner_handlers, handler);
 	return 0;
 }


main.c
```diff
@@ -74,10 +74,11 @@ static struct tcmur_handler *find_handler_by_subtype(gchar *subtype)
 
 int tcmur_register_handler(struct tcmur_handler *handler)
 {
-	if (handler->handle_cmd &&
-	    (handler->read || handler->write || handler->flush)) {
-		tcmu_err("Skip bad handler: %s\n", handler->name);
-		return -1;
+	if (handler->read || handler->write || handler->flush) {
+		if (handler->handle_cmd && !handler->cmdset) {
```
vshankar: A check for the inverse is also required (to be safe), else it will segfault during `->handler_cmd()`

rbd.c
```diff
@@ -203,6 +203,66 @@ static int tcmu_rbd_flush(struct tcmu_device *dev)
...
+	switch (cmd) {
+	case COMPARE_AND_WRITE:
+		iov.iov_base = malloc(half);
+		if (!iov.iov_base) {
+                        tcmu_err("out of memory\n");
+                        return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+                                                   ASC_READ_ERROR, NULL);
+                }
+                iov.iov_len = half;
+                ret = store->read(dev, &iov, 1, offset);
```
vshankar: nit: could be just `tcmu_rbd_read()` as control has already reached rbd handler.

rbd.c
```diff
@@ -203,6 +203,66 @@ static int tcmu_rbd_flush(struct tcmu_device *dev)
...
+                cmp_offset = tcmu_compare_with_iovec(iov.iov_base, iovec, half);
+                if (cmp_offset != -1) {
+                        return tcmu_set_sense_data(sense, MISCOMPARE,
+                                        ASC_MISCOMPARE_DURING_VERIFY_OPERATION,
+                                        &cmp_offset);
+                }
+                free(iov.iov_base);
+
+                tcmu_seek_in_iovec(iovec, half);
+                ret = store->write(dev, iovec, iov_cnt, offset);
```
vshankar: nit: same as for the `read` case.

main.c
liuy: This check just means to stop people using both .handle_cmd(no cmdset) and .{read, write, flush}. Strict checking of all the possible cases for store driver is not so necessary because we examine it by human review beforehand.  Let's keep it simple as started and add more gradients if necessary later.

liuy:
@agrover @vshankar patch updated, please review!

mikechristie:
For MAINTENANCE_IN/MAINTENANCE_OUT, SERVICE_ACTION, PERSISTENT_* etc type of commands where cdb[N> 0] is used you need a more complex table or instead add a new callout that returns if pass through should be used.

liuy:
> For MAINTENANCE_IN/MAINTENANCE_OUT, SERVICE_ACTION, PERSISTENT_* etc type of commands where cdb[N> 0] is used you need a more complex table or instead add a new callout that returns if pass through should be used.

First of all, do we need to pass MAINTENANCE_IN/MAINTENANCE_OUT, SERVICE_ACTION, PERSISTENT_* to store engine and why?

Even if we do, once we have MAINTENANCE_IN/MAINTENANCE_OUT, SERVICE_ACTION, PERSISTENT_* support, we can consider extending array to be a more complext k-v data structure.

mikechristie:
For STPG handling, rbd needs to grab the rbd/ceph exclusive lock and blacklist holders we cannot cleanup. For PGRs, we need to do something similar. We can either add main.c code and then call a new tcmu_handler callout or we can do the passthrough method you are implementing.

It is not clear to me when we add a new callout vs when you do passthrough.

mikechristie:
But yeah you are right, I can add that later when the ALUA stuff is sent.

liuy:
> It is not clear to me when we add a new callout vs when you do passthrough.

There is no rule but I think we should only add a callout when *all* store drivers should(or do) provide a backend implementation otherwise we should pass through the commands to the specific driver which is the very(or only) one support, especially it is non storage related generally. Passthru mechanism could be analogous with system call ioctl(2) :)

liuy:
Update the commit log about when we should add a callout or use passthrough.

mikechristie:
I think just drop the cmdset array and call handle_cmd from generic_handle_cmd if set. If the handler supports the command it executes it. If not it returns TCMU_NOT_HANDLED.

The thread_start() test should then check if it is a generic_handle_cmd type of handler. If so call generic_handle_cmd().

liuy:
> I think just drop the cmdset array and call handle_cmd from generic_handle_cmd if set. If the handler supports the command it executes it. If not it returns TCMU_NOT_HANDLED.

Yeah, looks neat. Patch updated!

mikechristie:
Looks good to me.

Will get it merged later today/tomorrow. Fixing up conflicts with PR 94.

mikechristie:
There were some complications with this patch due to how some non upstream handlers want their handle_cmd called. Will upload a new patch tomorrow.

liuy:
> There were some complications with this patch due to how some non upstream handlers want their handle_cmd called. Will upload a new patch tomorrow.

what is the matter? This patch will call handle_cmd() in the first place so any handlers registered .handle_cmd will be called before generic code path and only opcode returned with NOT_HANDLED will go through generic path. How this cause problem?

mikechristie:
There are handlers that do not want to do the low level tcmu command processing (they do not want to do a consumer.c type of example) but want to do all the SCSI command processing (they do not want to do generic_handle_cmd type of example).

Currently, if handle_cmd does not handle a command the command is failed. If we now drop down to generic_handle_cmd we will be adding functionality the above type of handler may not want. Because they may not be upstream, we have no idea what they are doing so I am worried about changing behavior on them. For example when PGR or ALUA support is implemented, the device state could be changed under the handler.

So to handle that type of case, with your patch if we add new behavior/commands how do we warn non upstream handlers in each release?

1. Do we take the linux kernel approach and not warn them and just say you should get your code upstream?

2. Add some info to the README and bump the libtcmu soname version, and let them figure everything out and change their handle_cmd accordingly?

3. ?


For my patch I was just going to move the callouts that generic_handle_cmd uses (read/write/flush) to a new struct, cmd_ops, and we basically do:

If cmd_ops
   generic_handle_cmd
else
   handle_cmd

generic_handle_cmd then calls handle_cmd like in your patch.

We then let the handle_cmd have exactly what they want.

liuy:
> For my patch I was just going to move the callouts that generic_handle_cmd uses (read/write/flush) to a new struct, cmd_ops, and we basically do:

I've updated the patch and now we keep the original behavior and you can add your patch on top of it. :)

liuy:
@mikechristie any comment? @vshankar 's patch set has conflict with this patch, so I think we'd better to work out this patch first.

mikechristie:
Patch is fine.

I just want to drop the rbd compare and write parts. It is ok as an example, but just adds extra porting work for Venky.

mikechristie:
If you are ok with dropping the rbd part of the patch, let me know and I will just drop that part when I merge it. No need for you to re-upload.

liuy:
> If you are ok with dropping the rbd part of the patch, let me know and I will just drop that part when I merge it. No need for you to re-upload.

Fine to me. Go ahead please

vshankar:
> Patch is fine.
> 
> I just want to drop the rbd compare and write parts. It is ok as an example, but just adds extra porting work for Venky.

ok with me too.

libtcmu.c
```diff
@@ -243,8 +243,13 @@ static int generic_handle_cmd(struct tcmu_device *dev,
 	size_t half = l / 2;
 	uint32_t cmp_offset;
 
-	cmd = cdb[0];
+	if (store->handle_cmd)
+		ret = store->handle_cmd(dev, tcmulib_cmd);
+
+	if (ret != TCMU_NOT_HANDLED)
```
vshankar: Initialize `ret` with `TCMU_NOT_HANDLED` in case `store->handle_cmd` is not exported by handlers.

liuy: Good catch. Will update it

liuy:
update re-updated @mikechristie @vshankar

mikechristie:
Thanks luiy.

I added a small addition to the README about handle_cmd's use for the passthrough use in this patch.

PR #94, opened 2017-01-25
[WIP] asynchronous runner core

vshankar:
This PR changes the core runner to asynchronously execute synchronous handler calls via a worker thread. This makes the command processing thread to be generic in nature without knowing the execution semantics of the handlers (sync vs async). The number of worker threads is kept to one to match the synchronous nature that already exists (although that could be changed to a higher value if desired). This PR sets up infrastructure to efficiently write asynchronous handlers without duplicating code.

NOTE:
* PR #89 will be resubmitted to make use of this PR
* Need an license friendly implementation of `list.h` type linked list (this PR uses the linux kernel version as PR #89 introduced it -- that needs to be changed). What about using one from CCAN (https://github.com/rustyrussell/ccan/tree/master/ccan/list)?

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -37,6 +37,7 @@
 #include "libtcmu.h"
 #include "libtcmu_log.h"
 #include "libtcmu_priv.h"
+#include "tcmu-runner.h"
 
 #define ARRAY_SIZE(X) (sizeof(X) / sizeof((X)[0]))
 
@@ -44,6 +45,8 @@ static struct nla_policy tcmu_attr_policy[TCMU_ATTR_MAX+1] = {
 	[TCMU_ATTR_DEVICE]	= { .type = NLA_STRING },
 	[TCMU_ATTR_MINOR]	= { .type = NLA_U32 },
 };
+
+static darray(struct tcmu_thread) g_threads = darray_new();
 
 static int add_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring);
 static void remove_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring);
@@ -155,6 +158,27 @@ static void teardown_netlink(struct nl_sock *sock)
 	nl_socket_free(sock);
 }
 
+static void cancel_thread(pthread_t thread)
+{
+	void *join_retval;
+	int ret;
+
+	ret = pthread_cancel(thread);
+	if (ret) {
+		tcmu_err("pthread_cancel failed with value %d\n", ret);
+		return;
+	}
+
+	ret = pthread_join(thread, &join_retval);
+	if (ret) {
+		tcmu_err("pthread_join failed with value %d\n", ret);
+		return;
+	}
+
+	if (join_retval != PTHREAD_CANCELED)
+		tcmu_err("unexpected join retval: %p\n", join_retval);
+}
+
 static struct tcmulib_handler *find_handler(struct tcmulib_context *ctx,
 					    char *cfgstring)
 {
@@ -169,6 +193,249 @@ static struct tcmulib_handler *find_handler(struct tcmulib_context *ctx,
 		if (!strncmp(cfgstring, handler->subtype, len))
 		    return handler;
 	}
+
+	return NULL;
+}
+
+static void cmdproc_thread_cleanup(void *arg)
+{
+	struct tcmu_device *dev = arg;
+	struct tcmulib_handler *handler = tcmu_get_dev_handler(dev);
+	struct tcmur_handler *r_handler = handler->hm_private;
+
+	r_handler->close(dev);
+	free(dev);
+}
+
+static int generic_handle_cmd(struct tcmu_device *dev,
+			      struct tcmulib_cmd *tcmulib_cmd)
+{
+	struct tcmulib_handler *handler = tcmu_get_dev_handler(dev);
+	struct tcmur_handler *store = handler->hm_private;
+	uint8_t *cdb = tcmulib_cmd->cdb;
+	struct iovec *iovec = tcmulib_cmd->iovec;
+	size_t iov_cnt = tcmulib_cmd->iov_cnt;
+	uint8_t *sense = tcmulib_cmd->sense_buf;
+	uint32_t block_size = tcmu_get_dev_block_size(dev);
+	uint64_t num_lbas = tcmu_get_dev_num_lbas(dev);
+	uint8_t cmd;
+	ssize_t ret, l = tcmu_iovec_length(iovec, iov_cnt);
+	off_t offset = block_size * tcmu_get_lba(cdb);
+	struct iovec iov;
+	size_t half = l / 2;
+	uint32_t cmp_offset;
+
+	cmd = cdb[0];
+
+	switch (cmd) {
+	case INQUIRY:
+		return tcmu_emulate_inquiry(dev, cdb, iovec, iov_cnt, sense);
+	case TEST_UNIT_READY:
+		return tcmu_emulate_test_unit_ready(cdb, iovec, iov_cnt, sense);
+	case SERVICE_ACTION_IN_16:
+		if (cdb[1] == READ_CAPACITY_16)
+			return tcmu_emulate_read_capacity_16(num_lbas,
+							     block_size,
+							     cdb, iovec,
+							     iov_cnt, sense);
+		else
+			return TCMU_NOT_HANDLED;
+	case READ_CAPACITY:
+		if ((cdb[1] & 0x01) || (cdb[8] & 0x01))
+			/* Reserved bits for MM logical units */
+			return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+						   ASC_INVALID_FIELD_IN_CDB,
+						   NULL);
+		else
+			return tcmu_emulate_read_capacity_10(num_lbas,
+							     block_size,
+							     cdb, iovec,
+							     iov_cnt, sense);
+	case MODE_SENSE:
+	case MODE_SENSE_10:
+		return tcmu_emulate_mode_sense(cdb, iovec, iov_cnt, sense);
+	case START_STOP:
+		return tcmu_emulate_start_stop(dev, cdb, sense);
+	case MODE_SELECT:
+	case MODE_SELECT_10:
+		return tcmu_emulate_mode_select(cdb, iovec, iov_cnt, sense);
+	case READ_6:
+	case READ_10:
+	case READ_12:
+	case READ_16:
+		ret = store->read(dev, iovec, iov_cnt, offset);
+		if (ret != l) {
+			tcmu_err("Error on read %x, %x\n", ret, l);
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		} else
+			return SAM_STAT_GOOD;
+	case WRITE_VERIFY:
+		return tcmu_emulate_write_verify(dev, tcmulib_cmd,
+						 store->read,
+						 store->write,
+						 iovec, iov_cnt, offset);
+	case WRITE_6:
+	case WRITE_10:
+	case WRITE_12:
+	case WRITE_16:
+		ret = store->write(dev, iovec, iov_cnt, offset);
+		if (ret != l) {
+			tcmu_err("Error on write %x, %x\n", ret, l);
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		} else
+			return SAM_STAT_GOOD;
+	case SYNCHRONIZE_CACHE:
+	case SYNCHRONIZE_CACHE_16:
+		ret = store->flush(dev);
+		if (ret < 0) {
+			tcmu_err("Error on flush %x\n", ret);
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		} else
+			return SAM_STAT_GOOD;
+	case COMPARE_AND_WRITE:
+		iov.iov_base = malloc(half);
+		if (!iov.iov_base) {
+			tcmu_err("out of memory\n");
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		}
+		iov.iov_len = half;
+		ret = store->read(dev, &iov, 1, offset);
+		if (ret != l) {
+			tcmu_err("Error on read %x, %x\n", ret, l);
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		}
+		cmp_offset = tcmu_compare_with_iovec(iov.iov_base, iovec, half);
+		if (cmp_offset != -1) {
+			return tcmu_set_sense_data(sense, MISCOMPARE,
+					ASC_MISCOMPARE_DURING_VERIFY_OPERATION,
+					&cmp_offset);
+		}
+		free(iov.iov_base);
+
+		tcmu_seek_in_iovec(iovec, half);
+		ret = store->write(dev, iovec, iov_cnt, offset);
+		if (ret != half) {
+			tcmu_err("Error on write %x, %x\n", ret, half);
+			return tcmu_set_sense_data(sense, MEDIUM_ERROR,
+						   ASC_READ_ERROR, NULL);
+		} else
+			return SAM_STAT_GOOD;
+	default:
+		tcmu_err("unknown command %x\n", cdb[0]);
+		return TCMU_NOT_HANDLED;
+	}
+}
+
+#define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 1)
+#define CDB_FIX_BYTES 64 /* 64 bytes for default */
+#define CDB_FIX_SIZE CDB_TO_BUF_SIZE(CDB_FIX_BYTES)
+static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
+{
+	int i, n, bytes;
+	char fix[CDB_FIX_SIZE], *buf;
+	uint8_t group_code = cmd->cdb[0] >> 5;
+
+	buf = fix;
+
+	switch (group_code) {
+	case 0: /*000b for 6 bytes commands */
+		bytes = 6;
+		break;
+	case 1: /*001b for 10 bytes commands */
+	case 2: /*010b for 10 bytes commands */
+		bytes = 10;
+		break;
+	case 3: /*011b Reserved ? */
+		if (cmd->cdb[0] == 0x7f) {
+			bytes = 7 + cmd->cdb[7];
+			if (bytes > CDB_FIX_SIZE) {
+				buf = malloc(CDB_TO_BUF_SIZE(bytes));
+				if (!buf) {
+					tcmu_err("out of memory\n");
+					return;
+				}
+			}
+		} else {
+			bytes = 6;
+		}
+		break;
+	case 4: /*100b for 16 bytes commands */
+		bytes = 16;
+		break;
+	case 5: /*101b for 12 bytes commands */
+		bytes = 12;
+		break;
+	case 6: /*110b Vendor Specific */
+	case 7: /*111b Vendor Specific */
+	default:
+		/* TODO: */
+		bytes = 6;
+	}
+
+	for (i = 0, n = 0; i < bytes; i++) {
+		n += sprintf(buf + n, "%x ", cmd->cdb[i]);
+	}
+	sprintf(buf + n, "\n");
+
+	tcmu_dbg(buf);
+
+	if (bytes > CDB_FIX_SIZE)
+		free(buf);
+}
+
+static void *tcmu_cmdproc_thread(void *arg)
+{
+	struct tcmu_device *dev = arg;
+	struct tcmulib_handler *handler = tcmu_get_dev_handler(dev);
+	struct tcmur_handler *r_handler = handler->hm_private;
+	struct pollfd pfd;
+	int ret;
+
+	pthread_cleanup_push(cmdproc_thread_cleanup, dev);
+
+	while (1) {
+		int completed = 0;
+		struct tcmulib_cmd *cmd;
+
+		tcmulib_processing_start(dev);
+
+		while ((cmd = tcmulib_get_next_command(dev)) != NULL) {
+			if (tcmu_get_log_level() == TCMU_LOG_DEBUG)
+				tcmu_cdb_debug_info(cmd);
+
+			if (r_handler->handle_cmd)
+				ret = r_handler->handle_cmd(dev, cmd);
+			else
+				ret = generic_handle_cmd(dev, cmd);
+			if (ret != TCMU_ASYNC_HANDLED) {
+				tcmulib_command_complete(dev, cmd, ret);
+				completed = 1;
+			}
+		}
+
+		if (completed)
+			tcmulib_processing_complete(dev);
+
+		pfd.fd = tcmu_get_dev_fd(dev);
+		pfd.events = POLLIN;
+		pfd.revents = 0;
+
+		poll(&pfd, 1, -1);
+
+		if (pfd.revents != POLLIN) {
+			tcmu_err("poll received unexpected revent: 0x%x\n", pfd.revents);
+			break;
+		}
+	}
+
+	tcmu_err("thread terminating, should never happen\n");
+
+	pthread_cleanup_pop(1);
 
 	return NULL;
 }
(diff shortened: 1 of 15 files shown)


mikechristie:
Just a dumb nit. For a overall patchset comment, I think we are using more linux kernel'ish coding style. Tabs instead of spaces and we don't do {}s and indent when we take a lock.

However, we are doing c and c++ comments.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
                 free(buf);
 }
 
-static void *tcmu_cmdproc_thread(void *arg)
+static void _cleanup_mutex_lock(void *arg)
+{
+        pthread_mutex_t *l = arg;
+        pthread_mutex_unlock(l);
```
mikechristie: It's just void * so just pass it directly here and below.

pthread_mutex_unlock(arg);

Also no need to cast to a void * in the pthread_cleanup_push calls below.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
+                pthread_mutex_lock(&aio_wq->aio_lock);
+                {
+                        while (list_empty(&aio_wq->aio_queue)) {
+                                pthread_cond_wait(&aio_wq->aio_cond,
+                                                  &aio_wq->aio_lock);
+                        }
+
+                        aio_entry = list_first_entry(&aio_wq->aio_queue,
+                                                     struct tcmu_aio_entry, entry);
+                        tcmulib_track_aio_request(dev, &aio_entry->entry);
```
mikechristie: Maybe just a comment somewhere about the locking order requirements (aio_lock must be taken before track_lock) so new devs don't have to dig in the code.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
+        ret = pthread_mutex_init(&aio_wq->aio_lock, NULL);
+        if (ret < 0) {
+                goto out;
+        }
+        ret = pthread_cond_init(&aio_wq->aio_cond, NULL);
+        if (ret < 0) {
+                goto cleanup_lock;
+        }
+
+        // TODO: >1 worker threads?
```
mikechristie: Probably just make it a tcmu.conf setting.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
+        int ret;
+        struct tcmu_aio_entry *curr, *next;
+        struct tcmu_aio_queue *aio_wq = &dev->work_queue;
+
+        if (cancel) {
+                cancel_thread(aio_wq->aio_wq_thread);
+        }
+
+        list_for_each_safe(&aio_wq->aio_queue, curr, next, entry) {
+                list_del(&curr->entry);
```
mikechristie: Do we need a tcmulib_command_complete call?

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
-        struct pollfd pfd;
+
         int ret;
+        int (*func)(struct tcmu_device *, struct tcmulib_cmd *) =
+                (r_handler->handle_cmd) ? : generic_handle_cmd;
+
+        if (r_handler->aio_supported && r_handler->aio_supported()) {
+                ret = func(dev, cmd);
+        } else {
+                ret = aio_schedule(dev, cmd, func);
```
mikechristie: If aio_schedule fails with a -EXYZ value I think we end up never completing the command. I think we need to do a tcmulib_command_complete on failure.

For aio_schedule memory allocation failure I think we want to return  SAM_STAT_TASK_SET_FULL to match what the kernel does when it cannot allocate memory .Something like:

tcmulib_command_complete(.... SAM_STAT_TASK_SET_FULL);

tcmu-runner.h
```diff
@@ -60,6 +60,9 @@ struct tcmur_handler {
 	int (*open)(struct tcmu_device *dev);
 	void (*close)(struct tcmu_device *dev);
 
+        /* device supports asynchronous IO */
+        bool (*aio_supported)(void);
```
mikechristie: Should this just be a bool? I do not think we will ever need it to be dynamic.

libtcmu.c
vshankar: Right, needs a call to `tcmulib_command_complete()`

libtcmu.c
mikechristie: Actually, just ignore this for now. I mean make sure the code is not racy with > 1 thread, but I think we can wait for how to config it. I imagine people will want something complicated like per device and/or handler values, so we do not need to complicate your patches with that.

libtcmu.c
```diff
@@ -618,7 +797,19 @@ static void remove_device(struct tcmulib_context *ctx,
...
+
+        /*
+         * The order of cleaning up worker threads and calling ->removed()
+         * is important: for sync handlers, the worker thread needs to be
+         * terminated before removing the handler (i.e., calling handlers
+         * ->close() callout) in order to ensure that no store callouts
+         * are getting invoked when shutting down the handler.
+         */
+        cancel_thread(aio_wq->aio_wq_thread);
         dev->handler->removed(dev);
```
mikechristie: What happens to executing commands that are running when cancel_thread is called? If it is on the track_queue will it ever get completed andcleaned up?

Is the problem for this case that we do not know what command processing state the tcmur_handler was in when the cancel was called?

vshankar: In case of librbd, the outstanding I/O request should get completed (upon call to `rbd_close()`). That would cleanup the tracking queue (via call from request aio completion). Other handlers should also properly cleanup in-flight I/Os when shutdown.

tcmu-runner.h
vshankar: Okay, I gave it the benefit of doubt. Will do the required changes.

libtcmu.c
vshankar: Right, but I'm not sure what should be the return value. Also, thinking that `tcmulib_processing_complete()` need to be called as that seems unnecessary after device removal.

libtcmu.c
vshankar: Yeh -- the point of having a single work thread was to emulate the current sync processing behaviour.

libtcmu.c
vshankar: That came due to some carry forward code to cleanup spinlocks as spinlocks are "volatile int"

libtcmu.c
vshankar: Does `SAM_STAT_COMMAND_TERMINATED` or `SAM_STAT_TASK_ABORTED` would be the return value of choice here?

mikechristie: Yeah, I think your first comment about it being unnecessary is correct for the valid cases.

I think we have these cases:

1. Under normal circumstances during device removal (TCMU_CMD_REMOVED_DEVICE from kernel) it looks like the kernel target layer will call this path only when there are no commands running. target_core_user will even BUG() if there were commands running.

2. For the add_device failures paths there will never be any commands running.

3. There is that tcmulib_close->close_devices->remove_device path, but I do not see tcmulib_close called anywhere. If it can be called by some userspace code when IO is running then we would have to worry about this case.

Maybe we can just delete that code? I would not worry about it until Andy chimes in about if it is necessary.

4. There is a bug/issue where if the target_core_user timed out a command and completed it, userspace could still be processing the command. The kernel could then allow device removal, but whether we call tcmulib_command_complete will not matter since target_core_user already completed the command during the time out handling.

So yeah, I think if we ignore #3 we do not want to call the tcmulib_command_complete.

vshankar:
I've addressed comments (except one which needs Andy's attention) given by Mike in the latest update.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
-	struct pollfd pfd;
+
 	int ret;
+	int (*func)(struct tcmu_device *, struct tcmulib_cmd *) =
+		(r_handler->handle_cmd) ? : generic_handle_cmd;
+
+	if (r_handler->aio_supported) {
+		ret = func(dev, cmd);
+	} else {
+		ret = aio_schedule(dev, cmd, func);
```
dillaman: Since ```func``` is just derived from ```dev```, how about creating a helper function for the logic on lines 549-550 and eliminate the need to pass a function pointer to the ```aio_schedule``` method (and the need to track the function pointer in addition to tracking the device).

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
+	}
+
+	if (ret != TCMU_ASYNC_HANDLED) {
+		tcmulib_command_complete(dev, cmd, ret);
+		tcmulib_processing_complete(dev);
```
dillaman: Nit: am I correct that this would now only be invoked when an error occurs?  Perhaps a comment since I got tripped up here for a little bit.

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
 		free(buf);
 }
 
-static void *tcmu_cmdproc_thread(void *arg)
+static void _cleanup_mutex_lock(void *arg)
+{
+	pthread_mutex_unlock(arg);
+}
+
+static void *aio_work_queue(void *arg)
```
dillaman: Nit: am I correct that this is really just an IO worker thread to wrap synchronous IO calls? If so, perhaps instead of naming it "aio", can it just be "io"?

libtcmu.c
```diff
@@ -383,18 +383,192 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
...
+		pthread_mutex_lock(&aio_wq->aio_lock);
+
+		while (list_empty(&aio_wq->aio_queue)) {
+			pthread_cond_wait(&aio_wq->aio_cond,
+					  &aio_wq->aio_lock);
+		}
+
+		aio_entry = list_first_entry(&aio_wq->aio_queue,
+					     struct tcmu_aio_entry, entry);
+		tcmulib_track_aio_request(dev, &aio_entry->entry);
```
dillaman: What's the purpose of tracking the "io" request? Is it just to try to batch wakeups to the kernel? Couldn't that be done w/ a spinlock around an integer? Is this going to be re-used by true aio devices to batch kernel wakeups?

libtcmu_priv.h
```diff
@@ -47,13 +48,40 @@ struct tcmulib_context {
...
+	int (*call) (struct tcmu_device *, struct tcmulib_cmd *); /* call */
+	struct list_node entry;
+};
+
+struct tcmu_track_aio {
+	pthread_spinlock_t track_lock;
+	struct list_head track_queue;
+};
+
+struct tcmu_aio_queue {
```
dillaman: Nit: this is the synchronous IO queue, correct?

libtcmu_priv.h
```diff
@@ -47,13 +48,40 @@ struct tcmulib_context {
 	GDBusConnection *connection;
 };
 
+struct tcmu_aio_entry {
```
dillaman: Nit: this is a synchronous io, correct?

libtcmu.c
vshankar: Sure, makes sense.

libtcmu.c
vshankar: yeh right, that's only during an error case. I'll comment it out for brevity (along with comments in other parts of the PR).

dillaman: The AIO handlers should always return ASYNC_HANDLED and never invoke command complete / processing complete here. This is really only an error check so comparing to ASYNC_HANDLED as if there was a chance for sync completion is what is confusing.

vshankar: This implies that `async_call_command` should schedule aio for commands such as `compare_and_write`. Handlers would need to selectively return if they are capable of handling IO asynchronously (read, etc..), else we kick commands off via worker thread.

I had `->aio_supported` dynamic earlier but reverted that to just a plain `bool`.

dillaman: Are you saying that a handler that sets ```->aio_supported``` to true might block?

vshankar: Nope, that needs to be avoided. What I'm not completely sure of is that `generic_handle_cmd` handles bunch of commands, some of which do not call into handlers (maybe the do in-memory operations, etc..) at all. Do we need to call these commands via worker thread?

mikechristie: I think we want to avoid per command detection.

For handlers doing handle_cmd, do not worry about them. Export aio_schedule so they do not have to duplicate their threading. Let them do what they want.

For aio_supported only applies to the generic_handle_cmd read/write/flush type of callouts. Commands like INQUIRY and MODE_SENSE can be completed immediately from the calling context. Multi operation commands built on read/write can work like in this example patch built over vshankar's patches. Patch is not even compile tested. The verify commands (https://github.com/open-iscsi/tcmu-runner/pull/90) and xcopy/extended copy type of commands could be implemented like compare and write.

[generic-handle-cmd-async.txt](https://github.com/open-iscsi/tcmu-runner/files/746529/generic-handle-cmd-async.txt)

vshankar: Thanks Mike. I'm introducing changes similar to your patch (generic async compare and write) as part of PR #89 (as that has async (rbd) handler for `generic_handle_cmd`). If that's ok, I'll update this PR having comments addressed or would it be beneficial to have those changes as part of this PR too?

mikechristie: It is fine to break it up into the 2 PRs. I just included the compare and write example in that patch as an example how we could restrict aio_supported use for only the generic_handle_cmd case and also support the more complex commands like compare and write.

mikechristie:
For the tcmulib issue, I think it looks like it could be used by out of tree handlers, but they would not be using this aio code. I think we can just add a TODO comment or info in the commit log describing how that combo is not supported.

main.c
```diff
@@ -157,275 +150,10 @@ static int open_handlers(void)
...
-	struct tcmu_device *dev = arg;
-	struct tcmulib_handler *handler = tcmu_get_dev_handler(dev);
-	struct tcmur_handler *r_handler = handler->hm_private;
-
-	r_handler->close(dev);
-	free(dev);
-}
-
-static int generic_handle_cmd(struct tcmu_device *dev,
-			      struct tcmulib_cmd *tcmulib_cmd)
```
mikechristie: Just a FYI, I busted this part of the patch with the WRITE_VERIFY PR merge. It added a new command to the switch below.

libtcmu.c
```diff
@@ -173,6 +197,433 @@ static struct tcmulib_handler *find_handler(struct tcmulib_context *ctx,
...
+	while (1) {
+		struct tcmulib_cmd *cmd;
+
+		tcmulib_processing_start(dev);
+
+		while ((cmd = tcmulib_get_next_command(dev)) != NULL) {
+			if (tcmu_get_log_level() == TCMU_LOG_DEBUG)
+				tcmu_cdb_debug_info(cmd);
+
+			/* invoke async command */
```
mikechristie: I would just drop the comment or rename the function. They differ by just one word (invoke vs call) :)

libtcmu.c
```diff
@@ -347,6 +821,20 @@ static void remove_device(struct tcmulib_context *ctx,
 
 	darray_remove(ctx->devices, i);
 
+	struct tcmu_io_queue *io_wq = &dev->work_queue;
```
mikechristie: Could you move this with the other variable declarations so we have similar coding style.

libtcmu.c
```diff
@@ -653,11 +1152,11 @@ void tcmulib_command_complete(
...
 		ent->rsp.scsi_status = SAM_STAT_CHECK_CONDITION;
 
-		buf[0] = 0x70;  /* fixed, current */
-		buf[2] = 0x5;   /* illegal request */
+		buf[0] = 0x70;	/* fixed, current */
+		buf[2] = 0x5;	/* illegal request */
 		buf[7] = 0xa;
 		buf[12] = 0x20; /* ASC: invalid command operation code */
-		buf[13] = 0x0;  /* ASCQ: (none) */
+		buf[13] = 0x0;	/* ASCQ: (none) */
```
mikechristie: My eyes are too tired. I could not figure out what changed in here. I am guessing it was some formatting?

I do not think we need it in these patches though.

libtcmu.c
```diff
@@ -695,3 +1197,83 @@ void tcmulib_processing_complete(struct tcmu_device *dev)
 	if (r == -1 && errno != EAGAIN)
 		perror("write");
 }
+
+void tcmulib_track_aio_request_start(struct tcmu_device *dev)
+{
+	struct tcmu_track_aio *aio_track = &dev->track_queue;
+
+	pthread_cleanup_push(_cleanup_spin_lock, (void *)&aio_track->track_lock);
```
mikechristie: Drop the void pointer cast.

libtcmu.c
```diff
@@ -695,3 +1197,83 @@ void tcmulib_processing_complete(struct tcmu_device *dev)
...
+
+	pthread_spin_unlock(&aio_track->track_lock);
+	pthread_cleanup_pop(0);
+}
+
+void tcmulib_track_aio_request_finish(struct tcmu_device *dev, int *is_idle)
+{
+	struct tcmu_track_aio *aio_track = &dev->track_queue;
+
+	pthread_cleanup_push(_cleanup_spin_lock, (void *)&aio_track->track_lock);
```
mikechristie: Same.

libtcmu_common.h
```diff
@@ -75,6 +77,9 @@ int tcmu_emulate_read_capacity_16(uint64_t num_lbas, uint32_t block_size, uint8_
 int tcmu_emulate_mode_sense(uint8_t *cdb, struct iovec *iovec, size_t iov_cnt, uint8_t *sense);
 int tcmu_emulate_mode_select(uint8_t *cdb, struct iovec *iovec, size_t iov_cnt, uint8_t *sense);
 
+void tcmulib_track_aio_request_start(struct tcmu_device *dev);
+void tcmulib_track_aio_request_finish(struct tcmu_device *dev, int *is_idle);
```
mikechristie: Why export these in the header? I do not think they are useful without being able to use aio_schedule right? Is it part of the rbd aio patches?

libtcmu.c
```diff
@@ -173,6 +197,433 @@ static struct tcmulib_handler *find_handler(struct tcmulib_context *ctx,
...
+	 * command (processing) completion is done when one of the
+	 * following scenario occurs:
+	 *  - synchronous handler:
+	 *	only if aio_schedule() returns an error
+	 *  - asynchronous handler:
+	 *	on an error or if the command was not asynchronously
+	 *	handled (see generic_handle_cmd(), non store callouts)
+	 */
+	if (ret != TCMU_ASYNC_HANDLED) {
+		tcmulib_command_complete(dev, cmd, ret);
```
mikechristie: On failure, aio_schedule needs to return a SAM_STAT value from scsi_defs.h. For ENOMEM type of errors I think SAM_STAT_TASK_SET_FULL is similar to what the kernel lio code uses.

libtcmu_common.h
vshankar: This is used in the current PR in the thread cleanup handler for io_work_queue thread (_untrack_in_flight_io). We have just a single worker thread atm, but if at all we move to >1 worker threads, this is useful (either way).

libtcmu.c
vshankar: well, that gave a compile time warning related to spinlocks (being volatile).

libtcmu.c
vshankar: guess that's tabs vs space. I moved everything to kernel coding style last time you commented about it -- guess this is one of the leftovers.

libtcmu.c
mikechristie: Ok ignore.

libtcmu_common.h
mikechristie: I think we misunderstood each other for part of the comment/question.

I was saying in the current patches they are only used in the aio_schedule -> io_work_queue code path in your patches. We agree on that use case.

In my original comment I was not sure if you meant to only use them there or it was going to be used in the async rbd. I see your comment you are going to use them in rbd.

Were you also going to use them in file_optical and file_example's async path in a similar way you were going to use them in your rbd patches?

vshankar: Yeh, they'll be used in async rbd handler. I agree that exporting these routines in this PR is a bit premature - I'll move these into async rbd handler PR. Also they can be used in other handlers, but for that we'd need to make `aio_schedule()` available for generic async execution. That's probably a separate PR altogether.

libtcmu.c
vshankar: Done.

libtcmu.c
vshankar: done.

libtcmu.c
vshankar: done.

vshankar:
Rebased with master and addressed comments.

mikechristie:
Patches looks ok to me. Will merge if no one speaks up today.

mikechristie:
Venky, I will fix that merge conflict manually when I merge it. Do not worry about it.

mikechristie:
Oh yeah, just in case we see missed completion events or some other completion races, we can track down that I messed up here :)

Why no locking around tcmulib_processing_complete?

For my last comment about the track helpers, I just meant that for the file_* handlers could we use the track functions like in the attached patch?

[optical-sort-of-port-to-venky-aio.txt](https://github.com/open-iscsi/tcmu-runner/files/772104/optical-sort-of-port-to-venky-aio.txt)

If so something similar could be done for file_example's async path too. Don't worry about doing this in your patches.

The question that should come from the patch is, why do the file_* handlers use a mutex around both tcmulib_command_complete and tcmulib_processing_complete, but in the patchset in this PR and my patch in this comment we only lock around tcmulib_processing_complete.

I think we do not need locking around tcmulib_processing_complete because we currently do not care what we are writing back to the kernel. We just want to do a uio operation that fires off the tcmu irq handler.  The tcmu module handles locking. We should just end up with some tcmu_handle_completions that do nothing.

vshankar:
> Oh yeah, just in case we see missed completion events or some other completion races, we can track down that I messed up here :)
> 
> Why no locking around tcmulib_processing_complete?
> 
> For my last comment about the track helpers, I just meant that for the file_* handlers could we use the track functions like in the attached patch?
> 
> optical-sort-of-port-to-venky-aio.txt
> 
> If so something similar could be done for file_example's async path too. Don't worry about doing this in your patches.

yeh. those tracking routines would be exported as part of PR #89 for use by handlers (as rbd would be the first one to use them).

> 
> The question that should come from the patch is, why do the file_* handlers use a mutex around both tcmulib_command_complete and tcmulib_processing_complete, but in the patchset in this PR and my patch in this comment we only lock around tcmulib_processing_complete.
> 
> I think we do not need locking around tcmulib_processing_complete because we currently do not care what we are writing back to the kernel. We just want to do a uio operation that fires off the tcmu irq handler. The tcmu module handles locking. We should just end up with some tcmu_handle_completions that do nothing.

That's my understanding too. The `write()` performed by `tcmulib_processing_complete()` doesn't seem to require thread safety at the userspace level at least. Also, just to mention here - `tcmulib_get_next_command()` too does not require locking as there's just a single (command processing) thread reading from the ring buffer.

vshankar:
> Venky, I will fix that merge conflict manually when I merge it. Do not worry about it.

Thanks a lot!

mikechristie:
Patches manually merged to handle the conflict (still not used to the githib GUI resolver). Thanks!

bgly:
With this patch I am seeing:

cancel_thread:174 : pthread_join failed with value 22 almost all the time when trying to delete user:fbo mappings from targetcli, which causes tcmu-runner to hang.

vshankar:
> With this patch I am seeing:
> 
> cancel_thread:174 : pthread_join failed with value 22 almost all the time when trying to delete user:fbo mappings from targetcli, which causes tcmu-runner to hang.
> 
> @vshankar @mikechristie

I'll take a look at this tomorrow. Thanks.

vshankar:
> With this patch I am seeing:
> 
> cancel_thread:174 : pthread_join failed with value 22 almost all the time when trying to delete user:fbo mappings from targetcli, which causes tcmu-runner to hang.
> 
> @vshankar @mikechristie

I cannot reproduce this w/ rbd handler.

Looks like `pthread_join` being called on the thread when another thread is already waiting to join. None of the handlers (except the example handler) spawn thread with `PTHREAD_CREATE_DETACHED` state, so this seems to be triggered by the signal handler when trying to cleanup all command processing thread.

Was tcmu-runner interrupted by any means?

bgly:
So all you need to reproduce this is to use user:fbo, map any ISO then just delete it right after you mapped it and you'll see the error on delete of the mapping. 

Another problem i see is with user:file... Same thing, map a raw file then delete the mapping... 
neo@z1403aneo:~$ sudo systemctl status tcmu-runner
● tcmu-runner.service - LIO Userspace-passthrough daemon
   Loaded: loaded (/lib/systemd/system/tcmu-runner.service; static; vendor preset: enabled)
   Active: active (running) since Wed 2017-02-22 09:27:24 CST; 23s ago
 Main PID: 11157 (tcmu-runner)
   CGroup: /system.slice/tcmu-runner.service
           └─11157 /usr/bin/tcmu-runner

Feb 22 09:27:24 z1403aneo systemd[1]: Starting LIO Userspace-passthrough daemon...
Feb 22 09:27:24 z1403aneo systemd[1]: Started LIO Userspace-passthrough daemon.
Feb 22 09:27:44 z1403aneo [11157]: remove_device:905 : could not close device fd uio0: 9
Feb 22 09:27:44 z1403aneo [11157]: remove_device:909 : could not unmap device uio0: 22

I think we need to make sure we do not break basic functionality of current handlers. We are trying to use tcmu-runner/target-cli fb in production code and we can't have stuff breaking on the fly.

bgly:
I have pushed up a patch to address these issues. https://github.com/open-iscsi/tcmu-runner/pull/100

PR #95, opened 2017-02-06
Add nonblock logger support

JefferyHu:
Add nonblock logger by using ring buffer. Log request is inserted into the rb as an array element firstly.A handler thread will fetch logs from the rb to output to syslog.
Output to stdout while log level is DEBUG is temporarily added in this patch.

Signed-off-by: Jianfei Hu <hujianfei@cmss.chinamobile.com>

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -15,12 +15,46 @@
  */
 
 #define _GNU_SOURCE
+#include <stdint.h>
+#include <pthread.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
 #include "libtcmu_log.h"
 #include "libtcmu_config.h"
 
-static int tcmu_log_level = TCMU_LOG_WARN;
+/* tcmu ring buffer for log */
+#define LOG_ENTRY_LEN 256 /* rb[0] is reserved for pri */
+#define LOG_MSG_LEN (LOG_ENTRY_LEN - 1) /* the length of the log message */
+#define LOG_ENTRYS (1024 * 32)
 
-static inline int tcmu_log_level_conf_to_syslog(int level)
+pthread_mutex_t g_mutex = PTHREAD_MUTEX_INITIALIZER;
+
+struct log_buf {
+	pthread_cond_t cond;
+	pthread_mutex_t lock;
+
+	bool thread_active;
+	int init_state;
+	bool finish_initialize;
+
+	unsigned int head;
+	unsigned int tail;
+	char buf[LOG_ENTRYS][LOG_ENTRY_LEN];
+	pthread_t thread_id;
+};
+
+static int tcmu_log_level = TCMU_LOG_WARN;
+static struct log_buf *tcmu_log_initialize(void);
+
+static struct log_buf *logbuf = NULL;
+static int initialized = false;
+
+/* covert log level from tcmu config to syslog */
+static inline int to_syslog_level(int level)
 {
 	switch (level) {
 		case TCMU_CONF_LOG_ERROR:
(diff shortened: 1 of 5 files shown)


JefferyHu:
@agrover and everyone, please help review!Thanks!

mikechristie:
Just status FYI, I am reviewing it. Should be done tomorrow.

libtcmu_log.c
```diff
@@ -42,40 +77,93 @@ unsigned int tcmu_get_log_level(void)
...
+				return;
+		len = read(fd, id, ID_MAX_LEN);
+		if (len < 0) {
+			close(fd);
+			return;
+		}
+		close(fd);
+	} else {
+		strncpy(id, ident, ID_MAX_LEN);
+	}
```
mikechristie: Do we need to make sure there is a closing '\0' in id?

libtcmu_log.c
```diff
@@ -84,18 +172,33 @@ tcmu_log_internal(int pri,
...
 
-	tcmu_log_to_syslog(pri, logbuf);
+	rb_update_head(logbuf);
+
+	if (logbuf->thread_active == false)
+		pthread_cond_signal(&logbuf->cond);
+
+	pthread_mutex_unlock(&logbuf->lock);
+
+	return;
```
mikechristie: return not needed.

libtcmu_log.c
```diff
@@ -113,14 +216,161 @@ void tcmu_info_message(const char *funcname, int linenr, const char *fmt, ...)
...
+{
+	void *join_retval;
+	int ret;
+
+	ret = pthread_cancel(thread);
+	if (ret) {
+		return;
+	}
+
+	ret = pthread_join(thread, &join_retval);
```
mikechristie: Do we need the rest of the function since there is nothing we can do.

libtcmu_log.c
```diff
@@ -113,14 +216,161 @@ void tcmu_info_message(const char *funcname, int linenr, const char *fmt, ...)
...
+
+	pthread_cleanup_push(log_thread_cleanup, arg);
+
+	open_syslog(NULL, 0, 0);
+
+	while (1) {
+		pthread_mutex_lock(&logbuf->lock);
+		if (!logbuf->finish_initialize) {
+			logbuf->finish_initialize = true;
+			pthread_cond_signal(&logbuf->cond);
```
mikechristie: Maybe move this chunk before the while since it is only needed for the initial sync up.

superlibj: While if the main thread is schduled out just after : 
Line 361 +	ret = pthread_create(&logbuf->thread_id, NULL, log_thread_start, logbuf);
and then the logger thread is schduled immediately, the cond wake up signal will be lost ?

superlibj: While if the main thread is schduled out just after : 
Line 361 +	ret = pthread_create(&logbuf->thread_id, NULL, log_thread_start, logbuf);
and then the logger thread is schduled immediately, the cond wake up signal will be lost ?

libtcmu_log.c
JefferyHu: We make the id lenth (ID_MAX_LEN + 1), and init id by"char id[ID_MAX_LEN + 1] = {0}",so after strncpy, the last chars will all remain 0.
Quato from man page, "If the length of src is less than n, strncpy() writes additional null bytes to dest to ensure that a total of n bytes are written.".
With above, we can make sure there is a closing '\0' in id.

libtcmu_log.c
JefferyHu: Fixed!

libtcmu_log.c
JefferyHu: Fixed!

libtcmu_log.c
JefferyHu: Fixed!Please review.

mikechristie:
It looks ok to me. Thanks.

I will merge on Monday to give people busy at work a chance to review over the weekend.

I wanted to make one enhancement when it is merged. If the rb gets full and we start dropping log entries, would it be a nice addition to insert a log message about dropping entries?

mikechristie:
Just to be clear, above I meant after it is merged. You do not have to worry about adding the dropping log entries warning message code.

superlibj:
For the enhancementation, output the dropped logs to one specific log file is one choice ?

mikechristie:
Merged. Thanks!

I was thinking about the same file. It is just for the case where you turn on debugging and so it is full because there is just too much info.

PR #96, opened 2017-02-06
Restart tcmu-runner on failure

seroyer:
If tcmu-runner fails for any reason, systemd should restart it.


bgly:
@agrover Please Review

PR #98, opened 2017-02-16
asynchronous compound store operation

vshankar:
Bulk of the changes are in `libtcmu_store.c` source which has all the asynchronous implementation of multi-op commands. I think given the nature of changes, there should be have more code comments, especially regarding some rules on how to implement asynchronous multi-op commands. I have an half finished doc regarding the same which I'll try to include in this PR as a future update.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -37,6 +37,7 @@
 #include "libtcmu.h"
 #include "libtcmu_log.h"
 #include "libtcmu_priv.h"
+#include "libtcmu_store.h"
 #include "tcmu-runner.h"
 
 #define ARRAY_SIZE(X) (sizeof(X) / sizeof((X)[0]))
@@ -224,31 +225,16 @@ static void cmdproc_thread_cleanup(void *arg)
 	r_handler->close(dev);
 }
 
-static int generic_handle_cmd(struct tcmu_device *dev,
-			      struct tcmulib_cmd *tcmulib_cmd)
+static int generic_cmd(struct tcmu_device *dev,
+		       struct tcmulib_cmd *tcmulib_cmd, uint8_t cmd)
 {
-	struct tcmulib_handler *handler = tcmu_get_dev_handler(dev);
-	struct tcmur_handler *store = handler->hm_private;
 	uint8_t *cdb = tcmulib_cmd->cdb;
 	struct iovec *iovec = tcmulib_cmd->iovec;
 	size_t iov_cnt = tcmulib_cmd->iov_cnt;
 	uint8_t *sense = tcmulib_cmd->sense_buf;
 	uint32_t block_size = tcmu_get_dev_block_size(dev);
 	uint64_t num_lbas = tcmu_get_dev_num_lbas(dev);
-	uint8_t cmd;
-	ssize_t ret = TCMU_NOT_HANDLED, l = tcmu_iovec_length(iovec, iov_cnt);
-	off_t offset = block_size * tcmu_get_lba(cdb);
-	struct iovec iov;
-	size_t half = l / 2;
-	uint32_t cmp_offset;
 
-	if (store->handle_cmd)
-		ret = store->handle_cmd(dev, tcmulib_cmd);
-
-	if (ret != TCMU_NOT_HANDLED)
-		return ret;
-
-	cmd = cdb[0];
 	switch (cmd) {
 	case INQUIRY:
 		return tcmu_emulate_inquiry(dev, cdb, iovec, iov_cnt, sense);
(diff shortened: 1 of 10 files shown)


libtcmu.c
```diff
@@ -277,78 +268,43 @@ static int generic_handle_cmd(struct tcmu_device *dev,
...
 	case MODE_SELECT_10:
 		return tcmu_emulate_mode_select(cdb, iovec, iov_cnt, sense);
+	default:
+		tcmu_err("unknown command %x\n", cdb[0]);
+		return TCMU_NOT_HANDLED;
+	}
+}
+
+static bool command_is_store_call(uint8_t cmd)
+{
```
liuy: How do we decide which is store call cmd? And it is overkill to choose one as store call cmd. for e.g. In this patch you make compare and write as store call cmd then all the handlers must implement compare_and_write cmd. But with passthrough mechanism, only the handler that pass down the cmd should handle it. If any one of the handler decide the cmd is a store cmd, then all the others would handle it, this break the reason we abstract out SCSI details from store drivers.

vshankar: There's no hard rule to "classify" commands as store and non-store calls, but a simple rule that I used here was - if a command needs to invoke handler calls which possibly would end up performing backend IO, these commands can be invoked via worker thread(s). Such classification is what's used in this PR to group commands as "store" calls. Non store calls are finished in the callers (command processing thread) context.

This is really not replacing the passthrough mechanism by any means. We'd need to see how PR #91 possibly benefit from this change. That said, I've not been following PR #91 closely and it has changed a bit from the last time I looked at it.

mikechristie: For the question/example about something like compare_and_write, you can make it a store callout, and handlers that do not have a native implementation can just set the compare_and_write callout to a lib function that uses read/write or if the handler's compare_and_write callout is not set just drop down the lib function.

Even if only one handler implements a handler specific compare_and_write callout, all handlers would still get the benefit of COMPARE_AND_WRITE preprocessing and error handling. For example, the SCSI offset and length and iov checks would be done in the generic code, and so would translation of errors to SAM codes.

liuy: I don't think we need a store call out abstraction if we come along with pass through mechanism. If rbd driver want to implement it on its own over the generic one, we just need to pass down it to rbd.c without affecting other handlers at all. With the call out abstraction, at least we need to call extra function in every handlers, which is not necessary for now.

vshankar: IMO, Mike's approach fits well with the async call mechanism, not that command passthrough  doesn't fit well here. The *only* extra operation with command passthrough is an extra asynchronous call to `->handle_cmd` when the command would not be handled.

mikechristie: > I don't think we need a store call out abstraction if we come along with pass through

@liuy just want to make sure I understand you.

Above do you mean in general you are opposed to these patches and you are saying we do not need the store callout abstraction in the patchset, or just specifically for compare_and_write?

liuy: > Above do you mean in general you are opposed to these patches and you are saying we do not need the store callout abstraction in the patchset, or just specifically for compare_and_write?

I'm opposed to callout abstraction in general. Why I introduce passthrough mechanism is to elimilate the need of callout abstraction of this kind. There is no necessity for callout abstraction yet.

vshankar:
Updated PR with command passthrough integration.

libtcmu.c
mikechristie: Maybe we are talking about different things, because I am confused about what the passthrough code has to do with this patchset.

These patches are laying the ground work for emulated commands that require multiple read/write calls but some handlers implement sync read/write callouts and some implement async ones. For example, glfs's read/write callouts are sync, and in Venky's PR 89 he has modified rbd's read/write callouts to use librbd's aio calls.

To support these two types of implementations for a command like WRITE_VERIFY, what is your proposal if not this patchset. For this command we want to write N bytes, then read N bytes and then compare the data to make sure it got written correctly, so libtcmu's tcmu_emulate_write_verify call needs to do a write into the handler then a read.

We could do:

1. Venky's patches here where libtcmu allows the generic write verify code to call the read or write callouts, and not worry about if they are sync or async. The store code figures it out. It assumes async and makes sync handler's calls look async.

2. The opposite of #1 and instead of making the handlers that have sync read/write callout implemenations look like async calls, we add wrappers around the handlers that have async read/write callout implementations that make them look like sync calls.

3. Not sure if this where the confusion is, but are you are saying each handler's passthrough code would just handle these types of commands? So glfs's handle_cmd would do a glfs_write/ and read for WRITE_VERIFY. rbd would do rbd_write and rbd_read.

liuy: @mikechristie I went through the patch set again and I guess I misunderstood the intention. Please go ahead and I have no objection to this patch set :)

vshankar:
I'll submit an update for this PR and #89 by tomorrow.

vshankar:
Latest push works with sync and async rbd handler (PR #89). I haven't tested it with other handlers yet (any tests is greatly appreciated).

vshankar:
@mikechristie ping?

PR #99, opened 2017-02-22
tcmu-runner: add '-D/--daemon' option support

superlibj:
option as:
-D, --daemon: make the program run as a daemon in background


liuy:
I think daemonizing should be default and only --debug mode should be on foreground. Any ideas, everyone else? @agrover @vshankar

agrover:
why? systemd should handle daemonization for it.

superlibj:
Yes, the systemd will handle it defaultly.

PR #100, opened 2017-02-22
tcmu-runner:Fix broken file and fbo handlers

bgly:
asynchronous runner core patch broken user:file and user:fbo
handlers, so this patch addresses both.

### DIFF
--- a/file_example.c
+++ b/file_example.c
@@ -42,10 +42,10 @@
 #include <pthread.h>
 
 #include "tcmu-runner.h"
+#include "libtcmu.h"
 
 #ifdef ASYNC_FILE_HANDLER
 #include <signal.h>
-#include "libtcmu.h"
 #endif
 
 #define NHANDLERS 2
@@ -87,6 +87,7 @@ file_handler_run(void *arg)
 		int result;
 		struct tcmulib_cmd *cmd;
 
+		pthread_cleanup_push(tcmulib_cleanup_mutex_lock, &h->mtx);
 		/* get next command */
 		pthread_mutex_lock(&h->mtx);
 		while (h->cmd_tail == h->cmd_head) {
@@ -94,11 +95,13 @@ file_handler_run(void *arg)
 		}
 		cmd = h->commands[h->cmd_tail];
 		pthread_mutex_unlock(&h->mtx);
+		pthread_cleanup_pop(0);
 
 		/* process command */
 		result = file_handle_cmd(h->dev, cmd);
-		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_command_complete(h->dev, cmd, result);
+
+		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_processing_complete(h->dev);
 		pthread_mutex_unlock(&state->completion_mtx);
 
@@ -132,10 +135,7 @@ file_handler_init(struct file_handler *h, struct tcmu_device *dev, int num)
 static void
 file_handler_destroy(struct file_handler *h)
 {
-	if (h->thr) {
-		pthread_kill(h->thr, SIGINT);
-		pthread_join(h->thr, NULL);
-	}
+	tcmulib_cancel_thread(h->thr);
 	pthread_cond_destroy(&h->cond);
 	pthread_mutex_destroy(&h->mtx);
 }
@@ -258,6 +258,7 @@ static int file_handle_cmd_async(
 
 	state->curr_handler = (state->curr_handler + 1) % NHANDLERS;
 
+	pthread_cleanup_push(tcmulib_cleanup_mutex_lock, &h->mtx);
 	/* enqueue command */
 	pthread_mutex_lock(&h->mtx);
 	while ((h->cmd_head + 1) % NCOMMANDS == h->cmd_tail) {
@@ -267,6 +268,7 @@ static int file_handle_cmd_async(
 	h->cmd_head = (h->cmd_head + 1) % NCOMMANDS;
 	pthread_cond_signal(&h->cond);
 	pthread_mutex_unlock(&h->mtx);
+	pthread_cleanup_pop(0);
 
 	return TCMU_ASYNC_HANDLED;
 }
@@ -331,13 +333,14 @@ static int do_verify_op(struct file_state *state, struct iovec *iovec, uint64_t 
 		return tcmu_set_sense_data(sense, HARDWARE_ERROR,
 					   ASC_INTERNAL_TARGET_FAILURE, NULL);
 
+	pthread_cleanup_push(tcmulib_cleanup_malloc, buf);
 	memset(buf, 0, length);
 
 	ret = pread(state->fd, buf, length, offset);
 	if (ret < 0) {
 		tcmu_err("read failed: %m\n");
-		free(buf);
-		return set_medium_error(sense);
+		rc = set_medium_error(sense);
+		goto cleanup;
 	}
 
 	cmp_offset = tcmu_compare_with_iovec(buf, iovec, length);
@@ -347,7 +350,9 @@ static int do_verify_op(struct file_state *state, struct iovec *iovec, uint64_t 
 		tcmu_err("Verify failed at offset %lu\n", cmp_offset);
 	}
 
+cleanup:
 	free(buf);
+	pthread_cleanup_pop(0);
 
 	return rc;
 }
@@ -391,7 +396,7 @@ static int file_handle_cmd(
 	uint64_t offset;
 	int length = 0;
 	uint64_t cur_lba = 0;
-	int rc;
+	int rc = SAM_STAT_GOOD;
 
 	cmd = cdb[0];
 
@@ -450,20 +455,24 @@ static int file_handle_cmd(
 		buf = malloc(length);
 		if (!buf)
 			return set_medium_error(sense);
+		pthread_cleanup_push(tcmulib_cleanup_malloc, buf);
 		memset(buf, 0, length);
 
 		ret = pread(state->fd, buf, length, offset);
 		if (ret == -1) {
 			tcmu_err("read failed: %m\n");
-			free(buf);
-			return set_medium_error(sense);
+			rc = set_medium_error(sense);
+			goto cleanup;
 		}
 
 		tcmu_memcpy_into_iovec(iovec, iov_cnt, buf, length);
+		rc = SAM_STAT_GOOD;
+		
+		cleanup:
+			free(buf);
+			pthread_cleanup_pop(0);
 
-		free(buf);
-
-		return SAM_STAT_GOOD;
+			return rc;
 	}
 	break;
 	case WRITE_VERIFY:
--- a/file_optical.c
+++ b/file_optical.c
@@ -121,12 +121,14 @@ static void *fbo_handler_run(void *arg)
 	struct tcmulib_cmd *cmd;
 
 	while (true) {
+		pthread_cleanup_push(tcmulib_cleanup_mutex_lock, &h->mtx);
 		/* get next command */
 		pthread_mutex_lock(&h->mtx);
 		while (h->cmd_tail == h->cmd_head)
 			pthread_cond_wait(&h->cond, &h->mtx);
 		cmd = h->commands[h->cmd_tail];
 		pthread_mutex_unlock(&h->mtx);
+		pthread_cleanup_pop(0);
 
 		/* process command */
 		result = fbo_handle_cmd(h->dev, cmd);
@@ -134,12 +136,14 @@ static void *fbo_handler_run(void *arg)
                     tcmu_dbg("FBO: Check condition\n");
                     tcmu_dbg("  Sense: 0x%08llx\n", *(uint64_t *)cmd->sense_buf);
                 }
-		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_command_complete(h->dev, cmd, result);
+
+		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_processing_complete(h->dev);
 		pthread_mutex_unlock(&state->completion_mtx);
 
 		/* notify that we can process more commands */
+
 		pthread_mutex_lock(&h->mtx);
 		h->commands[h->cmd_tail] = NULL;
 		h->cmd_tail = (h->cmd_tail + 1) % NCOMMANDS;
@@ -168,10 +172,7 @@ static void fbo_handler_init(struct fbo_handler *h, struct tcmu_device *dev,
 
 static void fbo_handler_destroy(struct fbo_handler *h)
 {
-	if (h->thr) {
-		pthread_kill(h->thr, SIGINT);
-		pthread_join(h->thr, NULL);
-	}
+	tcmulib_cancel_thread(h->thr);
 	pthread_cond_destroy(&h->cond);
 	pthread_mutex_destroy(&h->mtx);
 }
@@ -350,6 +351,8 @@ static int fbo_handle_cmd_async(struct tcmu_device *dev,
 
 	state->curr_handler = (state->curr_handler + 1) % NHANDLERS;
 
+	pthread_cleanup_push(tcmulib_cleanup_mutex_lock, &h->mtx);
+
 	/* enqueue command */
 	pthread_mutex_lock(&h->mtx);
 	while ((h->cmd_head + 1) % NCOMMANDS == h->cmd_tail)
@@ -358,6 +361,7 @@ static int fbo_handle_cmd_async(struct tcmu_device *dev,
 	h->cmd_head = (h->cmd_head + 1) % NCOMMANDS;
 	pthread_cond_signal(&h->cond);
 	pthread_mutex_unlock(&h->mtx);
+	pthread_cleanup_pop(0);
 
 	return TCMU_ASYNC_HANDLED;
 }
@@ -1272,7 +1276,7 @@ static int fbo_read(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	int length = 0;
 	void *buf;
 	size_t ret;
-	int rc;
+	int rc = SAM_STAT_GOOD;
 
 	// TBD: If we simulate start/stop, then fail if stopped
 	/* DPO and RelAdr bits should be 0 */
@@ -1297,6 +1301,8 @@ static int fbo_read(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	buf = malloc(length);
 	if (!buf)
 		return set_medium_error(sense);
+
+	pthread_cleanup_push(tcmulib_cleanup_malloc, buf);
 	memset(buf, 0, length);
 
 	pthread_mutex_lock(&state->state_mtx);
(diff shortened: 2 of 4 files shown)


bgly:
Please review @vshankar @mikechristie @agrover

file_example.c
```diff
@@ -97,8 +97,8 @@ file_handler_run(void *arg)
 
 		/* process command */
 		result = file_handle_cmd(h->dev, cmd);
-		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_command_complete(h->dev, cmd, result);
+		pthread_mutex_lock(&state->completion_mtx);
 		tcmulib_processing_complete(h->dev);
 		pthread_mutex_unlock(&state->completion_mtx);
```
mikechristie: Thanks for fixing and debugging bgly.

Was the locking call change in the completion paths here and below related to the bug, or just a change because  libtcmu's tcmulib_command_complete locks for the handler now?

bgly: It's related to the bug, but with larger scale testing this patch does not fix all the problems, so hold off on merging please. 

The problem I now see is if you create multiple mappings and then delete the mappings it will cause tcmu-runner to hang which needs a restart to be able to recover. The key thing that causes that breakage is the deleting of multiple mappings... If you delete one at a time and exit targetcli each time it is fine...

bgly: Actually if you just have multiple mappings then you delete one mapping while having multiple user:fbo or user:file_async mapped then tcmu-runner will crash.  The odd thing is that the tcmu-runner service will still show active, but if you open up targetcli after the hang then you wont see any tcmu backstores listed, which indicates that tcmu-runner prob crashed 

user:file on the other hand if you have just one mapped and then do a delete it wont show the:

Feb 22 13:04:10 z1403aneo [5778]: remove_device:905 : could not close device fd uio0: 9
Feb 22 13:04:10 z1403aneo [5778]: remove_device:909 : could not unmap device uio0: 22

error, but if you have multiple user:file mapped then delete one mapping it will post that error. 

This is with this patch running, if you dont have this patch running then doing just one mapping and one delete will cause all these problems....

bgly:
@mikechristie 

With the latest patch user:file, user:file_async, and user:fbo seems to be working. 

I have done a bunch of deploys through openstack which would do multiple mappings and simultaneous deletes, and no hanging at all.

Please review @mikechristie @agrover

libtcmu.c
```diff
@@ -913,6 +912,7 @@ static void remove_device(struct tcmulib_context *ctx,
 	if (ret < 0) {
 		tcmu_err("could not cleanup mailbox lock %s: %d\n", dev_name, errno);
 	}
+	dev->handler->removed(dev);
```
bgly: https://github.com/open-iscsi/tcmu-runner/pull/69

This patch above moved dev->handler->removed(dev); to end to fix freeing invalid fd.

file_optical.c
```diff
@@ -168,10 +168,6 @@ static void fbo_handler_init(struct fbo_handler *h, struct tcmu_device *dev,
 
 static void fbo_handler_destroy(struct fbo_handler *h)
 {
-	if (h->thr) {
-		pthread_kill(h->thr, SIGINT);
-		pthread_join(h->thr, NULL);
-	}
 	pthread_cond_destroy(&h->cond);
```
bgly: We dont need these anymore because cancel_thread is called and when you call dev->handler->removed(dev); that also calls cancel_thread

libtcmu.c
vshankar: Wouldn't it be a better fix if `free(dev)` is called in `remove_device()` rather than in the thread cleanup handler? The thread cleanup handler would then just call `r_handler->close()`.

The point of moving `dev->handler->removed` before closing `dev->fd` is to cleanup any users of `dev->fd` before closing`fd`.

bgly: That works too, ill move that

file_optical.c
mikechristie: I am not sure I understood this comment.

The cancel_thread call in libtcmu will only handle threads allocated by libtcmu right? It will not handle threads created by file_* will it?

bgly:
@mikechristie Yes, you are right, I overlooked the cancel_thread. The threads need to also be canceled within the file_* handlers. otherwise there is a memory leak. Its not correct to use pthread_kill(SIGINT) which pretty much acts like ctrl-c, and what if a thread is blocked on i/o? Anyways its easier to use pthread_cancel, which is used in the aio stuff.

file_optical.c
```diff
@@ -168,10 +168,7 @@ static void fbo_handler_init(struct fbo_handler *h, struct tcmu_device *dev,
 
 static void fbo_handler_destroy(struct fbo_handler *h)
 {
-	if (h->thr) {
-		pthread_kill(h->thr, SIGINT);
-		pthread_join(h->thr, NULL);
-	}
+	tcmulib_cancel_thread(h->thr);
```
seroyer: I'm concerned about using pthread_cancel given that pthread_mutex is used.  If there is any chance of a thread being canceled while holding the mutex, you must register a cleanup handler to release the mutex if there are any thread cancellation points in the critical section.

bgly: I went through all the cancellation points and cleaned up

seroyer: It looks like you handled the cases correct.  I have also tested with this.

Tested-by: Steven Royer seroyer@linux.vnet.ibm.com

mikechristie:
Which part of the patch is needed for the fix and which is converting the handlers to use some of the new libctmu threading code or they intertwined?  Was it just the cmdproc_thread_cleanup chunk and remove_device chunk?

@vshankar do you have patches to support more than 1 thread in setup_io_work_queue? If so, we can just drop the threading from these handlers like in this patch for file_example.c.

[file-drop-thread.txt](https://github.com/open-iscsi/tcmu-runner/files/819186/file-drop-thread.txt)

vshankar:
> Which part of the patch is needed for the fix and which is converting the handlers to use some of the new libctmu threading code or they intertwined? Was it just the cmdproc_thread_cleanup chunk and remove_device chunk?
> 
> @vshankar do you have patches to support more than 1 thread in setup_io_work_queue? If so, we can just drop the threading from these handlers like in this patch for file_example.c.

As of now, no. It should be straightforward to make file/fbo handler use threading code in `libtcmu`. The necessary routines in `libtcmu` are thread safe.

bgly:
It is all intertwined now... More or less the cmdproc_thread_cleanup chunk and remove_device chunk will fix the handlers not working... But this patch also addresses the way the handler is destroyed. We have seen messages where SIGINT is called and tcmu-runner is in a bad state, so this also addresses not calling "pthread_kill(h->thr, SIGINT);".

mikechristie:
Just to make sure I am on the same page. There are 2 issues not related to each other.

1. We need to move free(dev) to the end of remove_device() because cmdproc_thread_cleanup() was freeing the device too early while remove_device was still accessing it. This is a regression.

2. The file_*'s  pthread_kill SIGINT call was firing sighandler and the exit() call in there causes tcmu-runner process to exit.

This bug existed before the async patches. With your patch, file_* now uses all cancels, so tcmu-runner survives.

bgly:
Correct

mikechristie:
Could you test and review by branch here

https://github.com/mikechristie/tcmu-runner/tree/make-common

It converts qcow and file_example to use the core runner threading and also the read/write/flush callouts. And it converts file_optical to use the core runner threading.

I did some light testing with file_example and rbd so far.

bgly:
On 3/14/17 6:51 PM, mikechristie wrote:
>
> Could you test and review by branch here
>
> https://github.com/mikechristie/tcmu-runner/tree/make-common
>
> It converts qcow and file_example to use the core runner threading and 
> also the read/write/flush callouts. And it converts file_optical to 
> use the core runner threading.
>
> I did some light testing with file_example and rbd so far.
>
Hi Mike,

So I looked over some of the code changes, and I'm still testing but you will need to some stuff that I've attached.

I was able to get Ubuntu booted with qcow and file handlers, and tested fbo, which seems to be fine.

Also within the FBO handler can you use 1 thread instead of 16?

The only thing that I noticed that is broken is AIX clients can no longer boot off of any of the handlers... Still debugging.

-Bryant



From d8e34a2301915801b15d8dd6d0c3261283d329bc Mon Sep 17 00:00:00 2001
From: "Bryant G. Ly" <bryantly@linux.vnet.ibm.com>
Date: Wed, 15 Mar 2017 10:49:51 -0500
Subject: [PATCH] tcmu-runner: Add setting blocksize and num_lbas

---
 file_optical.c | 3 +++
 qcow.c         | 1 +
 2 files changed, 4 insertions(+)

diff --git a/file_optical.c b/file_optical.c
index 2f63b38..92a7798 100644
--- a/file_optical.c
+++ b/file_optical.c
@@ -174,9 +174,11 @@ static int fbo_open(struct tcmu_device *dev)
 		tcmu_err("Could not get device block size\n");
 		goto err;
 	}
+	tcmu_set_dev_block_size(dev, state->block_size);
 #else
 	/* MM logical units use a block size of 2048 */
 	state->block_size = 2048;
+	tcmu_set_dev_block_size(dev, state->block_size);
 #endif
 
 	size = tcmu_get_device_size(dev);
@@ -185,6 +187,7 @@ static int fbo_open(struct tcmu_device *dev)
 		goto err;
 	}
 
+	tcmu_set_dev_num_lbas(dev, size / state->block_size);
 	state->num_lbas = size / state->block_size;
 
 	tcmu_dbg("open: cfgstring %s\n", tcmu_get_dev_cfgstring(dev));
diff --git a/qcow.c b/qcow.c
index 89472b9..fe77e15 100644
--- a/qcow.c
+++ b/qcow.c
@@ -1435,6 +1435,7 @@ static int qcow_open(struct tcmu_device *dev)
 		goto err;
 	}
 
+	tcmu_set_dev_num_lbas(dev, bdev->size / bdev->block_size);
 	bdev->num_lbas = bdev->size / bdev->block_size;
 
 	config = strchr(tcmu_get_dev_cfgstring(dev), '/');
-- 
2.7.4

mikechristie:
Bgly, I merged the regression part of the patch in this commit:

https://github.com/open-iscsi/tcmu-runner/commit/1e349533716996f9ca525d5aa00846569a8e48d5

I have a update to the patches to covert file_example to the core threading and fix file_optical. I will post that this week.

mikechristie:
@bgly An updated branch with fixes for file_*'s shutdown are here:

https://github.com/mikechristie/tcmu-runner/tree/make-common

I did a push -f so make sure you do a new clone/pull.

bgly:
Tested and works!

bgly:
On 3/25/17 6:04 PM, mikechristie wrote:
>
> @bgly <https://github.com/bgly> An updated branch with fixes for 
> file_*'s shutdown are here:
>
> https://github.com/mikechristie/tcmu-runner/tree/make-common
>
> I did a push -f so make sure you do a new clone/pull.
>
>
Hi Mike,

I reviewed your code and with some of the changes it looks like you have to fix the fbo_write_verify when you switched to preadv/pwritev.
I have attached some untested code that I wrote up real quick. It should work.

-bryant



From a3fe8e41dc755ae5e0910571d9d5cb842e1bef7a Mon Sep 17 00:00:00 2001
From: "Bryant G. Ly" <bryantly@linux.vnet.ibm.com>
Date: Thu, 30 Mar 2017 17:58:03 -0500
Subject: [PATCH] file-optical: Fix write_verify for fbo

---
 file_optical.c | 61 +++++++++++++++++++++++++++++++++++++++-------------------
 1 file changed, 41 insertions(+), 20 deletions(-)

diff --git a/file_optical.c b/file_optical.c
index b9af367..0e34961 100644
--- a/file_optical.c
+++ b/file_optical.c
@@ -1164,6 +1164,7 @@ static int fbo_read(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	uint64_t cur_lba = 0;
 	uint64_t offset;
 	int length = 0;
+	int remaining;
 	size_t ret;
 	int rc;
 
@@ -1192,17 +1193,23 @@ static int fbo_read(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	state->flags |= FBO_DEV_IO;
 	pthread_mutex_unlock(&state->state_mtx);
 
-	ret = preadv(state->fd, iovec, iov_cnt, offset);
+	remaining = length;
+
+	while (remaining) {
+		ret = preadv(state->fd, iovec, iov_cnt, offset);
+		if (ret < 0) {
+			tcmu_err("read failed: %m\n");
+			return set_medium_error(sense, ASC_READ_ERROR);
+		}
+		tcmu_seek_in_iovec(iovec, ret);
+		offset += ret;
+		remaining -= ret;
+	}
 
 	pthread_mutex_lock(&state->state_mtx);
 	state->flags &= ~FBO_DEV_IO;
 	pthread_mutex_unlock(&state->state_mtx);
 
-	if (ret == -1) {
-		tcmu_err("read failed: %m\n");
-		return set_medium_error(sense, ASC_READ_ERROR);
-	}
-
 	return SAM_STAT_GOOD;
 }
 
@@ -1212,14 +1219,17 @@ static void fbo_cleanup_buffer(void *buf)
 }
 
 static int fbo_do_verify(struct fbo_state *state, struct iovec *iovec,
-			 uint64_t offset, int length, uint8_t *sense)
+			 size_t iov_cnt, uint64_t offset, int length,
+			 uint8_t *sense)
 {
+	struct iovec read_iov;
 	size_t ret;
 	uint32_t cmp_offset;
 	void *buf;
 	int rc = SAM_STAT_GOOD;
+	int remaining;
 
-	buf = malloc(length);
+	read_iov.iov_base = buf = malloc(length);
 	if (!buf)
 		return tcmu_set_sense_data(sense, HARDWARE_ERROR,
 					   ASC_INTERNAL_TARGET_FAILURE, NULL);
@@ -1227,24 +1237,32 @@ static int fbo_do_verify(struct fbo_state *state, struct iovec *iovec,
 	pthread_cleanup_push(fbo_cleanup_buffer, buf);
 	memset(buf, 0, length);
 
+	read_iov.iov_len = length;
+
 	pthread_mutex_lock(&state->state_mtx);
 	state->cur_lba = offset / state->block_size;
 	state->flags |= FBO_DEV_IO;
 	pthread_mutex_unlock(&state->state_mtx);
 
-	ret = pread(state->fd, buf, length, offset);
+	remaining = length;
+
+	while (remaining) {
+		ret = preadv(state->fd, &read_iov, iov_cnt, offset);
+		if (ret < 0) {
+			tcmu_err("read failed: %m\n");
+			free(buf);
+			rc = set_medium_error(sense, ASC_READ_ERROR);
+			goto cleanup;
+		}
+		tcmu_seek_in_iovec(&read_iov, ret);
+		offset += ret;
+		remaining -= ret;
+	}
 
 	pthread_mutex_lock(&state->state_mtx);
 	state->flags &= ~FBO_DEV_IO;
 	pthread_mutex_unlock(&state->state_mtx);
 
-	if (ret == -1) {
-		tcmu_err("read failed: %m\n");
-		free(buf);
-		rc = set_medium_error(sense, ASC_READ_ERROR);
-		goto cleanup;
-	}
-
 	cmp_offset = tcmu_compare_with_iovec(buf, iovec, length);
 	if (cmp_offset != -1)
 		rc = tcmu_set_sense_data(sense, MISCOMPARE,
@@ -1253,7 +1271,7 @@ static int fbo_do_verify(struct fbo_state *state, struct iovec *iovec,
 
 cleanup:
 	free(buf);
-	 pthread_cleanup_pop(0);
+	pthread_cleanup_pop(0);
 
 	return rc;
 }
@@ -1262,6 +1280,7 @@ static int fbo_write(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 		     size_t iov_cnt, uint8_t *sense, bool do_verify)
 {
 	struct fbo_state *state = tcmu_get_dev_private(dev);
+	struct iovec write_iovec[iov_cnt];
 	uint8_t fua = cdb[1] & 0x08;
 	uint64_t cur_lba = 0;
 	uint64_t offset;
@@ -1290,8 +1309,10 @@ static int fbo_write(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 
 	remaining = length;
 
+	*write_iovec = *iovec;
+
 	while (remaining) {
-		ret = pwritev(state->fd, iovec, iov_cnt, offset);
+		ret = pwritev(state->fd, write_iovec, iov_cnt, offset);
 		if (ret < 0) {
 			tcmu_err("write failed: %m\n");
 			rc = set_medium_error(sense, ASC_WRITE_ERROR);
@@ -1316,7 +1337,7 @@ static int fbo_write(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	if (!do_verify || rc != SAM_STAT_GOOD)
 		return rc;
 
-	return fbo_do_verify(state, iovec, offset, length, sense);
+	return fbo_do_verify(state, iovec, iov_cnt, offset, length, sense);
 }
 
 static int fbo_verify(struct tcmu_device *dev, uint8_t *cdb,
@@ -1345,7 +1366,7 @@ static int fbo_verify(struct tcmu_device *dev, uint8_t *cdb,
 
 	offset = state->block_size * cur_lba;
 
-	return fbo_do_verify(state, iovec, offset, length, sense);
+	return fbo_do_verify(state, iovec, iov_cnt, offset, length, sense);
 }
 
 static int fbo_do_format(struct tcmu_device *dev, uint8_t *sense)
-- 
2.7.4

bgly:
On 03/30/2017 06:03 PM, Bryant G. Ly wrote:
> 
> 
> On 3/25/17 6:04 PM, mikechristie wrote:
>>
>> @bgly <https://github.com/bgly> An updated branch with fixes for
>> file_*'s shutdown are here:
>>
>> https://github.com/mikechristie/tcmu-runner/tree/make-common
>>
>> I did a push -f so make sure you do a new clone/pull.
>>
>>
> Hi Mike, 
> 
> I reviewed your code and with some of the changes it looks like you have to fix the fbo_write_verify when you switched to preadv/pwritev.
> I have attached some untested code that I wrote up real quick. It should work. 
> 

Thanks for the review and patch.

One other item, I also forgot to fix from your other review comments was
reducing nr_threads to 1 for file_optical. I will fix that too.

bgly:
> On 03/30/2017 06:03 PM, Bryant G. Ly wrote:
>>
>> On 3/25/17 6:04 PM, mikechristie wrote:
>>> @bgly <https://github.com/bgly> An updated branch with fixes for
>>> file_*'s shutdown are here:
>>>
>>> https://github.com/mikechristie/tcmu-runner/tree/make-common
>>>
>>> I did a push -f so make sure you do a new clone/pull.
>>>
>>>
>> Hi Mike,
>>
>> I reviewed your code and with some of the changes it looks like you have to fix the fbo_write_verify when you switched to preadv/pwritev.
>> I have attached some untested code that I wrote up real quick. It should work.
>>
> Thanks for the review and patch.
>
> One other item, I also forgot to fix from your other review comments was
> reducing nr_threads to 1 for file_optical. I will fix that too.
>
Sounds good!

seroyer:
@mikechristie 

Found another issue with the file_optical write_verify.  Here is the fix.

fbo: reset offset on write_verify

The offset in write_verify was getting incremented during the write, but
then used again for the verify.  Need to reset offset back to the
beginning of the op.

Signed-off-by: Mike Cyr <mikecyr@us.ibm.com>
---
 file_optical.c | 1 +
 1 file changed, 1 insertion(+)

diff --git a/file_optical.c b/file_optical.c
index 0e34961..8563106 100644
--- a/file_optical.c
+++ b/file_optical.c
@@ -1337,6 +1337,7 @@ static int fbo_write(struct tcmu_device *dev, uint8_t *cdb, struct iovec *iovec,
 	if (!do_verify || rc != SAM_STAT_GOOD)
 		return rc;
 
+	offset = state->block_size * cur_lba;
 	return fbo_do_verify(state, iovec, iov_cnt, offset, length, sense);
 }
 
--

bgly:
@mikechristie Btw the patch @seroyer posted is a based off of your make-common branch/the fixes I gave you before.

mikechristie:
Thanks.

I am just integrating all the outstanding patches right now. Venky's 89 and 98 had lots of conflicts with this patchset. I am hoping to get all outstanding merges done in the next couple of days.

mikechristie:
@bgly and @seroyer, patches were merged in PR:

https://github.com/open-iscsi/tcmu-runner/pull/109

Note that I did some file_optical testing and did some other fixes on top of your guy's patches to support scatterlist/iovecs with more than one element.

PR #101, opened 2017-03-01
tcmu-runner: write verify fix

bgly:
This patch addresses tcmu_write_verify not working properly.
The caller needs to pass the length as a parm, otherwise the function doesnt
work properly, and the length will be 0. Thus the code does not run.

### DIFF
--- a/qcow.c
+++ b/qcow.c
@@ -1426,6 +1426,7 @@ static int qcow_open(struct tcmu_device *dev)
 		tcmu_err("Could not get device block size\n");
 		goto err;
 	}
+	tcmu_set_dev_block_size(dev, bdev->block_size);
 
 	bdev->size = tcmu_get_device_size(dev);
 	if (bdev->size < 0) {


bgly:
Please review @mikechristie 

I overlooked the last merge of this since I figured the patch would just work since we just abstracted the tcmu_write_verify from being within qcow to be an api... So I didn't test it, but I had some time to go back to looking at qcow stuff and so I found the bug, and have fixed it/verified the fix. 

Thanks!

seroyer:
Tested-by: Steven Royer seroyer@linux.vnet.ibm.com

mikechristie:
How about just having qcow call tcmu_set_dev_block_size like the other handlers that use lib calls.

bgly:
That works, too! Ive tested it and I will clean up bdev->size calls in a new patch, where I address the ability to have resizable tcmu-backstores... Mainly user:file and user:qcow(raw).

PR #102, opened 2017-03-09
user:file/user:qcow - Add resize capability

bgly:
This patch allows for one to use qemu-img resize or truncate
to resize raw file types.

### DIFF
--- a/file_example.c
+++ b/file_example.c
@@ -386,14 +386,23 @@ static int file_handle_cmd(
 	int remaining;
 	size_t ret;
 	uint32_t block_size = tcmu_get_dev_block_size(dev);
-	uint64_t num_lbas = tcmu_get_dev_num_lbas(dev);
+	uint64_t num_lbas;
 	bool do_verify = false;
 	uint64_t offset;
 	int length = 0;
 	uint64_t cur_lba = 0;
 	int rc;
+	int64_t size;
 
 	cmd = cdb[0];
+
+	size = tcmu_get_device_size(dev);
+	if (size < 0) {
+		tcmu_err("Could not get device size\n");
+		return TCMU_NOT_HANDLED;
+	}
+	tcmu_set_dev_num_lbas(dev, size / block_size);
+	num_lbas = tcmu_get_dev_num_lbas(dev);
 
 	switch (cmd) {
 	case INQUIRY:
--- a/qcow.c
+++ b/qcow.c
@@ -1498,8 +1498,17 @@ static int qcow_handle_cmd(
 	uint8_t cmd;
 	ssize_t ret;
 	uint64_t offset = bdev->block_size * tcmu_get_lba(cdb);
+	uint64_t size;
+	uint64_t num_lbas;
 
 	cmd = cdb[0];
+
+	size = tcmu_get_device_size(dev);
+	if (size < 0) {
+		tcmu_err("Could not get device size\n");
+		return TCMU_NOT_HANDLED;
+	}
+	num_lbas = size / bdev->block_size;
 
 	switch (cmd) {
 	case INQUIRY:
@@ -1510,7 +1519,7 @@ static int qcow_handle_cmd(
 		break;
 	case SERVICE_ACTION_IN_16:
 		if (cdb[1] == READ_CAPACITY_16)
-			return tcmu_emulate_read_capacity_16(bdev->num_lbas,
+			return tcmu_emulate_read_capacity_16(num_lbas,
 							     bdev->block_size,
 							     cdb, iovec,
 							     iov_cnt, sense);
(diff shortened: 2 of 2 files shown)


bgly:
This is just an extension of what @agrover did awhile back. https://github.com/open-iscsi/tcmu-runner/pull/60

mikechristie:
How do you tell the initiator that the size has changed?

We also need to fire off a UA (06/2A/09 - CAPACITY DATA HAS CHANGED), so the initiator sends a READ_CAPICITY to get the new size.

bgly:
Well I have been doing what they did in the previous posts...

I do this to update configfs
echo 'dev_size=10737418240' > /sys/kernel/config/target/core/user_0/TEST/control

Then I run on the client to update the sizes... 
 echo 1 > /sys/bus/scsi/devices/0\:0\:0\:3/rescan

But you are saying we can just fire a unit attention?

bgly:
```
neo@z1403aneo:~/tcmu-runner$ git diff file_example.c
diff --git a/file_example.c b/file_example.c
index 7a7f8c2..f1a4e3e 100644
--- a/file_example.c
+++ b/file_example.c
@@ -386,14 +386,32 @@ static int file_handle_cmd(
        int remaining;
        size_t ret;
        uint32_t block_size = tcmu_get_dev_block_size(dev);
-       uint64_t num_lbas = tcmu_get_dev_num_lbas(dev);
+       uint64_t num_lbas;
        bool do_verify = false;
        uint64_t offset;
        int length = 0;
        uint64_t cur_lba = 0;
        int rc;
+       int64_t size;

        cmd = cdb[0];
+
+       if( cmd != INQUIRY && cmd != REQUEST_SENSE && cmd != REPORT_LUNS) {
+               size = tcmu_get_device_size(dev);
+               if (size < 0) {
+                       tcmu_err("Could not get device size\n");
+                       return TCMU_NOT_HANDLED;
+               }
+               num_lbas = tcmu_get_dev_num_lbas(dev);
+               if (num_lbas != size / block_size) {
+                       tcmu_err("Device size changed\n");
+                       tcmu_set_dev_num_lbas(dev, size / block_size);
+                       num_lbas = tcmu_get_dev_num_lbas(dev);
+                       return tcmu_set_sense_data(sense, UNIT_ATTENTION, ASC_CAPACITY_DATA_HAS_CHANGED, NULL);
+               }
+       }

        switch (cmd) {
        case INQUIRY:
```

So I added the UA of 06/2A/09 and it seems like the Linux initiator doesn't send back a Read Capacity. When I update the configfs with that echo command, I see on the initiator:
[ 2413.921268] sd 0:0:0:3: Capacity data has changed
root@resizetest:~# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sda      8:0    0    3G  0 disk
├─sda1   8:1    0    8M  0 part
└─sda2   8:2    0    4G  0 part /
sr0     11:0    1  450K  0 rom

But on the target I ran:

neo@z1403aneo:~$ qemu-img resize -f raw Ubuntu_prox.img 4G
Image resized.
root@z1403aneo:/home/neo# echo 'dev_size=4294967296' > /sys/kernel/config/target/core/user_2/UBUNTU/control

For the initiator to see the changes I had to run:

root@resizetest:~# echo 1 > /sys/bus/scsi/devices/0\:0\:0\:3/rescan
root@resizetest:~# lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sda      8:0    0    4G  0 disk
├─sda1   8:1    0    8M  0 part
└─sda2   8:2    0    4G  0 part /
sr0     11:0    1  450K  0 rom
[ 2541.232751] sd 0:0:0:3: [sda] 8388608 512-byte logical blocks: (4.29 GB/4.00 GiB)
[ 2541.233307] sda: detected capacity change from 3221225472 to 4294967296

mikechristie:
In windows and solaris, it should just work. I am not sure about ESX.

For RHEL, you have to manually enable it by installing the libstoragemgmt-udev rpm and modifying 90-scsi-ua.rules to run the rescan commands you want to run. I am not sure how it is handled on other distros. Basically, the kernel sends an event to userspace (see drivers/scsi/scsi_error.c:scsi_report_sense()), and you have some daemon/listener handle it.

mikechristie:
@bgly, for now do not worry about the UAs. Patch looks ok. Will finish reviewing it later today.

mikechristie:
Hey,

I do not think we want to add a tcmu_get_device_size call in the main IO path. What do you think about tcmu-runner monitoring the lio configfs info file for status changes in the background? I do not think we can currently just do a monitor on a configfs file, but I can look into it.

mikechristie:
Ignore comment above. Too much ham and pie :)

We can just add a new netlink event TCMU_CMD_RECONFIG_DEVICE.  When a value like size is updated then in tcmu_set_configfs_dev_params we can fire the new event and tcmu-runner can revalidate its settings.

bgly:
yeah, that will work.

PR #106, opened 2017-03-21
logger: add tcmu_dbg_cdb() helper support

superlibj:
For the CDB debug messages, they will fill the logger buffer
full very quickly and will overwrite all the other err/warn/debug
messages.

Here just output the CDB debuginfo output to the stderr.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -405,7 +405,7 @@ static void tcmu_cdb_debug_info(const struct tcmulib_cmd *cmd)
 	}
 	sprintf(buf + n, "\n");
 
-	tcmu_dbg(buf);
+	tcmu_dbg_scsi_cmd(buf);
 
 	if (bytes > CDB_FIX_SIZE)
 		free(buf);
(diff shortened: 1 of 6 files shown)


mikechristie:
What about the other messages that are or will be added in the main IO path?

Also stderr will have the same problem too right?

superlibj:
For now if the tcmu-runner logger is in debug level, the syslog and journal buffer will be filled almost full with CDB debug messages, and it a bit hard to filter other messages.

For the other debug messages, they will still be output to stderr and at the same time will still to the logger ring. And only for the CDB debug messasges, they will output to the stderr only.

Though the stderr will filled full with the CDB debug messages, but we could still get the other debug messages from the syslog message.

mikechristie:
Is the issue more that you, and most people, do not care about the low level SCSI cdb info? It seems like it will be a common case where normally we do not need that low level info even though we want to get debug info.

If so, then I think it is better to add a new log level SCSI_CMD_DEBUG that when set will print info like the cdb. It will normally not be used and probably only used to debug protocol level problems like interop issues with initiators or debugging issues with validation tools like windows clustering's tool.

superlibj:
Agreed, and I have update it.

For the system config, here add a new and higher log level XX_DEBUG_SCSI_CMD to control and print the cdb info. While for the current -d parameter, the default log level will be XX_DEBUG_SCSI_CMD.

mikechristie:
Patch looks ok to me. Thanks.  Will merge this week. Trying to get Venky's patches in right now. There was one minor/fuzz conflict with his patches. I will just fix that up for you when I merge it.

superlibj:
Thanks.

mikechristie:
Thanks.

PR #108, opened 2017-04-13
runner: Add RESERVE and RELEASE cmds handle support.

superlibj:
For the EXSI(such as v5.5), when discovering and mounting the
targets, the runner will fail to handle the RESERVE cmd and then
lead the EXSI failed with I/O errors.

Here just set the sense key to NO_SENSE and return SAM_STAT_GOOD.

### DIFF
--- a/api.c
+++ b/api.c
@@ -731,6 +731,16 @@ static struct {
 };
 
 /*
+ * Handle RESERVE and RELEASE.
+ */
+int tcmu_emulate_reserve_release(uint8_t *sense)
+{
+	tcmu_set_sense_data(sense, NO_SENSE, 0, NULL);
+
+	return SAM_STAT_GOOD;
+}
+
+/*
  * Handle MODE_SENSE(6) and MODE_SENSE(10).
  *
  * For TYPE_DISK only.
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -273,6 +273,9 @@ static int generic_handle_cmd(struct tcmu_device *dev,
 							     block_size,
 							     cdb, iovec,
 							     iov_cnt, sense);
+	case RESERVE:
+	case RELEASE:
+		return tcmu_emulate_reserve_release(sense);
 	case MODE_SENSE:
 	case MODE_SENSE_10:
 		return tcmu_emulate_mode_sense(cdb, iovec, iov_cnt, sense);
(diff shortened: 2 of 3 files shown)


bgly:
We are trying to get kernel to handle all reserve type commands. https://patchwork.kernel.org/patch/9654337/

superlibj:
Yes, these days I had missed that  thread .

Thanks.

lxbsz:
@bgly @mikechristie 

What about the plan for the PGR commands ?

Thanks.

mikechristie:
We are waiting for upstream to figure out if they want bgly's kernel patch. At Vault, Christoph was saying he did not like it, but on the lists he has not objected, so how knows. We wait and see if Nick pushes it to Linus.

If it's ok, then for single node tcmu-runner does not have to do anything. It is all handled by the kernel. For HA, we need to come up with a interface to pass the PGR info to tcmu-runner so it can distribute it.

For ESX though, you should not need it. If you turn on VAAI ATS then ESX should use ATS/COMPARE_AND_WRITE instead of those old school reserve/release right?

bgly:
Nick did not object to the patch, so ill address his comments and maybe with your sign off he'll push it through.

lxbsz:
Sorry for late, yes I will use and test the VAAI ATS/CAW case later.

Yes I had aleady test the PGR patch and it works fine for me.

lxbsz:
~ #  esxcli system settings advanced list -o /DataMover/HardwareAcceleratedMove
   Path: /DataMover/HardwareAcceleratedMove
   Type: integer
   Int Value: 1
   ... 
   Description: Enable hardware accelerated VMFS data movement (requires compliant hardware)

~ # esxcli system settings advanced list -o /DataMover/HardwareAcceleratedInit
   Path: /DataMover/HardwareAcceleratedInit
   Type: integer
   Int Value: 1
   ...
   Description: Enable hardware accelerated VMFS data initialization (requires compliant hardware)

~ # esxcli system settings advanced list -o /VMFS3/HardwareAcceleratedLocking
   Path: /VMFS3/HardwareAcceleratedLocking
   Type: integer
   Int Value: 1
   ...
   Description: Enable hardware accelerated VMFS locking (requires compliant hardware)

On my current environment with exsi 5.5， with the ATS enabled, but still falls back to RESERVE/RELEASE.
No COMPARE_AND_WRITE received.
  
The backend is based file_example or rbd handler.

@mikechristie 
Any ideas about this ?

Thanks.

bgly:
@lxbsz It looks like Initial PGR support was accepted, so thats good news.

lxbsz:
@bgly 
Yes，That's good news. Days ago I have replied after your V3, but maybe there something wrong with my mail, I'm not sure you have ever gotten my reply mails.

lxbsz:
@bglk @mikechristie

On my esxi 5.5 test case ,even turn on VAAI ATS this patch still will be needed.

This will be used just before the ATS could be used.

Without this the ATS couldn't be support.

PR #109, opened 2017-04-17
Aio fixups

mikechristie:
Fix ups to Venky's aio patches and patches from PR

https://github.com/open-iscsi/tcmu-runner/pull/100

ported to the aio code.


tcmur_cmd_handler.c
```diff
@@ -628,26 +584,58 @@ static int handle_passthrough(struct tcmu_device *dev,
...
+	uint8_t *cdb = cmd->cdb;
+	struct iovec *iovec = cmd->iovec;
+	size_t iov_cnt = cmd->iov_cnt;
+	uint32_t block_size = tcmu_get_dev_block_size(dev);
+	off_t offset = block_size * tcmu_get_lba(cdb);
 
-	if (rhandler->handle_cmd) {
-		ret = handle_passthrough(dev, rhandler, cmd);
-		if ((ret == TCMU_ASYNC_HANDLED) || (ret != TCMU_NOT_HANDLED))
-			return ret;
```
liuy: wondering why you dropped 'command passthough' mechanism out. If only ceph can offload command and others can't, we should only passthrough the command to rbd driver and allow others enjoy generic handling. Besides, you are making tcmu-runner more and more ceph centeric, losing modularity. for e.g., transition_state and report_state should be in the upper geneic layer.  Store drier should only concentrate on the store details, not on the iscsi protocol details.

mikechristie: I am not sure what you mean. handle_passthrough is called from tcmur_cmd_handler.c. It is called for all handlers that use the tcmu-runner daemon. file_optical uses passthrough for all its commands for example. Is there a bug where it is not getting called?

I am also not sure what you mean by making it ceph specific. translate/report are called from the upper generic layer SCSI command processing aren't they? The upper level generic SCSI code in tcmu-runner processing the SCSI parts in alua.c.  Any handler can hook into the callbacks to implement any low level specifics like how they can for read/write/flush. For example, gluster is working on something similar to fence out other nodes during failover/failback. You could implement a drbd handler and for its transition callout you would do something like the drbdadm resrouce commands.  Like rbd, for these devices, we have to call into the handler for failover/failback to do low level operations like make sure commands from other nodes are not still running when switching nodes.

If you do not implement a transition/report callout and still wanted ALUA or your handler can do active/active and/or do not need any low level helper, you can setup ports group with targetcli, and tcmu-runner will report that ALUA info from the generic SCSI handler and it call just works for the handler.

mikechristie: Ah, I think I know what you mean for some of this.

For handlers that do not do all passthrough like file_optical, handle_passthrough is now called at the bottom of the command handling switch instead of before it. So instead of allowing handlers to override all commands, it now only allows handlers to implement commands that are not handled by the generic tcmu-runner SCSI code. Is that what you mean by your first comment about dropping the command passthrough out? If so, that was a mistake. We can add the original behavior back.

From there I am guessing you are saying I should do RTPG/STPG handling in rbd.. It sounds like you think I am making it ceph specific because I added callouts which only rbd.c implemented? For this, I disagree. Besides other handlers working on support and ALUA being generic for all other handlers, we can either have handlers:

1. Call the alua.c code from their handle_cmd callout.
2. Have tcmu-runner call the alua code and have runner call into the handlers.

I do not see an advantage to #1, but am open to a discussion.

I preferred #2 because we only have to have one caller of tcmu_emulate_set_tgt_port_grps() that we need to worry about. If we have tcmu-runner calling it and individual handlers, it makes it more of a pain to update later like the write verify code. I did make tcmu_emulate_set_tgt_port_grps a libtcmu function to allow outside daemons to use it since it seemed like it could be useful, but we do not have to worry about them.

It is like how you only converted a couple handlers to the read/write/flush callouts and then we had a mess of different behaviors. That makes maintaining multiple handlers difficult for me.

I also prefer tcmu-runner to do the ALUA calls because it needs to be coordinated with other commands like inquiry. I do not want handlers to be making alua.c calls then having to figure out how to handle port id changes. tcmu-runner should coordinate that. The handler just needs to handle the low level details.

liuy: No, you didn't get my point. I am going to explain my idea by code, gonna write some code and back with a pull request :)

liuy: For original partial command passthrough mechanism, I'll add it back by once we have compext ready from librbd to implement  caw and writesame for rbd and let other handlers remain the same. Until then, I guess you'll find it more useful.

mikechristie: Ok. Fair enough I will wait for the code.

What about UNMAP btw? Did you already hook rbd_aio_discard into it?

liuy: @mikechristie Done the pr. For VAAI support(xcopy, caw, unmap, etc), I(or someone else from my team) will come up with a complete serie once compex is merged by the ceph.

PR #113, opened 2017-04-18
file_example: Fix endless reading the file.

superlibj:
When the current file size is less than its real capacity for
some reason, such just after created by file_check_config,
the preadv will return 0, and the file_read will in a infinite
loop.

Then try to zero the iovecs left.

### DIFF
--- a/api.c
+++ b/api.c
@@ -320,6 +320,18 @@ int tcmu_set_sense_data(uint8_t *sense_buf, uint8_t key, uint16_t asc_ascq,
 }
 
 /*
+ * Zero iovec.
+ */
+void tcmu_zero_iovec(struct iovec *iovec, size_t iov_cnt)
+{
+	while (iov_cnt) {
+		bzero(iovec->iov_base, iovec->iov_len);
+
+		iovec++;
+		iov_cnt--;
+	}
+}
+/*
  * Copy data into an iovec, and consume the space in the iovec.
  *
  * Will truncate instead of overrunning the iovec.
--- a/file_example.c
+++ b/file_example.c
@@ -148,6 +148,13 @@ static int file_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 						  ASC_READ_ERROR, NULL);
 			goto done;
 		}
+
+		if (ret == 0) {
+			/* EOF, then zeros the iovecs left */
+			tcmu_zero_iovec(iov, iov_cnt);
+			break;
+		}
+
 		tcmu_seek_in_iovec(iov, ret);
 		offset += ret;
 		remaining -= ret;
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -78,6 +78,7 @@ uint64_t tcmu_get_lba(uint8_t *cdb);
 uint32_t tcmu_get_xfer_length(uint8_t *cdb);
 off_t tcmu_compare_with_iovec(void *mem, struct iovec *iovec, size_t size);
 void tcmu_seek_in_iovec(struct iovec *iovec, size_t count);
+void tcmu_zero_iovec(struct iovec *iovec, size_t iov_cnt);
 size_t tcmu_memcpy_into_iovec(struct iovec *iovec, size_t iov_cnt, void *src, size_t len);
 size_t tcmu_memcpy_from_iovec(void *dest, size_t len, struct iovec *iovec, size_t iov_cnt);
 size_t tcmu_iovec_length(struct iovec *iovec, size_t iov_cnt);


mikechristie:
If you do a write under the same conditions does it also return a 0?

superlibj:
For the write case, I have not found this.

For one normal case, after the initiater logged in successfully, then if you use "fdisk -l" or do the read operations, the fdisk or inotiator will wait forever till timeout.

superlibj:
I have update the patch by add some code to make sure the iovecs left are zeroed.
---
For the same conditions, if you do the write, for the current runner code, the file as backend will grow the file capacity size and will always return the length intend to write.

mikechristie:
The patch looks correct c/code wise.

SCSI wise though, I think for the command we are then supposed to return a ILLEGAL_REQUEST - LOGICAL BLOCK OUT OF RANGE right. In SBC-4/3 section 5.11 we have

If the specified LBA and the specified transfer length exceed the capacity of the medium (see 4.5), then the device server shall terminate the command with CHECK CONDITION status with the sense key set to ILLEGAL REQUEST and the additional sense code set to LOGICAL BLOCK ADDRESS OUT OF RANGE.

For the WRITE case, what happens if you try to write past the new file sizes's limit? Do you get an error?

lxbsz:
Yes， when the read/write size is exceed, an out of range error will be returned.

Before I did a test by just skipping the lba limitation check to verify the endless pwrite case.

mikechristie:
I think we are talking about different things.

In your test, you are shrinking the file after the capacity has been reported right? What is the use case for this type of action?

- Under normal user circumstances like a controlled shrink using some new targetcli command, then I think we need to readjust tcmu-runners/lio's internal size for the device, send a UA with asc/ascq CAPACITY DATA HAS CHANGED, and for commands that raced and made it to preadv we should fail with an ILLEGAL REQUEST. I am not 100% sure on the ILLEGAL REQUEST though, so if you have a different opinion let me know.

Note: we are working on the management parts (targetcli/tcmu/UA) here

https://github.com/open-iscsi/tcmu-runner/pull/102

- For the case, where we are trying to break things like shrinking the file and not telling tcmu just to see what happens, then I do not think it matters. It is the users fault. We can do the safe and/or easy route.

- If you are working on thin provisioning, then I think we need some more code. I was asking about what happens to writes because for a FS file, the write would just increase the file size normally. If the file being used as a special file like a /dev/block_device, then you get errors. I think we need something to report that behavior differently. With your patch, the FS file case is like LBPRZ=1. For the block device file case we would do LBPRZ=1 for READs, but for WRITEs I think it is broken.

lxbsz:
Maybe some misunderstanding on my part.

Yes. If shrink the backstore file manually after the capacity has been reported could also produce this.

But without any shrink action, like the following detail test case:
1, use targetcli to create a new LUN under user:file backstore, and the backstore file is not exist....
2, targetcli saveconfig
3, reboot the machine
4, targetcli restoreconfig
5, fire the tcmu-runner and the file handler will create the backstore file in file_check_config() and the size will be zero. (Or  must we do something just after this and make the new created file's size to the capacity size reported before step 6 ?)
6, then use the client try to do a read.

This is what I wanted to fix here.

================

I agree about readjusting tcmu-runner/lio's internal size for the device after shrinking.
And I think the  ILLEGAL REQUEST is okay.

lxbsz:
For the file_example backstore, since it allow the none exist backstore file, and it will check and create it automatically. So when the client write to it, the file size will grow but will be limited to the capacity size as reported. But when reading the file and at the same time the current file size is less than capacity size, if the file offset or lba is larger than the current file size it should return zero(as file hole do?).

lxbsz:
Since there is check_lba_and_length() before the file_read/file_write, so the (offset or lba) + data_length won't exceed the capacity size in file_read/file_write. 

What I did a test was just skipping the check_lba_and_length() then can grow larger than capacity size, just for test. Just forget this.

So the endless only exist in file_read() when EOF and the current file size is less than capacity size. And for the file_write() is will always return the size writing without consider the Thin Provisioning.

What as I know for the Thin Provisioning case, when the device is full, when the file grows with writing the pwrite will return an error ? not zero ? am I right ?

mikechristie:
It sounds like your case is thin provisioning (just gonna write thin-p below) right? See section 4.7 in SBC 3 for more info.

I was saying at the bottom of my last comment we report LBPME=0, so I think if we return a capacity of 10 gigs we must have physical storage for it. We need to fail at setup time if we do not. The ret==0 case is not supported and should be failed.

The behavior you are implementing looks more like thin-p support LBPME=1 LBPRZ=1. For that, besides your patch, I think we need to modify the READ CAPACITY response data to report our thin-p values and implement the other bits in section 4.7 of SBC-3.

If you are saying it is not related to provisioning, then I am open to your patch, but just need some more info, because I cannot find it in the spec. It is just not clear to me (SCSI spec wise) what the proper error is if we run out of space to do WRITEs. Is it ok to return a MEDIUM ERROR for WRITEs but still return zeros for READs.

mikechristie:
I think I am going to flip flop, and say it is not related to thin provisioning.

There is no easy way to tell what the backend storage is always doing. It could be a dm device that does all kinds of craziness. We can just the let the user do what they want and deal with the consequences in this case.

When we actually implement the thin-p stuff they can config it.

mikechristie:
Thanks.

PR #116, opened 2017-04-19
rbd/file_example: Update debug info

lxbsz:
For file_example add debug info support.
For rbd update the debug info and make the config and
size to the same line.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -65,8 +65,6 @@ static int tcmu_rbd_open(struct tcmu_device *dev)
 	tcmu_set_dev_private(dev, state);
 
 	config = strchr(tcmu_get_dev_cfgstring(dev), '/');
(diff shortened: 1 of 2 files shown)


rbd.c
```diff
@@ -144,6 +141,9 @@ static int tcmu_rbd_open(struct tcmu_device *dev)
 		goto rbd_close;
 	}
 
+	tcmu_dbg("config %s, size %lld\n", tcmu_get_dev_cfgstring(dev),
+			rbd_size);
+
```
mikechristie: I fixed the tabbing/spacing here, so the rbd_size is aligned with the opening "(", so it consistent with "most" (I know in a lot of places it is a mismatch) of the other coding style we are using.

mikechristie:
Merged. Thanks.

PR #117, opened 2017-04-19
Misc for CDB formats length

lxbsz:
libtcmu: Fix a bug with getting variable type CDB length

api: Use the group code the do the CDB formats check
At the same time this could simplify the code and be more readable.

libtcmu: Convert to use tcmu_get_cdb_length helper


mikechristie:
Nice. Thanks.

PR #120, opened 2017-04-20
libtcmu: Fix a bug with read capacity condition check

lxbsz:
For the generic cmd handler should keep the same with the
SCSI proto.

SBC-2 says:
  If the PMI bit is set to zero and the LOGICAL BLOCK
  ADDRESS field is not set to zero, the device server shall
  terminate the command with CHECK CONDITION status with
  the sense key set to ILLEGAL REQUEST and the additional
  sense code set to INVALID FIELD IN CDB.

In SBC-3, these fields are obsolete, but some SCSI
compliance tests actually check this, so we might as well
follow SBC-2.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -211,8 +211,8 @@ static int generic_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		else
 			return TCMU_NOT_HANDLED;
 	case READ_CAPACITY:
-		if ((cdb[1] & 0x01) || (cdb[8] & 0x01))
-			/* Reserved bits for MM logical units */
+		if ((cdb[1] & 0x01) || (!(cdb[8] & 0x01) &&
+					!!(cdb[2] | cdb[3] | cdb[4] | cdb[5]))
 			return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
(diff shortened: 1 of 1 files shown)


mikechristie:
Hey, What compliance tests are you using?

mikechristie:
I think the problem with this patch is:

1. We are reporting SPC 3, so I think we have to follow SBC 2. If that is correct, does the compliance test other parts? Does it set PMI=1 and send a valid LBA? We should be failing that test right?

2. I do not really want to support every possible SBC-2/SPC-3 test someone could make if there is no reason to. Are there initiators that will not handle SPC-4 ok?

It looks like SCST reports SPC-4, but lio (kernel target) and tgt report SPC-3.

lxbsz:
I was using the old version, just close it.

mikechristie:
Hey @lxbsz,

What do you think about setting our SPC version to 4 still?

I ended up quickly testing linux, windows and ESX and they seem to run ok. SCST is a widely used target, so it seems safe. Any thoughts? Did setting the SPC version to 4 help use pass the compliance test?

lxbsz:
Yes, 

This is one private suit application and for the newest version with SPC 4 works.
And agreed to setting SPC version 4 for runner.

Thanks,

PR #122, opened 2017-04-20
tcmur:  commands not handled will report as not supported

lxbsz:
The info will be helpful to do the debug.

### DIFF
--- a/main.c
+++ b/main.c
@@ -513,6 +513,9 @@ static void *tcmur_cmdproc_thread(void *arg)
 			else
 				ret = tcmur_generic_handle_cmd(dev, cmd);
 
+			if (ret == TCMU_NOT_HANDLED)
+				tcmu_warn("Command 0x%x not supported\n", cmd->cdb[0]);
+
 			/*
 			 * command (processing) completion is called in the following
 			 * scenarios:
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -626,7 +626,6 @@ static int handle_generic_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	case MODE_SELECT_10:
 		return tcmu_emulate_mode_select(cdb, iovec, iov_cnt, sense);
 	default:
-		tcmu_err("unknown command %x\n", cdb[0]);
 		return TCMU_NOT_HANDLED;
 	}
 }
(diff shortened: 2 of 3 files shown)


tcmur_cmd_handler.c
```diff
@@ -553,6 +553,8 @@ int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		/* Try to passthrough the default cmds */
 		if (rhandler->handle_cmd)
 			ret = handle_passthrough(dev, cmd);
+		else
+			tcmu_warn("unknown command %x\n", cdb[0]);
```
mikechristie: Since you added another null handle_cmd check to tcmur_cmd_passthrough_handler, I think you can just kill the above check and add the warning message to tcmur_cmd_passthrough_handler's check.

lxbsz: For the tcmur_cmd_passthrough_handler, this will only for the passthrough type. Yes, there should add the same warning message too.

But for the none passthrough type the warning message is need too.

So I will update this patch with adding warning message in the passthough handler.

lxbsz:
Maybe report this as "not handled : " is better?

For example as the RESERVE and RELEASE commands, using exsi without VAAI CTS support, with the warning info it will be more easy.

lxbsz:
s/CTS/ATS

mikechristie:
Yeah, sounds good.

lxbsz:
Has been updated, please review.

mikechristie:
How about:

Command 0x%x not supported.

If ok, I can edit that in when I merge. I think some users might wonder if "not handled" means the command was not handled because of an error or resources issue instead of us just not yet implementing support for it.

main.c
```diff
@@ -513,6 +513,9 @@ static void *tcmur_cmdproc_thread(void *arg)
 			else
 				ret = tcmur_generic_handle_cmd(dev, cmd);
 
+			if (ret == TCMU_NOT_HANDLED)
+				tcmu_warn("not handled command %x\n", cmd->cdb[0]);
+
```
mikechristie: If we add it here, I think we want to remove it from all other places. For example, handle_generic_cmd and fbo_handle_cmd.

lxbsz:
Yes please do it, the "Command 0x%x not supported."  looks much better.

main.c
lxbsz: Yes, it is.

mikechristie:
Thanks.

PR #125, opened 2017-04-24
runner: add possible repetitive handler registerring check.

lxbsz:
Just encountered this when testing the old and new versions at
the same by renaming the old handlerXX.so to handlerYY.so, and
forgot to make clean or remove them. But this won't report any
warning or error message.

Maybe my way doing the test is not very wise, but this could be
a potential bug.


lxbsz:
updated with s/beed/been

mikechristie:
Seems like a good idea to me. Thanks.

PR #126, opened 2017-04-25
ceph rbd ALUA failover/failback support

mikechristie:
Very limited ceph rbd failover/failback support using linux and windows initiators. You can export N images as SCSI LUNs through M targets. The major limitation is that you can only have one enabled port per LUN per node. In targetcli, if you were to export an image, testtcmulock, through one HA iscsi target that spans 2 nodes (so running LIO on 2 nodes and each node is exporting a iscsi target ion.1999-09.com.tcmu:alua) and each node has a iscsi portal, it would look like this on one node. On the other node, you would have the opposite tpgs enabled/disabled:

targetcli ls
o- / .................................................................... [...]
  o- backstores ......................................................... [...]
  | o- block ............................................. [Storage Objects: 0]
  | o- fileio ............................................ [Storage Objects: 0]
  | o- pscsi ............................................. [Storage Objects: 0]
  | o- ramdisk ........................................... [Storage Objects: 0]
  | o- user:qcow ......................................... [Storage Objects: 0]
  | o- user:rbd .......................................... [Storage Objects: 1]
  |   o- rbd1 .......................... [rbd/testtcmulock (10.0GiB) activated]
  |     o- alua .............................................. [ALUA Groups: 3]
  |       o- default_tg_pt_gp ........................... [ALUA state: Standby]
  |       o- tcmu2 ...................................... [ALUA state: Standby]
  |       o- tcmu3 ...................................... [ALUA state: Standby]
  o- iscsi ....................................................... [Targets: 1]
  | o- iqn.1999-09.com.tcmu:alua .................................... [TPGs: 2]
  |   o- tpg1 ...................................................... [disabled]
  |   | o- acls ..................................................... [ACLs: 0]
  |   | o- luns ..................................................... [LUNs: 1]
  |   | | o- lun0 ......................................... [user/rbd1 (tcmu2)]
  |   | o- portals ............................................... [Portals: 1]
  |   |   o- 192.168.32.101:3260 ......................................... [OK]
  |   o- tpg2 ............................................. [gen-acls, no-auth]
  |     o- acls ..................................................... [ACLs: 0]
  |     o- luns ..................................................... [LUNs: 1]
  |     | o- lun0 ......................................... [user/rbd1 (tcmu3)]
  |     o- portals ............................................... [Portals: 1]
  |       o- 20.15.0.18:3260 ............................................. [OK]

The next patches will add ESX support, but require changes to gwcli and ceph-iscsi-config and are a little more complicated because we have to be able to return all the port groups from all nodes.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -547,7 +547,7 @@ static int tcmu_rbd_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 
 	aio_cb = calloc(1, sizeof(*aio_cb));
 	if (!aio_cb) {
-		tcmu_err("could not allocated aio_cb\n");
+		tcmu_dev_err(dev, "Could not allocate aio_cb.\n");
 		goto out;
 	}
 
@@ -557,7 +557,7 @@ static int tcmu_rbd_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 
 	aio_cb->bounce_buffer = malloc(length);
 	if (!aio_cb->bounce_buffer) {
-		tcmu_err("could not allocate bounce buffer\n");
+		tcmu_dev_err(dev, "Could not allocate bounce buffer.\n");
 		goto out_free_aio_cb;
 	}
 
@@ -628,7 +628,7 @@ static int tcmu_rbd_write(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 
 	aio_cb = calloc(1, sizeof(*aio_cb));
 	if (!aio_cb) {
-		tcmu_err("could not allocated aio_cb\n");
+		tcmu_dev_err(dev, "Could not allocate aio_cb.\n");
 		goto out;
 	}
 
@@ -638,7 +638,7 @@ static int tcmu_rbd_write(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 
 	aio_cb->bounce_buffer = malloc(length);
 	if (!aio_cb->bounce_buffer) {
-		tcmu_err("failed to allocate bounce buffer\n");
+		tcmu_dev_err(dev, "Failed to allocate bounce buffer.\n");
 		goto out_free_aio_cb;
 	}
 
@@ -677,7 +677,7 @@ static int tcmu_rbd_flush(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 
 	aio_cb = calloc(1, sizeof(*aio_cb));
 	if (!aio_cb) {
-		tcmu_err("could not allocated aio_cb\n");
+		tcmu_dev_err(dev, "Could not allocate aio_cb.\n");
 		goto out;
 	}
 
--- a/rbd.c
+++ b/rbd.c
@@ -27,18 +27,33 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <endian.h>
-#include <scsi/scsi.h>
 #include <errno.h>
+#include <pthread.h>
+
+#include <scsi/scsi.h>
 
 #include "tcmu-runner.h"
 #include "libtcmu.h"
 
 #include <rbd/librbd.h>
 
+enum {
+	TCMU_RBD_OPENING,
+	TCMU_RBD_OPENED,
+	TCMU_RBD_CLOSING,
+	TCMU_RBD_CLOSED,
+};
+
 struct tcmu_rbd_state {
 	rados_t cluster;
 	rados_ioctx_t io_ctx;
 	rbd_image_t image;
+
+	char *image_name;
+	char *pool_name;
+
+	pthread_spinlock_t lock;	/* protect state */
+	int state;
 };
 
 struct rbd_aio_cb {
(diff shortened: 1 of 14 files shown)


rbd.c
```diff
@@ -49,9 +64,254 @@ struct rbd_aio_cb {
...
+			break;
+		} else if (ret == -ESHUTDOWN) {
+			ret = tcmu_rbd_image_reopen(dev);
+			continue;
+		} else if (ret < 0) {
+			sleep(1);
+			continue;
+		}
+
+		ret = rbd_lock_acquire(state->image, RBD_LOCK_MODE_EXCLUSIVE);
```
dillaman: My original thinking is that you should execute ```rbd_lock_get_owners``` first -- if you receive -ENOENT, that's no problem since it just means that no one currently owns the lock. Since it's an exclusive lock, you can assume only a single owner will be returned. Next, assuming this method is invoked on an ALUA path state change, you would invoke ```rbd_lock_break``` providing the lock owner retrieved from ```rbd_lock_get_owners```. Finally, acquire the lock as you are currently doing. Since you broke the lock, if you have a failure acquiring the lock -- it would be a real issue. If possible, to prevent the possibility of a second ALUA failover, it would be safest if you can somehow validate the returned lock owner with the ALUA path before breaking the lock (unless you know you cannot have another failover between looking up the owner and breaking its lock).

mikechristie:
Jason, if you see my repush above, ignore it. I repushed my rbd changes that did rbd break lock that I was using the other day. I had not seen your comments yet. Will fix up my patches to do what you suggested.

rbd.c
mikechristie: For the last part, if I understand you it is handled. Is the concern we could end up with node1 and node2 getting a failover command, and they both end up returning success? If so this is handled. tcmu-runner will serialize failover commands sent to the same device on the same node.

If we had:
1. initiator sends failover command to node1 then node2, but node1 grabbed lock then node2.

The initiator side, would have marked the path to node1 as bad before sending a failover command to node2. The initiator side will only see the failover command response from node2.  The response from node1 will be dropped. When the initiator gets the success response from node2, it will send IO. When node1's path comes back the initiator side will resync its state and will see node1 does not have the lock, so it will not send IO to it.

2.  initiator sends failover command to node1 then node2, but node2 grabs the lock then node1.

In this case, the initiator side will mark node1 as bad like in case 1. It would then get a success response from node2 for its failover command. But, new IO commands sent from the initiator will fail because node2 had been blacklisted when node1 took the lock. The target side fails those commands with a special error code, and to the initiator it will similar to an implicit failover. The initiator would respond by sending another failover command to node2 and that will succeed and we can do IO again.

The current issue with this case is what I mentioned in my status the other day. The special error code gets dropped by the Lio kernel code and replaced with a error code some initiators treat as a hard error. Instead of sending another failover command, they offline the device.

dillaman: @mikechristie My concern was if you had a failover from A (the current lock owner) to B, but B stalls retrieving the current lock owner, so the path fails over to C successfully which breaks the lock owned by A. Eventually -- B regains network connectivity to the cluster, sees C as the lock owner, issues the break lock request on C, and effectively incorrectly steals the lock.

rbd.c
```diff
@@ -49,9 +64,294 @@ struct rbd_aio_cb {
...
+	pthread_spin_lock(&state->lock);
+	if (state->state != TCMU_RBD_OPENED) {
+		tcmu_dev_dbg(dev, "skipping close. state %d\n", state->state);
+		pthread_spin_unlock(&state->lock);
+		return;
+	}
+	state->state = TCMU_RBD_CLOSING;
+	pthread_spin_unlock(&state->lock);
+
+	ret = is_exclusive_lock_owner(dev);
```
dillaman: Nit: technically -- the lock will be auto-released when the image is closed.

rbd.c
mikechristie: That will be handled like my #2 case above. The initiator will figure things out and do the right thing. We would just have a temp interruption while it figures out who actually has the lock (is active-optimized and who really is standby).

I added the code in the attached patch to make that window smaller. We can still hit the issue if before we even get a failover command the initiator has already sent a failover to another node, and we can hit it between that first get owners / break lock iteration, but I will be able to catch it if it happens on one of our internal retries. I think to do more closely what you want, we will need ESX failover. For that each node will know all the nodes port states, so we will be able to detect if between a report command and when we started to process the failover command someone came in and took the lock.

[rbd-check-owner.txt](https://github.com/open-iscsi/tcmu-runner/files/957021/rbd-check-owner.txt)

mikechristie: One review comment for myself on that patch. I forgot to free the orig_owner pointer.

dillaman: Agreed -- my concern was your scenario 2. I don't think it's worth the brain power at the moment since it will eventually recover in such a scenario. Long term solution would be the ESX solution which requires the kernel change, correct?

mikechristie: For ESX failover, we should not require a kernel change. We just needed to distribute the port state to all the nodes. For PGR support we will need a kernel change. We need PGRs for part of the ESX storage certification.

For ESX failover, I thought we said we could just to the easy route and just store the info in ceph. Something like at setup time, just initialize some "space" in the cluster with all the nodes's port info. When we get the report command, read the data and return it. When we get a failover command, update the data to reflect which port has the lock.

I was not sure where to put the data in the cluster though. I started to just make it an xattr, but also looked into extending the rbd header data.

dillaman: I hoping that extra iSCSI metadata could be stored using the generic RBD ```rbd_metadata_get``` and ```rbd_metadata_set``` calls. We just need some version-controlled format for what we stick in it -- i.e. base64 encode the raw SCSI structures or use JSON.

mikechristie:
For target side setup, you can use targetcli on each node.

I am using Paul Cuzner's ansible tools:
    
https://github.com/pcuzner/ceph-iscsi-ansible/tree/tcmu-fixes
[my config branch has some patches that paul as not yet merged]
https://github.com/mikechristie/ceph-iscsi-config/tree/alua
    
Paul also has a command line tool based off of targetcli:
https://github.com/pcuzner/ceph-iscsi-cli

For dependencies you need:

- a rtslib that supports alua and tcmu. For that I am using rtslib-fb:
https://github.com/open-iscsi/rtslib-fb/
alua/tcmu support was recently added so many distro versions will not have it.

- You also need a updated kernel that has tcmu alua support. That was added in a 4.11-rc, so the current 4.11-rc8 or newer should be ok to use.

- if you are trying to use targetcli manually then you need a version that supports alua and tcmu. The current targetcli-fb tree has support:

https://github.com/open-iscsi/targetcli-fb

PR #131, opened 2017-04-28
refactor out alua specific code out of store driver

liuy:
store drivers have no interest in ISCSI details.

### DIFF
--- a/alua.c
+++ b/alua.c
@@ -519,10 +519,28 @@ int tcmu_transition_tgt_port_grp(struct tgt_port_grp *group, uint8_t new_state,
 	return SAM_STAT_GOOD;
 }
 
+static int tcmu_report_state(struct tcmu_device *dev,
+			     struct tgt_port_grp *group)
+{
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	int ret;
+
+	/* TODO: For ESX return remote ports */
+
+	if (!rhandler->has_lock)
+		return group->state;
+
+	ret = rhandler->has_lock(dev);
+	if (ret <= 0) {
+		return ALUA_ACCESS_STATE_STANDBY;
+	} else {
+		return ALUA_ACCESS_STATE_OPTIMIZED;
+	}
+}
+
 int tcmu_emulate_report_tgt_port_grps(struct tcmu_device *dev,
 				      struct list_head *group_list,
-				      struct tcmulib_cmd *cmd,
-				      tcmu_report_state_fn_t *report_fn)
+				      struct tcmulib_cmd *cmd)
 {
 	struct tgt_port_grp *group;
 	struct tgt_port *port;
@@ -563,20 +581,16 @@ int tcmu_emulate_report_tgt_port_grps(struct tcmu_device *dev,
 		if (group->pref)
 			buf[off] = 0x80;
 
-		if (report_fn) {
-			state = report_fn(dev, group);
-			/*
-			 * Some handlers are not able to async update state,
-			 * so check it now and update.
-			 */
-			if (state != group->state) {
-				if (tcmu_transition_tgt_port_grp(group, state,
-							TPGS_ALUA_IMPLICIT,
-							NULL))
-					tcmu_err("Could not perform implicit state change for group %u\n", group->id);
-			}
-		} else {
-			state = group->state;
+		state = tcmu_report_state(dev, group);
+		/*
+		 * Some handlers are not able to async update state,
+		 * so check it now and update.
+		 */
+		if (state != group->state) {
+			if (tcmu_transition_tgt_port_grp(group, state,
+							 TPGS_ALUA_IMPLICIT,
+							 NULL))
+				tcmu_err("Could not perform implicit state change for group %u\n", group->id);
 		}
 
 		buf[off++] |= state;
--- a/alua.c
+++ b/alua.c
@@ -437,9 +437,45 @@ static bool alua_check_sup_state(uint8_t state, uint8_t sup)
 	return false;
 }
 
+static int tcmu_do_transition(struct tcmu_device *dev,
+			      struct tgt_port_grp *group, uint8_t new_state,
+			      uint8_t *sense)
+{
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	int ret;
+
+	switch (new_state) {
+	case ALUA_ACCESS_STATE_OPTIMIZED:
+	case ALUA_ACCESS_STATE_NON_OPTIMIZED:
+		if (rhandler->lock && rhandler->lock(dev))
+			return tcmu_set_sense_data(sense, HARDWARE_ERROR,
+						   ASC_STPG_CMD_FAILED, NULL);
+		/* TODO for ESX set remote ports to standby */
+		return SAM_STAT_GOOD;
+	case ALUA_ACCESS_STATE_STANDBY:
+	case ALUA_ACCESS_STATE_UNAVAILABLE:
+	case ALUA_ACCESS_STATE_OFFLINE:
+		if (rhandler->unlock) {
+			ret = rhandler->unlock(dev);
+			if (ret < 0)
+				/*
+				 * Return success even though we failed. The initiator
+				 * will send a STPG to the port it wants to activate,
+				 * and that node will grab the lock from us if it hasn't
+				 * already.
+				 */
+				tcmu_dev_err(dev, "Could not release lock. (Err %d)\n",
+					     ret);
+		}
+		return SAM_STAT_GOOD;
+	}
+
+	return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+				   ASC_INVALID_FIELD_IN_PARAMETER_LIST, NULL);
+}
+
 int tcmu_transition_tgt_port_grp(struct tgt_port_grp *group, uint8_t new_state,
-				 uint8_t alua_status, uint8_t *sense,
-				 tcmu_transition_state_fn_t *transition_fn)
+				 uint8_t alua_status, uint8_t *sense)
 {
 	struct tcmu_device *dev = group->dev;
 	int ret;
@@ -450,14 +486,14 @@ int tcmu_transition_tgt_port_grp(struct tgt_port_grp *group, uint8_t new_state,
 	if (!alua_check_sup_state(new_state, group->supported_states)) {
 		if (sense)
 			return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
-					ASC_INVALID_FIELD_IN_PARAMETER_LIST,
-					NULL);
+						   ASC_INVALID_FIELD_IN_PARAMETER_LIST,
+						   NULL);
 		else
 			return SAM_STAT_CHECK_CONDITION;
 	}
 
-	if (transition_fn) {
-		ret = transition_fn(dev, group, new_state, sense);
+	if (sense) {
+		ret = tcmu_do_transition(dev, group, new_state, sense);
 		if (ret != SAM_STAT_GOOD)
 			return ret;
 	}
@@ -536,7 +572,7 @@ int tcmu_emulate_report_tgt_port_grps(struct tcmu_device *dev,
 			if (state != group->state) {
 				if (tcmu_transition_tgt_port_grp(group, state,
 							TPGS_ALUA_IMPLICIT,
-							NULL, NULL))
+							NULL))
 					tcmu_err("Could not perform implicit state change for group %u\n", group->id);
 			}
 		} else {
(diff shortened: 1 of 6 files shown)


mikechristie:
Hey, yeah I like it. It looks nice to me.

What is your concern about SCSI details, the handler modules and the alua patches though? Is it that:

1. the callouts are based on SCSI spec's ALUA definition?

2. the callouts have to translate their internal state to/from SCSI ALUA state values?

3. Both or something else?

My only major issue is that I do not think you are making it more generic. I think the core code is more ceph specific with the patch.  gluster and ceph are using locks, but it is implementation specific how and if they want to clean up IO when transitioning states. I was asking about #1 above, because if you want to make it generic then  I think we need to rename the lock callouts to something not specific to how rbd implement it. I would suggest naming it transition, but if #1 was one of your issues then you are not going to like it :)

If you want to make the argument just go with what are have, then I am open to it since it should just be renames.

For report state, I like how handlers do not have to know about SCSI ALUA states, but the choice of AO (active optimized) and Standby is rbd specific. For example, for implicit ALUA you will probably use AO and ANO. If gluster does not do distributed PGRs they would want to use AO and unavailable. I was asking about #2 above because we will need some flexibility, but if you do not like handlers dealing with SCSI ALUA state values then I am not sure what to do. We could define TCMU_DEV_STATE_SOME_STATE, but they are going to end up matching ALUA ones.

My only minor issue is that the error values you are percolating up can be any old value librbd wants to use. To make it generic define some TCMU_WHATEVER_THE_CALLOUT_IS_CALLED_ERR error values and use them or use specific -Exyz values.

Also, to make it a little more generic, you would want to make one of the errors indicate a retryable error and move the retry handling from tcmu_rbd_lock to the core.

And, for minor nits, your editor switched tabs to spaces when you moved code around, and tcmu_rbd_reopen rename does not make sense to me. I named it image_reopen because it only handles those low level parts. It does not handle the higher level stuff done in tcmu_rbd_open.

liuy:
> My only major issue is that I do not think you are making it more generic. I think the core code is more ceph specific with the patch. gluster and ceph are using locks, but it is implementation specific how and if they want to clean up IO when transitioning states. I was asking about #1 above, because if you want to make it generic then I think we need to rename the lock callouts to something not specific to how rbd implemented it. I would suggest naming it transition, but if #1 was one of your issues then you are not going to like it :)

rbd.{lock, unlock, has_lock}, by any standard, is more abstract than ALUA state transtion & report, which is  just some functions used interally by ALUA implementation. {lock, unlock, has_lock} operate on the volume level, just like {read, write, flush} and probably in the future, lock family would be used for other SCSI implementation.

Mix ALUA/SCSI details and store driver should be our last choice. If you are right that gluster might want a different ALUA state transition & report, let's consider it until someone provide us a real implemention and re-evaluate it.  At that moment, the choice might be make ALUA state transition/report more versatile, instead of making every handler has its own different ALUA implementatioin, who knows. Besides, any handler spectial can just implement .handle_cmd to take over full control.

> For report state, I like how handlers do not have to know about SCSI ALUA states, but the choice of AO (active optimized) and Standby is handler specific. For example, for implicit ALUA you will probably use AO and ANO. If gluster does not do distributed PGRs they would want to use AO and unavailable. I was asking about #2 above because we will need some flexibility, but if you do not like handlers dealing with SCSI ALUA state values then I am not sure what to do. We could define TCMU_DEV_STATE_SOME_STATE, but they are going to end up matching ALUA ones.

This goes very detail but my patch set just re-factor and KEEP your original purpose.  So as above mentioned, we'd better leave it as is until someoe shows up a gluster implementation.

> Also, to make it a little more generic, you would want to make one of the errors indicate a retryable error and move the retry handling from tcmu_rbd_lock to the core.

I don't think so, at least for now. This is lock internal implementation, If all the store driver has retry logic, we might generalize it in the future.  

> And, for minor nits, your editor switched tabs to spaces when you moved code around, and tcmu_rbd_reopen rename does not make sense to me. I named it image_reopen because it only handles those low level parts. It does not handle the higher level stuff done in tcmu_rbd_open.

tab problem fixed. I renamed tcmu_rbd_reopen because your former ALUA impmentation called .reopen. Now it is not necessary so let me rename it as tcmu_rbd_image_reopen as you did.

mikechristie:
>> My only major issue is that I do not think you are making it more generic. I think the core code is more ceph specific with the patch. gluster and ceph are using locks, but it is implementation specific how and if they want to clean up IO when transitioning states. I was asking about #1 above, because if you want to make it generic then I think we need to rename the lock callouts to something not specific to how rbd implemented it. I would suggest naming it transition, but if #1 was one of your issues then you are not going to like it :)
> rbd.{lock, unlock, has_lock}, by any standard, is more abstract than ALUA state transtion & report, 


I am confused. Lock means something very specific right? Locking means that you must prevent others from accessing something right?

> which is just some functions used interally by ALUA implementation. {lock, unlock, has_lock} operate 

Transition and report match the ALUA spec commands, so they are not just some functions used internal by the ALUA implementation.  The locking design is handler implementation specific and specific to what it was supporting. The spec's use of the word transition instead of specifically lock is because you are changing the port's state and it does not require that you necessarily prevent other ports from accessing the device. It is just that way for the active/passive case.


> on the volume level, just like {read, write, flush} and probably in the future, lock family would be used for other SCSI implementation.

You mean the lock family would be used for other ALUA configurations? If so, I was saying in the other comment it wouldn't.  For AO/standby setups it does match well. For a ANO/AO setup, you can access the device from both port groups. For the STPG you would setup access through the requested port to be optimized instead of ANO. You could implement this in a callout called lock, but it would not have locking type of behavior would it?


> Mix ALUA/SCSI details and store driver should be our last choice. If you are right that gluster might want a different ALUA state transition & report, let's consider it until someone provide us a real implemention and re-evaluate it. At that moment, the choice might be make ALUA state 

So here you are saying that the implementation you are doing is based on rbd's locking design and it only supports active/passive right?

> transition/report more versatile, instead of making every handler has its own different ALUA implementatioin, who knows. 

Ok this is probably what I have been misunderstanding.  Are you saying we are adding too much SCSI/ALUA specifics in handlers by allowing them to implement whatever ALUA type of multipating/failover support they want (Active/active vs active/passive vs asymmetric access, etc) or are you also not liking that the handlers have to converting the SCSI ALUA state values to/from their internal values or both :)


>> Also, to make it a little more generic, you would want to make one of the errors indicate a retryable error and move the retry handling from tcmu_rbd_lock to the core.
> I don't think so, at least for now. This is lock internal implementation, If all the store driver has retry logic, we might generalize it in the future.

Maybe I should have not had used the word generic here.  If we do not want handlers to have to deal with SCSI issues like STPG timeouts, then the retry accounting/driving should be done in the generic code and not by handlers. The handler would tell the core code if the failure was retryable at the low level like if it was a transient network failure. The core code would determine if we can retry or the command is about to timeout.

liuy:
> You mean the lock family would be used for other ALUA configurations? If so, I was saying in the other comment it wouldn't. For AO/standby setups it does match well. For a asymmetric ANO/AO setup, you can access the device from both port groups at the same time similar to active/active but one port's access is going to be slower. For the STPG you would setup access through the requested port to be optimized instead of ANO. You could implement this in a callout called lock, but it would not have locking type of behavior would it?

I meant 'other than ALUA', for any SCSI logic need to grab the lock and do something exclusively then releasee the lock. I am not sure if there exist such logic or not in the future. ALUA is just very subset of iSCSI protocol. I meant the whole protocol.  I just refactored the code based on your ALUA implementation, thinking that SCSI layer should sit on top of store driver. For e.g, Linux kernel block layer <-- SCSI layer --> Device driver, each layer is clear and not inter-mixed.  

> Are you saying we are adding too much SCSI/ALUA specifics in handlers by allowing them to implement whatever ALUA type of multipating/failover support they want (Active/active vs active/passive vs asymmetric access, etc) 

Yes, this is what I meant. Why store driver should know all about SCSI details? Does ata driver in linux kernel know about IO scheduler? No. I guess you don't like the idea of modularization of the code, that store driver should only concentrate on the operations of volume itself. If store driver (in your words, handler) want to take full control of SCSI details, it should implement .handle_cmd(). Otherwise, it should allow upper SCSI layer to make SCSI strategy for it. The store driver should only provide volume operation mechanims to uppper layer.

>  If we do not want handlers to have to deal with SCSI issues like STPG timeouts, then the retry accounting/driving should be done in the generic code and not by handlers. The handler would tell the core code if the failure was retryable at the low level like if it was a transient network failure. The core code would determine if we can retry or the command is about to timeout.

For transient network failure, I'd suggest connection repair done in the store driver level within a timeout window. In this way, upper layer should only deal with 'timeout', be it too slow to response or the network is down. The store layer should provide an abstraction as simple as possible to upper layer. The connection should be maintained in the store layer. But for request, which SCSI layer maitains, it can still retry the timeouted request or just return the error up. This is out of scope of this topic.

Anyway, If you don't like the idea of modularization, please simply close this pr.

mikechristie:
Hey, Ok. I get it. I will merged the patches and deal with other failover types later. Have just one other comment.

And, just a side note. You know that handler and r_handler (tcmu-runner handler) name I use is just the naming used in the docs and name of the structure with the callouts already right? :) I didn't make it up. If you want to call it a store, just change it. You started to do it with the generic_make_cmd patches, but it was only that instance, so I changed it back to match the docs and other code. I really do not care as long as it is consistent.

alua.c
```diff
@@ -483,10 +519,28 @@ int tcmu_transition_tgt_port_grp(struct tgt_port_grp *group, uint8_t new_state,
...
+	} else {
+		return ALUA_ACCESS_STATE_OPTIMIZED;
+	}
+}
+
 int tcmu_emulate_report_tgt_port_grps(struct tcmu_device *dev,
 				      struct list_head *group_list,
-				      struct tcmulib_cmd *cmd,
-				      tcmu_report_state_fn_t *report_fn)
+				      struct tcmulib_cmd *cmd)
```
mikechristie: There is just one other issue. The code is designed to support tcmu-runner and other daemons and mixed modes. libtcmu was supposed is supposed to be general and not be tied to one daemon or mode.

The alua.c code is currently in libtcmu and was able to support non tcmu-runner based daemons. That is why we passed in the report/transition callouts. So, you need to either pass the lock callouts in or just move the alua.c code from libtcmu to tcmu-runner.

liuy: > The alua.c code is currently in libtcmu and was able to support non tcmu-runner based daemons. That is why we passed in the report/transition callouts. So, you need to either pass the lock callouts in or just move the alua.c code from libtcmu to tcmu-runner.

By design, yes you are right but in reality, I think it is okay to move fast, leaving this to be settled down later. If anyone finds alua mechanism too fixed to change for him, he should make a pr. Besides, at least he can implement .handle_cmd to take over and put all the ALUA logic in its own deamon.   Anyway, I guess there is no serieous libtcmu user that implement its own target daemon. Most users will simply run tcmu :)

liuy:
> And, just a side note. You know that handler and r_handler (tcmu-runner handler) name I use is just the naming used in the docs and name of the structure with the callouts already right? :) I didn't make it up. If you want to call it a store, just change it. You started to do it with the generic_make_cmd patches, but it was only that instance, so I changed it back to match the docs and other code. I really do not care as long as it is consistent.

I do not care it either. I just referred it as 'store' after I moved the SCSI logic out of handlers. Since all the docs and codes still refer it as handler, so l'll simply follow the rule call it handler for now.

alua.c
mikechristie: You saw where I said I am fine with making it runner specific, right? Are you also not going to modify the cmake so it is built as part of tcmu-runner instead of libtcmu?

It is fine to start small and build on it, but if everyone has their own policy about where to add fields and build files we end up with a mix match and someone has to clean it up eventually.

If we are saying that there is no libtcmu user that implements its own daemon and there probably will not be, then lets discuss that in another PR. Trying to maintain that design is a pain, but I was under the impression at some point there were outside users, because they did some of the low level libtcmu work.

liuy: > You saw where I said I am fine with making it runner specific, right?
No, I just personally think that runner is needed more than libtcmu. It would make dev/maintain life easier withtout attention to libtcmu.

>  Are you also not going to modify the cmake so it is built as part of tcmu-runner instead of libtcmu?
Done make the alua runner specific.

mikechristie: > No, I just personally think that runner is needed more than libtcmu. It would make dev/maintain life easier withtout attention to libtcmu.

Agree. Lol. We completely agree on something! :)

I will try to find some time to dig into all the past commits and contact the original users and open a PR to depreciate support.

liuy:
so any more comment about this patch set?

mikechristie:
Thanks.

PR #132, opened 2017-04-28
qcow: Clarify bad magic error message

seroyer:
Previously, if you mapped a raw file using user:qcow, it would display
an error message saying "bad magic" with a stale errno message.  Now
make it clear that this may be expected if it's a raw file.


mikechristie:
Thanks.

PR #133, opened 2017-05-01
logger: libtcmu/qcow convert to use logger

lxbsz:
---
These have been missed before.

### DIFF
--- a/qcow.c
+++ b/qcow.c
@@ -307,15 +307,18 @@ static int qcow2_probe(struct bdev *bdev, int dirfd, const char *pathname)
 	tcmu_dbg("%s\n", __func__);
 
 	if (faccessat(dirfd, pathname, R_OK|W_OK, AT_EACCESS) == -1) {
-		perror("no file access");
+		tcmu_err("faccessat failed dirfd %d, pathname %s, errno %d\n",
+			 dirfd, pathname, errno);
 		return -1;
 	}
 	if ((fd = openat(dirfd, pathname, O_RDONLY)) == -1) {
-		perror("openat failed");
+		tcmu_err("openat failed dirfd %d, pathname %s, errno %d\n",
+			 dirfd, pathname, errno);
 		return -1;
 	}
 	if (pread(fd, &head, sizeof(head), 0) == -1) {
-		perror("read fail");
+		tcmu_err("pread failed dirfd %d, pathname %s, errno %d\n",
+			 dirfd, pathname, errno);
 		goto err;
 	}
 	if (be32toh(head.magic) != QCOW_MAGIC) {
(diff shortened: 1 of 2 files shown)


qcow.c
```diff
@@ -279,16 +279,29 @@ static int qcow_probe(struct bdev *bdev, int dirfd, const char *pathname)
...
 		return -1;
-	if (pread(fd, &head, sizeof(head), 0) == -1)
+	}
+	if (pread(fd, &head, sizeof(head), 0) == -1) {
+		tcmu_err("read failed fd: %d\n", fd);
 		goto err;
-	if (be32toh(head.magic) != QCOW_MAGIC)
+	}
+	if (be32toh(head.magic) != QCOW_MAGIC) {
+		tcmu_warn("not qcow: will treat as raw: %s", pathname);
```
seroyer: Next it will try qcow2, so you will get two of these messages for each file if it is raw.

qcow.c
```diff
@@ -279,16 +279,29 @@ static int qcow_probe(struct bdev *bdev, int dirfd, const char *pathname)
...
 		goto err;
-	if (be32toh(head.version) != 1)
+	}
+	if (be32toh(head.version) != 1) {
+		tcmu_err("version = %d\n", head.version);
```
seroyer: If the file is qcow2, this will show up as an error in the log, which is misleading.

qcow.c
lxbsz: Yes, for other errors if occurs will almost be twice too.

I will remove all logger messages in qcow later.

For the current route path the qcow/qcow2 probe will be duplicated, maybe could check the image type before probe and then just jump to it. Then we could know explicitly when and where the error triggered.

seroyer: Yes, that sounds like a good plan.

mikechristie:
@seroyer, it looks ok to me. Do you have any more comments?

qcow.c
```diff
@@ -307,15 +307,17 @@ static int qcow2_probe(struct bdev *bdev, int dirfd, const char *pathname)
...
 	}
 	if ((fd = openat(dirfd, pathname, O_RDONLY)) == -1) {
-		perror("openat failed");
+		tcmu_err("openat failed dirfd %d, pathname %s, errno %d\n",
+			 dirfd, pathname, errno);
 		return -1;
 	}
 	if (pread(fd, &head, sizeof(head), 0) == -1) {
-		perror("read fail");
+		tcmu_err("read failed fd: %d\n", fd);
```
seroyer: It would be more helpful if you printed the pathname and errno.  I'm not sure that the fd is overly useful here.

lxbsz: Has updated.

seroyer:
Looks good to me.

mikechristie:
Merged thanks.

PR #134, opened 2017-05-03
api: mode sense add control page support

lxbsz:

### DIFF
--- a/api.c
+++ b/api.c
@@ -632,13 +632,68 @@ int handle_cache_page(uint8_t *buf, size_t buf_len)
 	return 20;
 }
 
+static int handle_control_page(uint8_t *buf, size_t buf_len)
+{
+	if (buf_len < 12)
+		return -1;
+
+	buf[0] = 0x0a;
+	buf[1] = 0x0a;
+
+	/* From spc4r31, section 7.5.7 Control mode Page
+	 *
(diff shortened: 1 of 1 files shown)


api.c
```diff
@@ -632,13 +632,35 @@ int handle_cache_page(uint8_t *buf, size_t buf_len)
...
+	/* The logical unit won't implicitly save any log parameters */
+	buf[2] = 0x02;
+
+	buf[3] = 0x0;
+	buf[4] = 0x0;
+	buf[5] = 0x40;
+	buf[8] = 0xff;
+	buf[9] = 0xff;
+	buf[11] = 0x20;
+
```
mikechristie: Could you add a comment for the bits you are setting similar to the other code when we are setting cdb bits, because no one wants to dig in the specs :) For example:

/* GLTSD set because we don't implicitly save log parameters */
but[2]  = 0x02;
....

/* TAS = 1. Currently not settable by tcmu. Using the LIO default. */
buf[5] = 0x04;
..
/* unlimited BUSY TIMEOUT PERIOD */
buf[8] =
buf[9] =

...

lxbsz: Sure, will update this.

lxbsz:
Has be updated, please review.

mikechristie:
Hey, was just asking for a one or two line description. Thanks for being so thorough.

PR #136, opened 2017-05-07
api: Enable COMPARE_AND_WRITE command

lxbsz:
Since the runner has already supported the CAW command, enable it.

If the MAXIMUM COMPARE AND WRITE LENGTH field, which in "SCSI Commands
Reference Manual" the section "Block Limits VPD Page (B0h)" set to
zero, means disable the CAW command.


mikechristie:
Thanks.

PR #137, opened 2017-05-07
tcmur: Allow the handlers to handle the cmds through its own method 

lxbsz:
This will allow the handlers could handle part commands through its
own ways firstly, and then try to handle the other through the
generic methods.

This will be helpful for some case, for example the XCOPY command:
The runner could implement one generic method by using handle_read()
and handle_write(), and the device could also implement its own
higher efficiency ways to do this, and the other handlers could fall
back to the generic methods.

Generically, if the generic command handle callout could handle the
specify commands correctly and hasn't any lower efficiency problem,
the handlers shouldn't implement it again.

For the ->handle_cmd callout by handlers, it should obey a rule that
only when the cmds are not supported by it should return
TCMU_NOT_HANDLED. Then the runner will try to handle the cmds in
generic methods.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -774,22 +774,15 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		ret = handle_format_unit(dev, cmd);
 		break;
 	case MAINTENANCE_IN:
-		if ((cdb[1] & 0x1f) == MI_REPORT_TARGET_PGS) {
+		if ((cdb[1] & 0x1f) == MI_REPORT_TARGET_PGS)
 			ret = handle_rtpg(dev, cmd);
-			break;
-		}
-		goto passthrough;
+		break;
 	case MAINTENANCE_OUT:
-		if (cdb[1] == MO_SET_TARGET_PGS) {
+		if (cdb[1] == MO_SET_TARGET_PGS)
 			ret = handle_stpg(dev, cmd);
-			break;
-		}
-		goto passthrough;
+		break;
 	default:
-passthrough:
-		/* Try to passthrough the default cmds */
-		if (rhandler->handle_cmd)
-			ret = handle_passthrough(dev, cmd);
+		ret = TCMU_NOT_HANDLED;
 	}
 
 	if (ret != TCMU_ASYNC_HANDLED)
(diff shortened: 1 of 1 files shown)


tcmur_cmd_handler.c
```diff
@@ -745,6 +745,18 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 
 	track_aio_request_start(rdev);
 
+	/*
+	 * Try to passthrough the cmds to the device.
+	 */
+	if (rhandler->handle_cmd) {
+		ret = handle_passthrough(dev, cmd);
+		if (ret != TCMU_NOT_HANDLED)
```
mikechristie: If it is moved here, we will have to handle the case where handlers implement a handle_cmd and use the runner threading which is going to be a pain.

I think maybe we have to go back to the idea where the handler exposed an array of commands it offloaded and check before calling.

lxbsz: Yes it will be a pain.

Exposing an array of commands is a good idea, but IMO to simplify this, if the handler is not passthrough only type && it has a handle_cmd callout, it should mean that the handler want to handle some commands by itself, and we should try handle_cmd callout first.

And all the handler_cmd callout should obey a rule that only when the cmds are not supported by it should return TCMU_NOT_HANDLED. Then the runner will try to handle the cmds in generic.

Thanks,

mikechristie: I think this sounds good to me.

For the case where a handler cannot do its own aio (nr_threads > 1) then they can just call async_handle_cmd themselves to schedule execution.

Will try to test tonight and merge tomorrow.

lxbsz: Just one confusion about the nr_threads > 1, does that mean a handler couldn't do its own aio ?

lxbsz:
Has updated this again.

For now there are two options:
1, pass through only case, this will let the handler has a full control about all the commands.
2, partially pass through case, handlers could implement its own .handle_cmd to control some commands and the others will fall back to runner's generic routine.

====================

According to my understanding if handlers' .nr_threads > 0, then the runner will do the aio for handle_cmd/read/write/flush all. Or the handlers will do it own aio if supported.

That's right ?

Thanks,

tcmur_cmd_handler.c
liuy: @mikechristie @lxbsz we actually discussed this in #91 and settled down try .handle_cmd firstly to avoid array. Just Mike removed this passthrough mechanism accidentally or not

mikechristie: lxbsz, nr_threads means handler can't do its own aio, so every command is kicked off in a runner thread.

liuy, yes we did. There is a new case with Venky's aio code though and we were discussing if we want to rearrange the code to handle it, do an array, or just leave it to handlers to figure it out. We decided on the latter since it was the most simple.

mikechristie: lxbsz meant nr_threads > 0.

mikechristie:
lxbsz,

I guess it depends on the version of the patch.

- When you were calling handle_cmd directly, then the handler had to do its own aio/threading for those commands.

This had the benefit that it was simple.

I was saying above I was fine with that approach. I was just running some tests with the other handlers to make sure it all worked ok.

- When you call tcmur_cmd_passthrough_handler like in the current patch,  the problem is that for nr_threads > 0 handlers, this will just kick of the execution of handle_cmd to a thread, so  ret here:

+	ret = tcmur_cmd_passthrough_handler(dev, cmd);
 +	if (ret != TCMU_NOT_HANDLED)
 +		return ret;

will be TCMU_ASYNC_HANDLED. The thread will then get run and execute handle_cmd and if it was just a READ/WRITE command then handle_cmd would return TCMU_NOT_HANDLED. In passthrough_work_fn you would need to handle that by then executing the generic code. One issue with that would then be do we want to rearrange the code so this is handled from the existing thread we are executing from or just use the existing functions where it will get rescheduled in a new thread.

It kinda becomes a good deal of work for a case that might not come up.

tcmur_cmd_handler.c
lxbsz: Yes, that what I thought and understood about nr_threads.

lxbsz:
For the 2nd version, that's to say if the handler is a partial pass through type, the handle_cmd will be out of control by .nr_threads and it will be wholly another path and decided by the handler itself to do aio or not.

And the .nr_threads will only affect the write/read/flush.

For the 3nd version, yes it is.

And I will fall back to 2nd version.

mikechristie:
Thanks.

PR #141, opened 2017-05-11
tcmur: Add WRITE_SAME_16 command support

lxbsz:
This will add a emulator for WRITE_SAME_16 command in runner. For
now the unmap option is not supported, and will be added later
when needed.

The Write Same Not Zero bit is set to one, and will be added later.

### DIFF
--- a/api.c
+++ b/api.c
@@ -479,6 +479,7 @@ finish_page83:
 		int max_xfer_length;
 		uint16_t val16;
 		uint32_t val32;
+		uint64_t val64;
 
 		memset(data, 0, sizeof(data));
 
@@ -486,6 +487,12 @@ finish_page83:
 
 		val16 = htobe16(0x3c);
 		memcpy(&data[2], &val16, 2);
+
+		/* WSNZ = 1: the device server won't support a value of zero
+		 * in the NUMBER OF LOGICAL BLOCKS field in the WRITE SAME
+		 * command CDBs
+		 */
+		data[4] = 0x01;
 
 		/*
 		 * From SCSI Commands Reference Manual, section Block Limits
@@ -521,6 +528,10 @@ finish_page83:
 		/* Optimal xfer length */
 		memcpy(&data[12], &val32, 4);
 
+		/* MAXIMUM WRITE SAME LENGTH */
+		val64 = VPD_MAX_WRITE_SAME_LENGTH;
+		memcpy(&data[36], &val64, 8);
+
 		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, sizeof(data));
 
 		return SAM_STAT_GOOD;
@@ -550,6 +561,8 @@ finish_page83:
 	}
 	break;
 	default:
+		tcmu_dev_err(dev, "Vital product data page code 0x%x not support\n",
+			     cdb[2]);
 		return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
 					   ASC_INVALID_FIELD_IN_CDB, NULL);
 	}
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -53,6 +53,8 @@ struct tcmulib_cmd;
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
+#define VPD_MAX_WRITE_SAME_LENGTH 0xFFFF
+
 typedef void (*cmd_done_t)(struct tcmu_device *, struct tcmulib_cmd *, int);
 
 struct tcmulib_cmd {
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -144,6 +144,158 @@ static int write_work_fn(struct tcmu_device *dev,
 				block_size * tcmu_get_lba(cmd->cdb));
 }
 
+struct write_same {
+	pthread_cond_t cond;
+	pthread_mutex_t lock;
+	bool cbk_finished;
+};
+
+static void handle_write_same_sync(struct tcmu_device *dev,
+				   struct tcmulib_cmd *cmd,
+				   int ret)
+{
+	struct write_same *write_same = cmd->cmdstate;
+
+	pthread_mutex_lock(&write_same->lock);
+	write_same->cbk_finished = true;
+	pthread_cond_signal(&write_same->cond);
+	pthread_mutex_unlock(&write_same->lock);
+}
+
+static int write_same_16_work_fn(struct tcmu_device *dev,
+				 struct tcmulib_cmd *cmd)
+{
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	uint32_t block_size = tcmu_get_dev_block_size(dev);
+	uint32_t lba_cnt = tcmu_get_xfer_length(cmd->cdb);
+	uint64_t start_lba = tcmu_get_lba(cmd->cdb);
+	struct write_same *write_same = cmd->cmdstate;
+	void *iov_base;
+	size_t iov_len;
+	int ret;
+
+	tcmu_dev_dbg(dev, "Start lba: %llu, number of lba:: %hu, end lba: %llu\n",
+		     start_lba, lba_cnt, start_lba + lba_cnt);
+
+
+	iov_base = cmd->iovec->iov_base;
+	iov_len = cmd->iovec->iov_len;
+
+	write_same->cbk_finished = false;
+	cmd->done = handle_write_same_sync;
+
+	/*
+	 * Write contents of the logical block data(from the Data-Out Buffer)
+	 * to each LBA in the specified LBA range.
+	 */
+	while (lba_cnt--) {
+		cmd->iovec->iov_base = iov_base;
+		cmd->iovec->iov_len = iov_len;
+		ret = rhandler->write(dev, cmd, cmd->iovec, cmd->iov_cnt,
+				      tcmu_iovec_length(cmd->iovec, cmd->iov_cnt),
+				      block_size * start_lba);
+		if (ret)
+			goto out;
+
+		pthread_mutex_lock(&write_same->lock);
+		if (!write_same->cbk_finished)
+			pthread_cond_wait(&write_same->cond, &write_same->lock);
+		write_same->cbk_finished = false;
+		pthread_mutex_unlock(&write_same->lock);
+
+		start_lba++;
+	}
+
+	ret = SAM_STAT_GOOD;
+
+out:
+	aio_command_finish(dev, cmd, ret);
+	pthread_cond_destroy(&write_same->cond);
+	pthread_mutex_destroy(&write_same->lock);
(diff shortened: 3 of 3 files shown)


bgly:
Nice, this has been on my todo list.

tcmur_cmd_handler.c
```diff
@@ -144,6 +144,124 @@ static int write_work_fn(struct tcmu_device *dev,
...
+	if (!blocks) {
+		tcmu_err("The WSNZ = 1 & WRITE_SAME blocks = 0 is not supported!\n");
+		return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+					   ASC_INVALID_FIELD_IN_CDB,
+					   NULL);
+	}
+
+	/*
+	 * The MAXIMUM WRITE SAME LENGTH field in Block Limits VPD page (B0h)
+	 * limit the maximum block number for the WRITE SAME.
```
bgly: You maybe want to try to move this into check_lba_and_length and also utilize that function?

tcmur_cmd_handler.c
```diff
@@ -747,6 +865,9 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	case WRITE_VERIFY_16:
 		ret = handle_write_verify(dev, cmd);
 		break;
+	case WRITE_SAME_16:
```
bgly: Might as well also add the WRITE_SAME 0x41 (10), WRITE_SAME_32 0x0d (32) since its all pretty much the same code.

tcmur_cmd_handler.c
```diff
@@ -144,6 +144,124 @@ static int write_work_fn(struct tcmu_device *dev,
...
+	 * For now not support UNMAP bit and ANCHOR bit.
+	 */
+	if (cdb[1] & 0x10 || cdb[1] & 0x08) {
+		tcmu_err("The ANCHOR or UNMAP bit is set not support!\n");
+		return tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+					   ASC_INVALID_FIELD_IN_CDB,
+					   NULL);
+	}
+
+	cmd->done = handle_write_same_16_null_cbk;
```
bgly: Just a question, but does WRITE_SAME zero work with qcow implemented this way? I am not sure....

tcmur_cmd_handler.c
lxbsz: Yes, I think this could be done in the following patches. For now I only test the _16 case.

tcmur_cmd_handler.c
lxbsz: I only test the ESXI 5.5 case and it works.

tcmur_cmd_handler.c
lxbsz: But the check_lba_and_length() couldn't be used here.
For the Write Same commands and for now the iov_cnt will be 1, and data_length in Data Out buffer will be only on block size (?). The VPD_MAX_WRITE_SAME_LENGTH is num_lbas/blocks it intend to write to.

api.c
```diff
@@ -550,6 +561,8 @@ int tcmu_emulate_evpd_inquiry(
 	}
 	break;
 	default:
+		tcmu_err("Vital product data page code 0x%x not support\n",
```
mikechristie: Can you use the dev loggers when possible? tcmu_dev_err(dev, ...) here and tcmu_dev_dbg below.

I have a patch that converts everyone, but for new cases it would help if you can use them.

tcmur_cmd_handler.c
```diff
@@ -144,6 +144,124 @@ static int write_work_fn(struct tcmu_device *dev,
...
+	uint32_t lba_cnt = tcmu_get_xfer_length(cmd->cdb);
+	uint64_t start_lba = tcmu_get_lba(cmd->cdb);
+	uint64_t end_lba;
+	uint32_t max_lbas, copy_lbas;
+	uint32_t max_sectors;
+	int ret;
+
+	end_lba = start_lba + lba_cnt;
+
+	max_sectors = tcmu_get_attribute(dev, "hw_max_sectors");
```
mikechristie: Shoot. What is hw_max_sectors? In tcmu_emulate_evpd_inquiry, it looks like the max number of 512 byte sectors.

I think  tcmu_emulate_evpd_inquiry is wrong. From the kernel target_core_iblock use  it looks like it is supposed to be max number of block_size sectors, right? So your code is right and we need to fix  tcmu_emulate_evpd_inquiry.

tcmur_cmd_handler.c
```diff
@@ -144,6 +144,124 @@ static int write_work_fn(struct tcmu_device *dev,
...
+	max_lbas = min(max_sectors, ((uint32_t)(~0U)));
+
+	tcmu_dbg("Start lba: %llu Number of lba:: %hu, end lba: %llu\n",
+		 start_lba, lba_cnt, end_lba);
+
+	/*
+	 * Write contents of the logical block data(from the Data-Out Buffer)
+	 * to each LBA in the specified LBA range.
+	 */
+	while(start_lba < end_lba) {
```
mikechristie: Just a nit. Add space between "while" and "("

tcmur_cmd_handler.c
```diff
@@ -144,6 +144,124 @@ static int write_work_fn(struct tcmu_device *dev,
...
+		copy_lbas = min(max_lbas, lba_cnt);
+
+		if (start_lba + copy_lbas >= end_lba)
+			cmd->done = handle_generic_cbk;
+
+		ret = rhandler->write(dev, cmd, cmd->iovec, cmd->iov_cnt,
+				      tcmu_iovec_length(cmd->iovec, cmd->iov_cnt),
```
mikechristie: I think same comment you got from liuy in the xcopy review applies here.

Also the iovec hack fix comment applies here too.

api.c
lxbsz: Yes, will be update all of them.

tcmur_cmd_handler.c
lxbsz: Yes

tcmur_cmd_handler.c
lxbsz: Yes it is.

I'll update them both later.

Thanks very much for liuy's advice.

tcmur_cmd_handler.c
lxbsz: Sorry for late.

As I know for bdevs, the hw_max_sectors usually will be decided by the limitation of the bdevs controller's DMA, which will allow a maximum sectors/blocks to be transfered each IO. This will be reduce the IO merge count or split count.

For the file, this will be similar and more or less decided by the filesystem's limitation.

lxbsz: And in vpd_b0 the maximum transfer length should be hw_max_sectors.

But in inquiry, it is not exactly the maximum of sectors, maybe less/equal/more than/to/than maximum of sectors. 

In kernel iblock, it is what as we see, it's the hw_max_sectors sectors(the sector size is block_size).

lxbsz:
sg3_utils test case:

The command:
# sg_write_same --in=/dev/zero --lba=10 --num=100 /dev/sdc

The log:
[DEBUG] write_same_16_work_fn:177 yourlun: Start lba: 10 Number of lba:: 100, end lba: 110

tcmur_cmd_handler.c
mikechristie: Don't worry about replying late. It is no problem. I am most guilty of slow replies :)

Agree on the bdev info.

I was more asking about what Andy intended vs how you were using it. If I understand your second comment, we agree that hw_max_sectors == max_xfer_length and it should be in block_sized units and we do not have to convert between 512 bytes.

I think the 512b conversion we did in the vpd code was a carry over where in many places in the kernel sectors means 512b units and we have to swap back and forth. The user will not know this, so when they are specifying hw_max_sectors in their target config files, they are just going to specify number of block_sized sectors.

Here is a patchset that fixes that:
https://github.com/open-iscsi/tcmu-runner/pull/147

tcmur_cmd_handler.c
lxbsz: @bgly 
I have manuall add the WRITE_SAME 0x41 (10) and WRITE_SAME (32) support just to do the test, and for the current code, tested using the sg_utils, the WRITE_SAME 0x41 (10) could work:

[root@localhost ~]# sg_write_same -S --in=/dev/zero --lba=10 --num=100 /dev/sdb --verbos
[root@localhost ~]# sg_write_same -R --in=/dev/zero --lba=10 --num=100 /dev/sdb --verbos

But the WRITE_SAME_32 0x0d (32) maybe need some extra code.

[root@localhost ~]# sg_write_same -T --in=/dev/zero --lba=10 --num=100 /dev/sdb --verbos
Write same: pass through os error: Invalid argument
Write same(32) command failed
[root@localhost ~]#

tcmur_cmd_handler.c
lxbsz: Yes, that also what I understand about this.

The hw_max_sectors is just about how many block_size sectors, and the block_size mostly will be 512, but not have too.

lxbsz:
@mikechristie The same with XCOPY, currently is also the sync way.

bgly:
Do you have to force it to be sync? I was comparing FORMAT UNIT to WRITE SAME. Issuing a FORMAT UNIT to a 10GB file took under a minute, and with your WRITE SAME code it took 10mins. That means if a client were to ever issue a WRITE SAME TO a disk of 100GB it would take a very long time....

mikechristie:
The xcopy and unmap PRs have the same issue. We were saying to get the scsi spec and esx use parts correct initially, we could do the simple approach and do it sync. In the next set of patches we could optimize and modify the lower level code to do async like format unit, caw, etc.

mikechristie:
Also now that we know how most of these more compound commands are going to be implemented, we can fix up the iovec handling properly and then when we covert these commands to async, we will not have to do iovec hacks.

bgly:
@mikechristie works for me

lxbsz:
Yes, for now I'm addressing and testing the issues mentioned before in this and other PRs.

They all will be as enhancement patches later.

libtcmu_common.h
```diff
@@ -53,6 +53,8 @@ struct tcmulib_cmd;
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
+#define VPD_MAX_WRITE_SAME_LENGTH 0xFFFF
```
mikechristie: Where did this limit come from? Is it a vmware limit or default? It seems like we have 2 more bytes in the number of logical blocks field for write same 16.

lxbsz: The ESXI's clone also will use the Write Same command some time, and for now the write same is a sync way, so just limit it to 64K, make sure the other commands won't cost too much time to wait Write Same to complete, this will be update later with enhancement patches.

mikechristie:
Thanks!

PR #142, opened 2017-05-11
README: Update option 2 about writting a tcmu-runner plugin handler

lxbsz:


mikechristie:
Thanks.

PR #144, opened 2017-05-11
crash fix + load_module correction + logging improvements

pkalever:
* double free crash fix
* load_module() fix
* log dir path option
* enable gfapi logging

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -33,6 +33,11 @@
 #define ALLOWED_BSOFLAGS (O_SYNC | O_DIRECT | O_RDWR | O_LARGEFILE)
 
 #define GLUSTER_PORT "24007"
+#define TCMU_GLFS_LOG_FILENAME "tcmu-runner-glfs.log"  /* MAX 32 CHAR */
+#define TCMU_GLFS_DEBUG_LEVEL  4
+
+/* tcmu log dir path */
+extern char *tcmu_log_dir;
 
 typedef enum gluster_transport {
 	GLUSTER_TRANSPORT_TCP,
@@ -337,6 +342,7 @@ fail:
 static glfs_t* tcmu_create_glfs_object(char *config, gluster_server **hosts)
 {
 	gluster_server *entry = NULL;
+	char logfilepath[PATH_MAX];
     glfs_t *fs =  NULL;
     int ret = -1;
 
@@ -371,6 +377,17 @@ static glfs_t* tcmu_create_glfs_object(char *config, gluster_server **hosts)
 		goto unref;
 	}
 
+	ret = tcmu_make_absolute_logfile(logfilepath, TCMU_GLFS_LOG_FILENAME);
+	if (ret < 0) {
+		tcmu_err("tcmu_make_absolute_logfile failed: %m\n");
+		goto unref;
+	}
+
+	ret = glfs_set_logging(fs, logfilepath, TCMU_GLFS_DEBUG_LEVEL);
+	if (ret < 0) {
+		tcmu_err("glfs_set_logging failed: %m\n");
+		goto unref;
+	}
 
 	ret = glfs_init(fs);
 	if (ret) {
@@ -378,7 +395,7 @@ static glfs_t* tcmu_create_glfs_object(char *config, gluster_server **hosts)
 		goto unref;
 	}
 
-    return fs;
+	return fs;
 
  unref:
 	gluster_cache_refresh(fs, config);
--- a/glfs.c
+++ b/glfs.c
@@ -247,18 +247,18 @@ static void gluster_cache_refresh(glfs_t *fs, const char *cfgstring)
 	}
 }
 
-static void gluster_free_server(gluster_server *hosts)
+static void gluster_free_server(gluster_server **hosts)
 {
-	if (!hosts)
+	if (!*hosts)
 		return;
-	free(hosts->volname);
-	free(hosts->path);
+	free((*hosts)->volname);
+	free((*hosts)->path);
 
-	gluster_free_host(hosts->server);
-	free(hosts->server);
-	hosts->server = NULL;
-	free(hosts);
-	hosts = NULL;
+	gluster_free_host((*hosts)->server);
+	free((*hosts)->server);
+	(*hosts)->server = NULL;
+	free(*hosts);
+	*hosts = NULL;
 }
 
 /*
@@ -328,7 +328,7 @@ static int parse_imagepath(char *cfgstring, gluster_server **hosts)
 	return 0;
 
 fail:
-	gluster_free_server(entry);
+	gluster_free_server(&entry);
 	free(origp);
 
 	return -1;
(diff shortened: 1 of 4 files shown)


pkalever:
@mikechristie @lxbsz  please-review

glfs.c
```diff
@@ -385,8 +386,8 @@ static glfs_t* tcmu_create_glfs_object(char *config, gluster_server **hosts)
 
  fail:
 	gluster_free_server(entry);
-
-    return NULL;
+	*hosts  = NULL;
```
lxbsz: Here are two spaces before '=' by mistake.

glfs.c
```diff
@@ -329,6 +329,7 @@ static int parse_imagepath(char *cfgstring, gluster_server **hosts)
 
 fail:
 	gluster_free_server(entry);
+	*hosts = NULL;
```
lxbsz: At the last of gluster_free_server(), there has one "hosts = NULL;", I think it intended to do this, but it just pasre one pionter of entry, may it should be "&entry" ?

main.c
```diff
@@ -675,6 +678,27 @@ static void dev_removed(struct tcmu_device *dev)
...
+      tcmu_err("mkdir(%s) failed (%s)", path, strerror(errno));
+      return FALSE;
+    }
+  } else {
+    tcmu_err("opendir(%s) failed (%s)", path, strerror(errno));
+    return FALSE;
+  }
+
+  return TRUE;
+}
```
lxbsz: Code style problems ?

main.c
```diff
@@ -675,6 +678,27 @@ static void dev_removed(struct tcmu_device *dev)
 	free(rdev);
 }
 
+static bool
+tcmu_logdir_create(const char *path)
+{
+  DIR* dir = opendir(path);
+
+
```
lxbsz: Find one redundant white line :)

main.c
```diff
@@ -675,6 +678,27 @@ static void dev_removed(struct tcmu_device *dev)
...
+  if (dir) {
+    closedir(dir);
+  } else if (errno == ENOENT) {
+    if (mkdir(path, 0755) == -1) {
+      tcmu_err("mkdir(%s) failed (%s)", path, strerror(errno));
```
lxbsz: The log message should be end with '\n', this will be helpful for the std/file type loggers.

main.c
```diff
@@ -456,39 +456,62 @@ static void dbus_name_lost(GDBusConnection *connection,
...
-
-		if (err != 0) {
-			tcmu_err("kmod_module_probe_insert_module() for %s failed\n",
-			    kmod_module_get_name(mod));
-			return -1;
-		}
+		state = kmod_module_get_initstate(mod);
+		switch (state) {
+		case KMOD_MODULE_BUILTIN:
+			tcmu_info("Module '%s' is builtin", kmod_module_get_name(mod));
```
mikechristie: We are trying to follow 80 col limit. Logging strings can go past 80 so it is easier to match, but above you would put the kmod_module_get_name on a new line. Also do that below and other places in the patch. I won't blow up your inbox with comment for each instance :)

main.c
```diff
@@ -456,39 +456,62 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+			break;
 
-		tcmu_dbg("Module %s inserted (or already loaded)\n", kmod_module_get_name(mod));
+		case KMOD_MODULE_LIVE:
+			tcmu_dbg("Module '%s' is already loaded", kmod_module_get_name(mod));\
```
mikechristie: Extra "\".

main.c
```diff
@@ -456,39 +456,62 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+			break;
 
+		default:
+			err = kmod_module_probe_insert_module (
+			      mod, KMOD_PROBE_APPLY_BLACKLIST, NULL, NULL, NULL, NULL);
```
mikechristie: I think this does not follow any coding style in the existing code. Pick something you see in the code and follow it :)

I think for super long names we do:

err = kmod_module_probe_insert_module (mod, 
                                                KMOD_PROBE_APPLY_BLACKLIST,
                                                NULL, NULL, NULL, NULL);

where we just tab over.

main.c
```diff
@@ -456,39 +456,62 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+
+			if (err == 0) {
+				tcmu_info("Inserted module '%s'", kmod_module_get_name(mod));
+			} else if (err == KMOD_PROBE_APPLY_BLACKLIST) {
+				tcmu_info("Module '%s' is blacklisted", kmod_module_get_name(mod));
```
mikechristie: I did not get why we do not return a error here? This case would mean the module failed to load right?

main.c
```diff
@@ -724,6 +751,16 @@ int main(int argc, char **argv)
 			if (option_index == 1)
 				handler_path = strdup(optarg);
 			break;
+		case 'l':
+			if (strlen(optarg) > PATH_MAX - 32) {   /* 32 is for filename */
```
mikechristie: Could you just add a define for the log filename max.

glfs.c
```diff
@@ -371,22 +378,34 @@ static glfs_t* tcmu_create_glfs_object(char *config, gluster_server **hosts)
 		goto unref;
 	}
 
+	ret = snprintf(logfilepath, PATH_MAX, "%s/%s",
+	               tcmu_log_dir?tcmu_log_dir:DEFAULT_LOG_DIR, TCMU_GLFS_LOG_FILENAME);
```
mikechristie: Instead of having the glfs handler module know this many details about the runner logging code, I think it would be nicer to make s function that handles this.

glfs.c
pkalever: sorry for being lazy :-)

main.c
pkalever: The problem here is, I prefer to use tab with as 2 in my .vimrc :-)
I think so that killed it all, and are you following tab = 4 ? Hope so. Will make changes on it then.

This is always with me working with different coding styles. It will be nice to see tab expand to spaces, to avoid ...

main.c
pkalever: I think you mean "\"

pkalever: Hey did you see \ is escaped.
it is extra "\\"

main.c
pkalever: Yeah, it make sense to error out.

pkalever:
All good comments, incorporated :+1:

main.c
mikechristie: We use 8 space tabs like the linux kernel. I think since we have to bounce back and forth due to also supporting target_core_user, it is easier to have similar styles.

mikechristie:
@ lxbsz, the patches look ok to me, and I think all your review comments are now handled.

Let me know if you have any issues still. Will give it a day. Let me know if you need more time though.

mikechristie:
Thanks.

PR #147, opened 2017-05-13
fix IO max xfer limits

mikechristie:
This patchset fixes the calculation for the max xfer size and allows handlers like rbd to set their preferred value.

### DIFF
--- a/api.c
+++ b/api.c
@@ -35,8 +35,6 @@
 #include "libtcmu_priv.h"
 #include "alua.h"
 
-#define SECTOR_SIZE 512
-
 int tcmu_get_cdb_length(uint8_t *cdb)
 {
 	uint8_t group_code = cdb[0] >> 5;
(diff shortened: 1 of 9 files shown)


api.c
```diff
@@ -503,10 +503,20 @@ int tcmu_emulate_evpd_inquiry(
...
+		 */
+		if (dev->max_xfer_len) {
+			max_xfer_length = dev->max_xfer_len / block_size;
+		} else {
+			max_xfer_length = tcmu_get_attribute(dev,
+							     "hw_max_sectors");
+			if (max_xfer_length < 0) {
+				return tcmu_set_sense_data(sense,
+						ILLEGAL_REQUEST,
+						ASC_INVALID_FIELD_IN_CDB, NULL);
```
lxbsz: Shouldn't this case the sense return:  HARDWARE_ERROR and ASC_INTERNAL_TARGET_FAILURE ?

mikechristie: Yeah, I just blindly copied from original code. Will fix. Thanks.

PR #148, opened 2017-05-13
glfs: implement async read/write/flush versions

pkalever:

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -76,6 +76,17 @@ struct glfs_state {
 	 */
 };
 
+typedef struct glfs_cbk_cookie {
+	struct tcmu_device *dev;
+	struct tcmulib_cmd *cmd;
+	size_t length;
+	enum {
+		TCMU_GLFS_READ  = 1,
+		TCMU_GLFS_WRITE = 2,
+		TCMU_GLFS_FLUSH = 3
+	} op;
+} glfs_cbk_cookie;
+
 struct gluster_cacheconn {
     char *volname;
 	gluster_hostdef *server;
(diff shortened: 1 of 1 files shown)


pkalever:
@mikechristie @lxbsz please-review

glfs.c
```diff
@@ -539,54 +550,122 @@ static void tcmu_glfs_close(struct tcmu_device *dev)
...
 
-	ret = glfs_pwritev(state->gfd, iov, iov_cnt, offset, ALLOWED_BSOFLAGS);
-	if (ret != length) {
-		ret =  tcmu_set_sense_data(cmd->sense_buf, MEDIUM_ERROR,
-					   ASC_WRITE_ERROR, NULL);
-	} else {
-		ret = SAM_STAT_GOOD;
+	cookie = calloc(1, sizeof(*cookie));
+	if (!cookie) {
+		tcmu_err("Could not allocate cookie: %m\n");
```
lxbsz: I think as Mike mentioned before in my some pull requests, we'd better convert to use tcmu_dev_XXX() instead.

The same with the other places.

glfs.c
```diff
@@ -539,54 +550,122 @@ static void tcmu_glfs_close(struct tcmu_device *dev)
...
-			   struct iovec *iov, size_t iov_cnt, size_t length,
-			   off_t offset)
+static int tcmu_glfs_write(struct tcmu_device *dev,
+                           struct tcmulib_cmd *cmd,
+                           struct iovec *iov, size_t iov_cnt,
+                           size_t length, off_t offset)
 {
 	struct glfs_state *state = tcmu_get_dev_private(dev);
-	ssize_t ret;
+	glfs_cbk_cookie *cookie = NULL;
```
lxbsz: Maybe no need to initialize to NULL here.

glfs.c
pkalever: Xiubo Lee, you are right, and that is in my mind too, but I would like to make that as part of an another patch, as it needs changes all over the file and may not be related to the subject of the patch.

So for now, kept it similar with other places.

glfs.c
pkalever: Yeah, right!
The glitch was due to some last minute restructure, will send an another version.

Thanks!

glfs.c
lxbsz: Yes, that sounds nice :)

PR #150, opened 2017-05-13
glfs: implement async read/write/flush versions

pkalever:

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -76,6 +76,17 @@ struct glfs_state {
 	 */
 };
 
+typedef struct glfs_cbk_cookie {
+	struct tcmu_device *dev;
+	struct tcmulib_cmd *cmd;
+	size_t length;
+	enum {
+		TCMU_GLFS_READ  = 1,
+		TCMU_GLFS_WRITE = 2,
+		TCMU_GLFS_FLUSH = 3
+	} op;
+} glfs_cbk_cookie;
+
 struct gluster_cacheconn {
     char *volname;
 	gluster_hostdef *server;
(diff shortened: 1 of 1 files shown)


pkalever:
@mikechristie @lxbsz please-review

@lxbsz Apologies, unintentionally my previous PR got closed due to some bad script (automate github rebase and commit). Hence I have created an another PR. Sorry for loosing your valuable comments there, I have rectified the unwanted NULL initialization, and s/tcmu_err/tcmu_dev_err/ I will make this as a separate patch.

Thanks!

lxbsz:
That's okay for me :)

glfs.c
```diff
@@ -539,54 +550,122 @@ static void tcmu_glfs_close(struct tcmu_device *dev)
...
+	if (!cookie) {
+		tcmu_err("Could not allocate cookie: %m\n");
+		goto out;
+	}
+	cookie->dev = dev;
+	cookie->cmd = cmd;
+	cookie->length = length;
+	cookie->op = TCMU_GLFS_READ;
+
+	if (glfs_preadv_async(state->gfd, iov, iov_cnt, offset, SEEK_SET,
```
mikechristie: Will your async readv/writev  calls modify the iovec's iov_len or iov_base fields? For example for short read/writes would it have partially updated the iov_len field to indicated where it left off?

Just asking, because I am trying to track down what all the handlers do.

mikechristie:
Current patch looks ok to me.

Will give it a day in case lbsz has more comments.

Also, pkalever you do not need to add an extra comment about reviewing. I get a email when you do a pull already. I do not need double :)

lxbsz:
It looks okay for me :)

glfs.c
pkalever: Actually the api doesn't modify the iovec input, it has its own iov that it uses for sending it to lower xlators, so to confirm it is not updating iov_len .

However short writes, it returns the return value such that in failure case it indicates not all data is written/read to the file.

pkalever:
Mike, sure I will keep that in mind :-)
This is how we practice with in gluster, we submit a patch and include the reviewers list in the gerrit review link.

I will try to adopt your practice in tcmu-runner :-)

mikechristie:
Thanks for all the help reviewing lxbsz!

PR #151, opened 2017-05-14
rbd: Add UNMAP/discard support

lxbsz:

### DIFF
--- a/api.c
+++ b/api.c
@@ -334,6 +334,8 @@ int tcmu_emulate_evpd_inquiry(
 	size_t iov_cnt,
 	uint8_t *sense)
 {
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+
 	switch (cdb[2]) {
 	case 0x0: /* Supported VPD pages */
 	{
@@ -647,6 +649,14 @@ finish_page83:
 		 * The logical block data represented by unmapped LBAs is set to zeros
 		 */
 		data[5] = 0x04;
+
+		/*
+		 * The logical block provisioning unmap (LBPU) field.
+		 *
+		 * This will enable the UNMAP command for the device server.
+		 */
+		if (rhandler->unmap)
+			data[5] |= 0x80;
 
 		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, sizeof(data));
 		return SAM_STAT_GOOD;
--- a/api.c
+++ b/api.c
@@ -748,6 +748,14 @@ int tcmu_emulate_read_capacity_16(
 	val32 = htobe32(block_size);
 	memcpy(&buf[8], &val32, 4);
 
+	/*
+	 * Logical Block Provisioning Management Enabled (LBPME) bit
+	 *
+	 * The LBPME bit sets to one and then the logical unit implements
+	 * logical block provisioning management
+	 */
+	buf[14] = 0x80;
+
 	/* all else is zero */
 
 	tcmu_memcpy_into_iovec(iovec, iov_cnt, buf, sizeof(buf));
--- a/api.c
+++ b/api.c
@@ -577,6 +577,14 @@ finish_page83:
 		/* Optimal xfer length */
 		memcpy(&data[12], &val32, 4);
 
+		/* MAXIMUM UNMAP LBA COUNT */
+		val32 = htobe32(VPD_MAX_UNMAP_LBA_COUNT);
+		memcpy(&data[20], &val32, 4);
+
+		/* MAXIMUM UNMAP BLOCK DESCRIPTOR COUNT */
+		val32 = htobe32(VPD_MAX_UNMAP_BLOCK_DESC_COUNT);
+		memcpy(&data[24], &val32, 4);
+
 		/* MAXIMUM WRITE SAME LENGTH */
 		val64 = htobe64(VPD_MAX_WRITE_SAME_LENGTH);
 		memcpy(&data[36], &val64, 8);
@@ -632,6 +640,13 @@ finish_page83:
 		 */
 		val16 = htobe16(0x0004);
 		memcpy(&data[2], &val16, 2);
+
+		/*
+		 * The logical block provisioning read zeros (LBPRZ) field.
+		 *
+		 * The logical block data represented by unmapped LBAs is set to zeros
+		 */
+		data[5] = 0x04;
 
 		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, sizeof(data));
 		return SAM_STAT_GOOD;
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -39,6 +39,9 @@ struct tcmulib_cmd;
 #define CFGFS_ROOT "/sys/kernel/config/target"
 #define CFGFS_CORE CFGFS_ROOT"/core"
 
+#define VPD_MAX_UNMAP_LBA_COUNT            1048576
+#define VPD_MAX_UNMAP_BLOCK_DESC_COUNT     0x04
+
 #define max(a, b) ({			\
 	__typeof__ (a) _a = (a);	\
 	__typeof__ (b) _b = (b);	\
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -1678,6 +1678,287 @@ static int handle_rtpg(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return ret;
 }
 
+struct unmap_state {
+	pthread_mutex_t lock;
+	unsigned int refcount;
+};
+
+struct unmap_descriptor {
+	uint64_t offset;
+	uint64_t length;
+
+	struct tcmulib_cmd *origcmd;
+};
+
+static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd, int ret)
+{
+	struct unmap_descriptor *desc = ucmd->cmdstate;
+	struct tcmulib_cmd *origcmd = desc->origcmd;
+	struct unmap_state *state = origcmd->cmdstate;
+
+	free(desc);
+	free(ucmd);
+
+	pthread_mutex_lock(&state->lock);
+	if (--state->refcount > 0) {
+		pthread_mutex_unlock(&state->lock);
+		return;
+	}
+	pthread_mutex_unlock(&state->lock);
+
+	pthread_mutex_destroy(&state->lock);
+	free(state);
+
+	aio_command_finish(dev, origcmd, ret);
+}
+
+static int unmap_work_fn(struct tcmu_device *dev, struct tcmulib_cmd *ucmd)
+{
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	struct unmap_descriptor *desc = ucmd->cmdstate;
+	uint64_t offset = desc->offset, length = desc->length;
+
+	ucmd->done = handle_unmap_cbk;
+
+	return rhandler->unmap(dev, ucmd, offset, length);
+}
+
+static int handle_unmap_internal(struct tcmu_device *dev, struct tcmulib_cmd *origcmd,
+				 uint16_t bddl, uint8_t *par)
+{
+	struct unmap_state *state = origcmd->cmdstate;
+	uint32_t block_size = tcmu_get_dev_block_size(dev);
+	uint64_t end_lba = tcmu_get_dev_num_lbas(dev) - 1;
+	uint8_t *sense = origcmd->sense_buf;
+	uint16_t offset = 0;
+	int ret, i = 0, refcount = 0;
+
+	/* The first descriptor list offset is 8 in Data-Out buffer */
+	par += 8;
+
+	pthread_mutex_lock(&state->lock);
+	while (bddl) {
+		struct unmap_descriptor *desc;
+		struct tcmulib_cmd *ucmd;
+		uint64_t lba;
+		uint32_t nlbas;
+
+		lba = be64toh(*((uint64_t *)&par[offset]));
+		nlbas = be32toh(*((uint32_t *)&par[offset + 8]));
+
+		tcmu_dev_dbg(dev, "Parameter list %d, lba: %llu, nlbas: %u\n",
+			     i++, lba, nlbas);
+
+		if (nlbas > VPD_MAX_UNMAP_LBA_COUNT) {
+			tcmu_err("Illegal parameter list LBA count %lu exceeds:%u\n",
+				 nlbas, VPD_MAX_UNMAP_LBA_COUNT);
+			ret = tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+						  ASC_INVALID_FIELD_IN_PARAMETER_LIST,
+						  NULL);
+			goto state_unlock;
+		}
+
+		if (lba + nlbas - 1 > end_lba || lba + nlbas < lba) {
+			tcmu_err("Illegal parameter list (lba + nlbas) %llu "
+				 "exceeds last lba %llu\n",
+				 lba + nlbas - 1, end_lba);
+			ret = tcmu_set_sense_data(sense, ILLEGAL_REQUEST,
+						  ASC_LBA_OUT_OF_RANGE,
+						  NULL);
+			goto state_unlock;
+		}
+
+		desc = calloc(1, sizeof(*desc));
+		if (!desc) {
+			tcmu_dev_err(dev, "Failed to calloc desc!\n");
+			ret = tcmu_set_sense_data(sense, HARDWARE_ERROR,
+						  ASC_INTERNAL_TARGET_FAILURE,
+						  NULL);
+			goto state_unlock;
+		}
+
+		ucmd = calloc(1, sizeof(*ucmd));
+		if (!ucmd) {
+			tcmu_dev_err(dev, "Failed to calloc unmapcmd!\n");
+			ret = tcmu_set_sense_data(sense, HARDWARE_ERROR,
+						  ASC_INTERNAL_TARGET_FAILURE,
+						  NULL);
+			free(desc);
+			goto state_unlock;
+		}
+
+		state->refcount++;
+
+		desc->origcmd = origcmd;
+		desc->offset = lba * block_size;
+		desc->length = nlbas * block_size;
+		ucmd->cmdstate = desc;
+
+		ret = async_handle_cmd(dev, ucmd, unmap_work_fn);
+		if (ret != TCMU_ASYNC_HANDLED) {
+			free(ucmd);
+			free(desc);
+			goto state_unlock;
+		} else {
+			refcount++;
+		}
+
+		/* The unmap block descriptor data length is 16 */
+		offset += 16;
+		bddl -= 16;
+	}
+state_unlock:
+	pthread_mutex_unlock(&state->lock);
+
+	/* Will let the cbk to do the release */
+	if (refcount != 0)
+		return ret;
+
+	free(state);
+	return ret;
+}
+
+static int handle_unmap(struct tcmu_device *dev, struct tcmulib_cmd *origcmd)
+{
+	uint8_t *cdb = origcmd->cdb;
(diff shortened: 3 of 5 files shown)


lxbsz:
This patch depend on the XCOPY/WRITE_SAME/.... patches ahead, so here just as one RFC temporarily.

tcmur_cmd_handler.c
```diff
@@ -469,6 +474,216 @@ static int handle_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		/*
+		 * For now for the UNMAP emulator we will emulate the
+		 * fallocate() system call and will write zero to the
+		 * pre-allocated range.
+		 *
+		 * For example, for the file backstore it could pass
+		 * through the UNMAP command and implement its own
+		 * UNMAP callout by using fallocate() system call to
+		 * speed up this.
+		 */
```
mikechristie: I think umap should be a callout instead of override. CAW offload will probably be done by just rbd. unmap seems like every handler will implement it.

tcmur_cmd_handler.c
```diff
@@ -735,6 +950,10 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	case WRITE_16:
 		ret = handle_write(dev, cmd);
 		break;
+	case UNMAP:
+		tcmu_dbg("UNMAP ------------------------------------------------------------------------\n");
```
mikechristie: Probably drop this dbg statement.

tcmur_cmd_handler.c
```diff
@@ -469,6 +474,216 @@ static int handle_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	uint64_t end_lba = tcmu_get_dev_num_lbas(dev) - 1;
+	struct iovec iovec;
+	size_t iov_cnt;
+	uint8_t *sense = cmd->sense_buf;
+	uint8_t *par;
+	uint16_t bddl = unmap->bddl, offset = 0;
+	void *iov_base;
+	int ret, i = 0;
+
+	iov_base = calloc(1, block_size);
```
mikechristie: I think the is missing a free of the iov_base.

tcmur_cmd_handler.c
lxbsz: Yes, just for debuging and forgot to delete.
Thanks

tcmur_cmd_handler.c
lxbsz: Will be fixed.

 I just test  it by using sg3 utils, and I haven't any ESXI environmemt at home, will do a verify in ESXI environmemt later.

tcmur_cmd_handler.c
lxbsz: I am not sure about glustrfs and qcow.

The file and the rbd are all could and needed.

So here, I just create one RFC, and also want to get some ideas about this.

mikechristie: gluster has a unmap/discard command. qcow does too, but I think it is more complicated when supporting all the formats.

I think we want to end up where the SCSI command parsing like looping over param/descr lists is done here and calls a unmap callout for each chunk, because I do not think any one can do vectored/sg unmaps (they can just do the one start to end chunk in each call).

The code that does the write zeros based emulation below could then go into a lib function., so each handler could just call or set its unmap callout to the lib function initially. As we start to convert handlers to native unmap calls then they can just drop the lib call.

The write same code can then call the same callout if the unmap bit is set.

api.c
```diff
@@ -521,6 +521,14 @@ int tcmu_emulate_evpd_inquiry(
...
+		val32 = htobe32(VPD_MAX_UNMAP_LBA_COUNT);
+		memcpy(&data[20], &val32, 4);
+
+		/* MAXIMUM UNMAP BLOCK DESCRIPTOR COUNT */
+		val32 = htobe32(VPD_MAX_UNMAP_BLOCK_DESC_COUNT);
+		memcpy(&data[24], &val32, 4);
+
 		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, sizeof(data));
 
 		return SAM_STAT_GOOD;
```
mikechristie: I think in the logical provisioning VPD page we would want to set the LBPRZ field to 1b right?

lxbsz: Yeah, it is.

And together with the LBPU fileld will set later in another patch.

lxbsz:
There are still some work to do to enable the UNMAP for ESXI.

For example the LBPME and LBPRZ bits in READ_CAPACITY_16 command needed to be set.

tcmur_cmd_handler.c
lxbsz: Yes, agree and sounds nice.

How about to do this after all of these ? For now all about XCOPY/WS/UNMAP patches has been test using sg3_utils and ESXI 5.5 and are all stable for now.

For current UNMAP emulator will just zero all the lbas without discarding them. Though for now for the current hanlders they could all implemented their own discard ways, but maybe in future some new handler could not ? So IMO, the generic routine here could be reserved.

mikechristie: > How about to do this after all of these ? For now all about XCOPY/WS/UNMAP patches has been test using sg3_utils and ESXI 5.5 and are all stable for now.

Sounds ok.

> For current UNMAP emulator will just zero all the lbas without discarding them. Though for now for the current hanlders they could all implemented their own discard ways, but maybe in future some new handler could not ? So IMO, the generic routine here could be reserved.

Do you mean we could do something like

if (rhander->unmap)
     rhandler->unman()
else
    emulated write zeros based unmap

Then yeah, I agree.

lxbsz: Yes, it is.

superlibj:
Updated by  setting  LBPME and LBPRZ bits in READ_CAPACITY_16.

The LBPME is a must for EXSI to work.

Had intended to create one new PR about this, but since this and the two early patches about unmap will togther effect the EXSI, so just apend it here.

api.c
mikechristie: You also need to set the page length to 0x3c when setting the thinp related values. If thinp is not used them it is just 0x10.

Linux' checks for this in sd_read_block_limits().

tcmur_cmd_handler.c
```diff
@@ -469,6 +474,247 @@ static int handle_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	uint8_t *sense = cmd->sense_buf;
+	uint8_t *par;
+	uint16_t dl, bddl;
+	int ret;
+
+	/*
+	 * ANCHOR bit check
+	 *
+	 * The ANCHOR in the Logical Block Provisioning VPD page is not
+	 * suppoted, so the ANCHOR bit shouldn't be set here.
```
mikechristie: Missing a "r" in supported.

api.c
lxbsz: Yes, it is. While the page length has already set to 0x3c for VPD 0xB0 by Andy before.

tcmur_cmd_handler.c
lxbsz: Will be updated.

lxbsz:
@mikechristie 
This has been update as supporting rbd.c only for now.
Please review.

lxbsz:
Tested this using sg_untils:

The command:
[root@premote ~]# /usr/local/bin/sg_unmap --lba=0x10 --num=4000 /dev/sdb -v
    unmap cdb: 42 00 00 00 00 00 00 00 18 00

The logs:
2017-07-19 17:53:19.685 4307 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1056 : 42 0 0 0 0 0 0 0 18 0 
2017-07-19 17:53:19.689 4307 [DEBUG] tcmu_rbd_discard:758 mylun1G: Data-Out Buffer Length: 24, dl: 22, bddl: 16
2017-07-19 17:53:19.689 4307 [DEBUG] tcmu_rbd_discard:816 mylun1G: Parameter list 0, lba: 16, nlbas: 4000
2017-07-19 17:53:19.692 4307 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1056 : 0 0 0 0 0 0

lxbsz:
Another test by using mkfs.xfs:

The command:
[root@promote ~]# mkfs.xfs /dev/sdb -f
meta-data=/dev/sdb               isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0, sparse=0
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
[root@promote ~]# 

The logs:
2017-07-19 18:02:25.843 4307 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1056 : 42 0 0 0 0 0 0 0 18 0 
2017-07-19 18:02:25.843 4307 [DEBUG] tcmu_rbd_discard:758 mylun1G: Data-Out Buffer Length: 24, dl: 22, bddl: 16
2017-07-19 18:02:25.843 4307 [DEBUG] tcmu_rbd_discard:816 mylun1G: Parameter list 0, lba: 0, nlbas: 1048576
2017-07-19 18:02:25.843 4307 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1056 : 42 0 0 0 0 0 0 0 18 0 
2017-07-19 18:02:25.843 4307 [DEBUG] tcmu_rbd_discard:758 mylun1G: Data-Out Buffer Length: 24, dl: 22, bddl: 16
2017-07-19 18:02:25.843 4307 [DEBUG] tcmu_rbd_discard:816 mylun1G: Parameter list 0, lba: 1048576, nlbas: 1048576
2017-07-19 18:02:25.843 4307 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1056 : 0 0 0 0 0 0

lxbsz:
@mikechristie Please review.

Thanks,

mikechristie:
I do not want handlers to be doing that much discard work. I think it should be a callout. I am working on ESX support right now. I was just going to fix up your patch.

lxbsz:
Yes, I tried to,  but to do that there still need much other extra work. So in this I just add almost all code in rbd handler to have an example to known what extra work could there will be, and will try to disscuss and adapt it after this. Thanks.

lxbsz:
Such as  moving tcmur_cmd_handler.c to libtcmu.so? And then add the callout to it ?

lxbsz:
After this with your fix up is done, the other ws/xcopy/caw will be much easy to develop because they encountered the same issue.

mikechristie:
> Such as moving tcmur_cmd_handler.c to libtcmu.so? And then add the callout to it ?

Yeah. Most of tcmu_rbd_discard looks like generic scsi processing work.

tcmu_rbd_aio_discard looks like what you would do in a rbd.c discard callout.

I think the way you did it has the benefit we can have multiple discards in flight in parallel, but it looked like we could do something similar but still make it common. The tcmur_cmd_handler could would have a srtuct that tracks how many discards calls were done and when they are completed then we can complete the original.

Something like:

> 

struct unmap_state {
        int refcount;
       tcmulib_cmd *orig;
      pthread lock;
     has_error
};

struct unmap_descriptor {
    u64 offset;
    u64 length;
};

handle_unmap_done(int rc)
{
   lock(unmap_state->lock)
   if (--unmap_state->refcount == 0) {
             if (unmap_state->has_error)
                     aio_command_finish(dev, cmd, unmap_state->has_error)
             else
                  aioio_command_finish(dev, cmd, rc);   
    } else
          if (rc)
             unmap_state->has_error = rc;
   unlock(unmap_state->lock)
}

unmap_work_fn()
{
    rhandler->unmap(dev, cmd, descriptor-offset, descriptor->length);
}

int handle_unmap(... cmd) {
.....
    unmap = malloc(struct discard_state);
    unamp->orig = cmd;
    unmap->refcount = UNMAP cdb's paramter_list_length / 16;
    cmd->cmdstate = discard;
   cmd->done = handle_unmap_done;

    while (bddl) {
         descriptor = malloc(struct unmap_descriptor)
        descriptor->offset = lba * block_size;
        descriptor->length = nlbas * block_size;
       ret =  async_handle_cmd(cmd, unmap_work_fn
       if (ret != TCMU_ASYNC_NOT_HANDLED) {
           goto fail;
       }
    }

fail:
   lock(unmap_state->lock)
   if (unmap->refcount indicates all unmaps have completed)
          set sense and return SAM_STAT_CHECK_CONDITION
  else
    adjust refcount
    set sense and has_error
    unmap->has_error
    return TCMU_ASYNC_HANDLED
   unlock(unmap_state->lock)
}

mikechristie:
Uggggg, sorry for the bad formatting. You get the idea right? If not, let me know and I can email it to you.

lxbsz:
Yes, the above dummy code look nice, and I think I have gotten the idea.

Have you already fix up this ? Or I will do it.

mikechristie:
I have not. Go for it if you have time.

lxbsz:
Sure. Thanks

lxbsz:
@mikechristie 

Changed by this version:
1, temporarily export the tcmur_handle_unmap() from tcmur_cmd_handler.c through main-syms.txt.
2, struct tcmur_handler adds flags support to tell whether the VAAI preemitives is supported or not.
3, for each hander who will support UNMAP, it will provide an unmap_fn hook to tcmur_cmd_handler.c to do the actual unmap.
4, move the generic scsi code to tcmur_cmd_handler.c.

Thanks,

lxbsz:
The mkfs.xfs will unmap all the /dev/sdb block device firstly, and then write the xfs format to it.

Tested by:

[root@promote ~]# mkfs.xfs /dev/sdb -f
meta-data=/dev/sdb               isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0, sparse=0
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
[root@promote ~]# 

The logs:
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 42 0 0 0 0 0 0 0 18 0 
[DEBUG] tcmur_handle_unmap:1905 mylun1G: Data-Out Buffer Length: 24, dl: 22, bddl: 16
[DEBUG] handle_unmap:1753 mylun1G: Parameter list 0, lba: 0, nlbas: 1048576
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 42 0 0 0 0 0 0 0 18 0 
[DEBUG] tcmur_handle_unmap:1905 mylun1G: Data-Out Buffer Length: 24, dl: 22, bddl: 16
[DEBUG] handle_unmap:1753 mylun1G: Parameter list 0, lba: 1048576, nlbas: 1048576
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 0 0 0 0 0 0 
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 28 0 0 0 0 0 0 0 1 0 
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 2a 0 0 0 0 0 0 1 0 0 
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 2a 0 0 0 0 0 0 0 1 0 
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1064 : 2a 0 0 1f ff 0 0 1 0 0 
...

mikechristie:
Ok. I have a better understanding of what you wanted to do.

Ignore other commands and doing generic passthrough for a minute. For just this unmap patch why don't you just add a tcmur_handler->unmap() function like is done for read/write/flush? unmap is a generic operation that will be supported by almost every handler.

In general for how to do passthrough while minimizing how much scsi command processing a handler does the patch's idea seems ok. I just did not get why it is used for unmap vs adding a new callout. We could actually use your design for all commands and I think we could end up simplifying the runner code since it would not have to worry about if the handlers was async or sync. I am not saying to do this. I am just saying it could be done.

lxbsz:
Yes, before as I was understanding and to keep the tcmu handlers something like the posix proto, the .handle_cmd() is just like the ioctl() and will handle what all the open/close/read/write/flush could do.

I am okay if you agree to add a .unmap func instead.

And which one should be used? If the second is better, I will fix it today.

Thanks

mikechristie:
Just add a unmap callout.

If you look at tgt and lio you get a good idea of how many other device/handler/backend/bs (whatever you want to call them) modules implement certain functions and when it would make sense to be common. Pretty much everyone implements unmap/discard and it is a very commonly used command, so we should just do it in a way that is easiest for all handlers and try to limit different/special code paths at the same time.

lxbsz:
I'll update it.

BTW, how about the others, XCOPY/CAW/WS? Also the same like UNMAP will do ?

mikechristie:
I do not know. Case by case.

As far as I know only rbd/ceph offloads CAW. I have not heard of anyone else wanting it or working on it. I was thinking just use handle_cmd since ceph might end up being the only user.

For WS, I am not sure. For now just do handle_cmd. I do not think any handlers will support a generic WS command like ceph. However, for the common case for just writing zeros, I think we can translate it to some device's discards depending on the implementation. Some other handlers could just do the BLKZERO ioctl. If the zeroing case is commonly supported and the generic WS case turns out not to be, then maybe we can just add common code to detect the zero case add a callout named write_zeros().

I do not know about xcopy. As you know since you wrote the code :) the scsi xcopy processing is complicated compared to other commands so we do not want to duplicate it in every handler. But at least ceph has restrictions like it is has to be an entire object.

lxbsz:
@mikechristie  

Updated. Thanks,

lxbsz:
Test outputs:

[root@promote ~]# mkfs.xfs /dev/sdb -f
meta-data=/dev/sdb               isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0, sparse=0
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
[root@promote ~]# /usr/local/bin/sg_unmap --lba=0x10 --num=40000 /dev/sdb -v
    unmap cdb: 42 00 00 00 00 00 00 00 18 00 
[root@promote ~]#

libtcmu_common.h
```diff
@@ -39,6 +39,9 @@ struct tcmulib_cmd;
 #define CFGFS_ROOT "/sys/kernel/config/target"
 #define CFGFS_CORE CFGFS_ROOT"/core"
 
+#define VPD_MAX_UNMAP_LBA_COUNT            1048576
```
mikechristie: Was this the ESX default? Could you add what version of ESX and vmfs it was?

I think in ESX 6.5 with vmfs 6, esx will use the value returned to it, so instead of making the ceph object size match esx's limit, in future patches we can report the ceph object size as a handler limit like we do with max sectors and ESX will use that.

mikechristie:
Thanks!

libtcmu_common.h
lxbsz: Assuming the block size is 512.

Currently I just set the count to 1MB and the max size will be 1MB * 512 = 512MB, which will be larger than 32MB.

From the "VMware vSphere Storage APIs – Array Integration (VAAI).pdf", we can know: the Data Mover is asked to move data in terms of 32MB transfers, and there has not limitation for the UNMAP. So just set size max limitation in UNMAP command to 512MB.

But this is not correct, actually this should depend on device's max_hw_segment_size (here is the size of the object for ceph).

PR #152, opened 2017-05-17
api: mode sense add rwrecovery page support

lxbsz:

### DIFF
--- a/api.c
+++ b/api.c
@@ -711,6 +711,17 @@ int tcmu_emulate_read_capacity_16(
 	return SAM_STAT_GOOD;
 }
 
+int handle_rwrecovery_page(uint8_t *buf, size_t buf_len)
+{
+	if (buf_len < 12)
+		return -1;
+
+	buf[0] = 0x1;
+	buf[1] = 0xa;
+
+	return 12;
+}
+
 int handle_cache_page(uint8_t *buf, size_t buf_len)
 {
 	if (buf_len < 20)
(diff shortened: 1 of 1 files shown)


mikechristie:
Where did the retry count values come from?

And what initiator/OS uses this?

lxbsz:
This will be used by some sg3_utils commands ,such as sg_format.

The retry count I just copied from the file_optical.

mikechristie:
I am not sure if that is correct, but I am not sure what the correct value is because it looks like it would depend on the backend devices.

Runner itself does not do any retries, but for example if a Ceph OSD does a read/write() to a SCSI disk, then you would get 5 retries for medium type errors by the SCSI mid layer. If the OSD were using nvme or some other block device then the retry count would be different. Same for file_example and other handlers that end up using local linux disks or remote machines running linux.

Maybe 0 would be safer so the initiator side is not relying on the target to do N retries and will retry itself in some cases. On the other hand, for say a linux SCSI disk use case the SCSI mid layer will retry 5 times and most likely the physical disk will have already retried M times, so we could end up with too many retries.

What do you think?

lxbsz:
I was not very sure too, so just added the counters in the second version.

I do agree that it depends on the specfied backstores, and the runner itself will without any retries.

lxbsz:
Updated.

mikechristie:
Ok. Looks like we at least have similar behavior to the kernel lio code. It looks like scst sets the retry values to 1, so not too different.

Thanks.

Issue #149, opened 2017-05-13
glfs handler is opening the block file using O_SYNC even though FLUSH op does fdatasync


pranithk:
I am just familiarizing myself with tcmu code and I see that there is an op called FLUSH which is doing fdatasync() for glfs handler. but we are still using O_SYNC in the open/write/read flags.

in glfs.c
#define ALLOWED_BSOFLAGS (O_SYNC | O_DIRECT | O_RDWR | O_LARGEFILE)

I feel it is redundant to have O_SYNC if we are going to anyway get FLUSH operations.

What do you guys think? @mikechristie @pkalever

mikechristie:
Does SYNC and DIRECT mean the same thing in gluster as when you use them to open a block device in linux? If so, then yeah, then doing a SYNCHRONIZE CACHE (tcmu runner handler->flush) call would not be needed.

If you were going to remove the flush callout I think you need this patch, so the initiator knows there are no caches in use.

[runner-cache.txt](https://github.com/open-iscsi/tcmu-runner/files/999888/runner-cache.txt)

pranithk:
@mikechristie gluster-block is a file on glusterfs, so all the flags like O_SYNC|O_DIRECT are same as for a file on linux. So cache won't be used. I think it is better to use cache and SYNCHRONIZE CACHE than to disable it. So I was wondering if we can remove O_SYNC.

mikechristie:
Looks resolved in PR #159

PR #153, opened 2017-05-18
qcow: Add O_SYNC flag

bgly:
This is a requirement for live partition mobility (LPM), to ensure
integrity.

### DIFF
--- a/qcow.c
+++ b/qcow.c
@@ -1437,7 +1437,7 @@ static int qcow_open(struct tcmu_device *dev)
 	tcmu_dbg("%s\n", tcmu_get_dev_cfgstring(dev));
 	tcmu_dbg("%s\n", config);
 
-	if (bdev_open(bdev, AT_FDCWD, config, O_RDWR) == -1)
+	if (bdev_open(bdev, AT_FDCWD, config, O_RDWR | O_SYNC) == -1)
 		goto err;
 	return 0;
 err:


mikechristie:
Hey bgly,

Could you give some more details about why it is needed?

@pkalever and @pranithk were just asking if they can remove its use in glfs in this issue:

https://github.com/open-iscsi/tcmu-runner/issues/149

and I might have then given them some bad info when I said it would be ok to remove O_SYNC.

seroyer:
@mikechristie We need this for PowerVM/ibmvscsis because we have multiple hosts, each running LIO+tcmu-runner with shared storage.  We need to be able to safely transfer the initiator from one target on one host to another target on a different host, but maintain data integrity _almost transparently to the initiator_.  I say almost in that it knows that it will switch hosts at some point, but not exactly when (so cannot stop sending ops).

The goal is to be sure that we are doing write-through to the backstore.  Kernel block and fileio are write-through (or have a setting).  tcmu backstores currently do not, that I can tell...

mikechristie:
Ok makes sense. It sounds like read caching is ok though right? If not then did you need O_DIRECT too?

Here is untested patch so we report the correct cache settings. Since qcow does not implement flush/SYNC CACHE, then it will prevent some errors in the initiator side logs (at least for linux):

[runner-cache-disable-qcow.txt](https://github.com/open-iscsi/tcmu-runner/files/1015067/runner-cache-disable-qcow.txt)

bgly:
Yes @mikechristie, you are correct. Read caching is fine. Thanks for the patch btw on reporting the correct settings. I wasn't aware the settings weren't being reported correctly.

mikechristie:
I am going to merge some other sync cache fixes, and I wanted to get this one, but I can't remember where we left this one.

bgly,  I know you added the emulate_write_cache attr, but did that patch have anything to do with this?

Did emulate_write_cache=0 mean we would report WCE=0 in MODE_SENSE and qcow would use O_SYNC.

I was not sure what to do for emulate_write_cache=1 though. We could report WCE=1, but qcow does not implement SYNCHCRONIZE_CACHE/flush, so I do not think that is safe right? Do we currently always need to use O_SYNC for qcow, so I can just merge the patch you submitted in the PR and the patch to return the correct WCE value.

bgly:
@mikechristie I think for now for the qcow handler we have the WCE=0 mean use O_SYNC flag and when it does WCE=1 then reopen the file without O_SYNC flag. That way it would take care of not having SYNCHRONIZE_CACHE/Flush

mikechristie:
> when it does WCE=1 then reopen the file without O_SYNC flag. That way it would take care of not having SYNCHRONIZE_CACHE/Flush.

For this you mean when we get flush support later then we can add support for reopen and WCE=1 right? Hopefully so because that is what I did. I mean the patches set WCE=0 and uses O_SYNC to be safe for qcow. I did not implement reopen and WCE=1 because i do not have time to implement flush right now. Without you just get a errors for the SYNC CACHE and I was worried the data might be left in a cache as a result.

I made a patchset for all this here:
https://github.com/open-iscsi/tcmu-runner/pull/220

bgly:
Yes, that is what I meant. But do you see any consequences for doing basic flush support now ie: just calling fsync? If not then we can toggle WCE=1/0 without any consequences because the client should call SYNCHRONIZE_CACHE which in turn our calls just use the device's flush op. Then that would also take away the O_SYNC stuff

mikechristie:
When I looked at it the other day, I thought we could just use fsync like file_example. It looked like the qcow write function just did a pwrite so it didn't do anything fancy.

However, I did not have a chance to look at what the cache and decompress code was doing, so I was not sure.

SYNC CACHE is only sent if the initiator sees the WCE bit set, so if you have WCE=0 then you need to emulate write through behavior and need to open with O_SYNC or you could have data in the page cache or disk/ssd and not written to the medium.

mikechristie:
I merged this patch in PR 220:
https://github.com/open-iscsi/tcmu-runner/pull/220

To support the non O_SYNC/WCE=1 case, we can open a new PR or use this one.  It does not matter to me. This one just has a lot of info, so I was not sure if we wanted to just close it.

bgly:
I will abandon this, we were actually about to push up flush patch for qcow. We did some testing to verify it was working as intended. I'll get a separate PR following the flush qcow patch that will utilize the write cache attribute

PR #155, opened 2017-05-21
tcmur: format unit make sure the write length is muti of max_xfer_length 

lxbsz:
For the FORMAT_UNIT write options, the default length is 1MB and
this will round it up to the device's maximum xter length.

Then we can make sure that all the requests/BIO in IO queue is full.


mikechristie:
@bgly, this patch looks ok to me. You are the primary user, so will give you some time to review/test if you want. Let me know if you need more time.

bgly:
Looks good to me! Thanks @lxbsz

mikechristie:
Thanks!

PR #156, opened 2017-05-21
runner: Convert to _dev_ helper and fix debug message

lxbsz:
Also will fix some debug messages who are missing '\n'.

Such as the following, it will be a little hard to read:
[DEBUG] dev_added:630 lun1: Got block_size 512, size in bytes 10737418242017-05-21 11:33:45.844 30321 [DEBUG] file_open:103 : config file/file1


mikechristie:
Thanks.

PR #157, opened 2017-05-21
tcmur: Covert WRITE SAME 16 to async

lxbsz:
For the old code, the WRITE SAME would write the iovec data(from
Data-Out buffer)sector by sector, and it would take to much time
for large size.

This will set the default write length to 1MB, and will round it
up to the device hardware maximum transfer length.

==
This will depend on the PR#155.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -149,87 +149,96 @@ static int write_work_fn(struct tcmu_device *dev,
 }
 
 struct write_same {
-	pthread_cond_t cond;
-	pthread_mutex_t lock;
-	bool cbk_finished;
+	uint64_t cur_lba;
+	uint64_t lba_cnt;
+
+	struct iovec iovec;
+	size_t iov_cnt;
+	void *iov_base;
+	size_t iov_len;
 };
-
-static void handle_write_same_sync(struct tcmu_device *dev,
-				   struct tcmulib_cmd *cmd,
-				   int ret)
-{
-	struct write_same *write_same = cmd->cmdstate;
-
-	pthread_mutex_lock(&write_same->lock);
-	write_same->cbk_finished = true;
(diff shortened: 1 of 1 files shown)


lxbsz:
@mikechristie 

If we will support the WRITE SAME (10) later the VPD_MAX_WRITE_SAME_LENGTH must be keep to 0xFFFF. 
From "SCSI Commands Reference Manual", section 3.73 WRITE SAME (10) command, we can see that the NUMBER OF BLOCKS field only 2 Bytes.

So here without updating of VPD_MAX_WRITE_SAME_LENGTH.

lxbsz:
@bgly 
For your old test case about the WRITE SAME 16 command, the reason why it was much slower than FORMAT UNIT is that the WRITE SAME 16 command was just writing iovec data sector by sector, while the FORMAT UNIT is writing 1MB each time.

So in the current enhancement patch, I have refered to the FORMAT UNIT command, using 1MB length as default too.

lxbsz:
Tested this using sg3_utils's sg_write_same package and ESXI 5.5 creating and cloning VMs.

mikechristie:
Patch looks ok to me. Ran a really basic test. Will run some more later today.

bgly, let me know if you have issues.

For WRITE_SAME_10 and 16 support, if we report support for more than 0xfff does the initiator return error? I would think if it knows we support WS10 and 16 and we report > 0xffff then it would figure things out and know that WS10 is supported up to 0xffff and 16 supports whatever value greater we report over 0xffff.

I guess since we do not implement report opcodes, it does not know what we support during initialization, so it does the safest and does a min on all possible values.

lxbsz:
Yes, that sounds wonderful. Without reporting the limitations then this will be limited by each command's field capacity. And I think this is also what the spec saying.

mikechristie:
Thanks. Will deal with WS10and max write same len in your other PR.

PR #158, opened 2017-05-21
CTRL^C SIGINT + Systemd Unit Changes + darray fix

pkalever:


tcmu-runner.service
```diff
@@ -4,5 +4,5 @@ Description=LIO Userspace-passthrough daemon
 [Service]
 Type=dbus
 BusName=org.kernel.TCMUService1
+KillMode=process
```
bgly: Why remove -Restart=on-failure?

pkalever: OKAY - I do not see any reason why this should be restarted in case if it exist with a failure.
At-least I cannot track if tcmu-runner had restarted in between or running non stopped (With no hacks or dirty scripts). The auto restart still makes many reasons for failure hidden, at-least for me  if I notice an abnormal stop/exit of the daemon, I could have investigate the cause and try to fix it a bit.

@mikechristie Have we discussed the data inconsistency that an auto restart bring in ?
do you remember the exact scenario ?

seroyer: We run this in cloud environments, we need it to recover upon failure with no human interaction.  The failure should be logged in syslog, though.

mikechristie: Right now runner restart on failure does not work for any useful case.

Restart on failure only works if IO is not running, which is probably the least useful and uncommon case runner will fail, so for now I am ok with just removing it from systemd. To implement restart support we need to do:

1. if cmd_time_out > 0, fix target_core_user so that commands that have been timeout by the module have their ring buffer resources like data/cmd buffer memory cleaned up when it is safe. So I think when tcmu-runner restarts it needs to figure out which commands have been failed by the module due to timeout a do this cleanup operation on them.

2. If cmd_timeout == 0, some handlers may need to figure out the status of outstanding commands, and we may need to clean them up on the handler side or resetup the runner state for them, and then we need to clean up the kernel resources like in #1 if we are aborting them.

seroyer: Hmmm...  Granted it's been a while since I tested this, but I have a trusted/reliable test suite that creates continuous I/O load explicitly checking for data integrity.  I ran that tool on a backstore via tcmu-runner and repeatedly killed tcmu-runner.  The test tool ran perfectly detecting no errors...  That was certainly before cmd_timeout support was added, though.

mikechristie: It did work at some point. See Andy's explanation twords the bottom of this email:

https://www.spinics.net/lists/target-devel/msg14773.html

It also might work on your tests now if you do not have a high load and do not fail too many times. If on each run you are not using the entire ring, then when you restart runner, the available ring memory will just be less than the previous run. If you run the tests enough times or were using the entire ring, you will get "tcmu: command timed out" errors and failed commands, because tcmu_queue_cmd_ring is timing out wait for memory to free.

mikechristie:
Thanks.

PR #159, opened 2017-05-21
glfs: remove O_SYNC

pkalever:
O_SYNC makes IO a bit slow and as anyway we are ensuring the data
reaches onto nonvolatile disk with fdatasync() as part of flush op,
O_SYNC seems to be unworthy.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -30,7 +30,7 @@
 
 #include "tcmu-runner.h"
 
-#define ALLOWED_BSOFLAGS (O_SYNC | O_DIRECT | O_RDWR | O_LARGEFILE)
+#define ALLOWED_BSOFLAGS (O_DIRECT | O_RDWR | O_LARGEFILE)
 
 #define GLUSTER_PORT "24007"
 #define TCMU_GLFS_LOG_FILENAME "tcmu-runner-glfs.log"  /* MAX 32 CHAR */


pranithk:
IMO, we should wait for perf results before getting this in.

seroyer:
For ibmvscsis, we actually need write-through for live migration support.  I understand that it might have a performance impact, but if that turns out to be true, can we make it configurable via targetcli instead?

mikechristie:
Yes. Making it configurable with targetcli would be the preferred way.

If it can be translated to an existing LIO attr like device emulate_write_cache, then it would be a simple patch to target_core_user to enable that.

If we need a new setting like target_core_file's write_back, then it is a simple change to target_core_user, but will also require changes to rtslib and targetcli to take in the new setting.

pranithk:
@mikechristie We did our tests with and without O_SYNC on the file and we see 10X performance improvement. I think now we can start the discussion about why was O_SYNC there to begin with. I see that on qcow there are patches for Adding O_SYNC as well. So what would give us the confidence that nothing bad will happen by removing O_SYNC? Since glfs handler uses O_DIRECT, write doesn't get cached until it hits disk-cache. 
@seroyer O_DIRECT is enough for transparency as well for live migration. O_SYNC is needed if the write needs to be durable as soon as the write is issued. I am curious to know the reason why you guys went for O_RDWR|O_SYNC instead of O_RDWR|O_DIRECT in https://github.com/open-iscsi/tcmu-runner/pull/153 it just kills IOPS compared to O_DIRECT.
@mikechristie When does tcmu-runner get flush command? Do let us know the kind of tests we need to make sure this is safe and we can get the results.

mikechristie:
I do not know all the reasons a flush is sent. It depends on the OS, FS, use case, etc.

For linux for example, it is typically sent when a unmount is done, the OS is shutdown/rebooted, device is removed, FS is doing a barrier type of operation like when a journal and metadata update operation, app does a sync/flush call itself.

One question for you guys. Did you not add O_SYNC originally? Was it copied from other code like maybe the original writer of glfs.c saw it in some sort of qemu gluster code? I am just wondering where it came from originally to check if maybe it was needed for another case and it was just copied over.

pranithk:
@mikechristie It is there from the initial commit from Andy (bc0ba10b). We just want to make sure we can remove it safely. @agrover any pointers about why did we add O_SYNC? It seems like O_DIRECT suffices from the info we gathered. It would be really helpful for us to understand if we are missing some detail here.

bgly:
@pranithk How do you know O_DIRECT is enough for Live partition migration? Just wondering since I dont know of a way to verify

pkalever:
Some cross reference could be our glfs driver in qemu, from [1] we can clearly see O_SYNC is not considered/used. In case if cache=none, we open with O_DIRECT.

[1] https://github.com/qemu/qemu/blob/master/block/gluster.c#L715

pranithk:
@bgly What I gathered from the comment from @seroyer on #153  is that the writes need to end up on the shared storage so that the changes can be witnessed from all the hosts. If we open file with just O_RDWR we are saying caching is allowed at all layers. So the writes may not reach the storage layer at all after a write. Both O_SYNC and O_DIRECT tell the layers that caching is not allowed. O_SYNC goes the extra step of doing the write on the device where as with O_DIRECT the writes can still be in the disk cache. At least doing O_DSYNC is better than O_SYNC IMHO. May be I could be completely wrong :-). May be you guys should decide based on https://lwn.net/Articles/457667/ if that helps. All that said, if you find that your solution meets the performance needs with O_SYNC, there is no need to change. The reason why I am bringing up this point in glfs.c is that the performance is just BAD!

We have seen live migrations with Ovirt+Gluster, even in that solution, kvm opens the qcow disk file which is hosted on gluster with O_DIRECT and does fsyncs at different points. Live migration of the VM works fine with this approach, this part is very well tested. I am wondering if tcmu-runner can do the same considering it also has flush() operation which does fdatasync on the backstore in glfs handler.

pranithk:
@mikechristie @pkalever Interesting thing is that gluster.c in qemu most probably made similar from: https://github.com/qemu/qemu/blob/master/block/file-posix.c#L362 

Interesting thing is in tcmu-runner project, I see O_SYNC only for glfs.c and for nothing else. Which makes us even more curious as to why it could be :-)

seroyer:
@pranithk I know we haven't been real clear about how we use LIO, but it is probably different than you expect.  We are not using it for iSCSI targets.  Instead we are using it for ibmvscsis targets.  It's easier to see with a picture, but I'll try to describe it for you.  Imagine you have some shared network storage (could be NFS, could be iSCSI volumes with a cluster aware filesystem on it, etc...).  Then we have multiple Linux hosts running LIO that all see that same common storage.  Live migration for us involves moving a ibmvscsis target from one LIO instance to a different LIO instance.  So we must ensure integrity (with no cooperation from the guest VM!!!) across multiple LIO instances.  We temporarily have the target active on both systems, but have a trigger to ensure all outstanding I/O has completed (or timed out) on the source side LIO before we start actually sending I/O to the new LIO instance.

Whether we need O_SYNC or O_DIRECT (or something else) may vary depending on the details of the storage.  For example, if we use user:qcow on top of IBM GPFS, we don't need to do anything (no O_SYNC, O_DIRECT, or anything) because GPFS handles that integrity magically.  Maybe the same is true for glfs?  I don't know.

pranithk:
@seroyer thanks for the response. At least this clarifies some things. I guess your Last paragraph sums it best. I think to be on the safer side, it is probably better to use O_SYNC as long as you are okay with the performance impact. In case of user:glfs we need at least O_DIRECT and a setting on the volume which honours O_DIRECT. This will guarantee that the data ends up on storage which can be seen by multiple hosts. This way we don't sacrifice performance and still be safe.

@mikechristie What tests would you recommend we do so that we can close this discussion and merge the patch?

mikechristie:
@pranithk, when the glfs backend (I think it is called a brick) gets a write and sends it to the FS it is using, what does it do to make sure the data is written to the medium? It sounds like it uses O_DIRECT to bypass the kernel page cache. How does it make sure the data in the physical device being used by the FS is synced? Does it do a sync/flush after each group of writes sent by a client, or does the FS above the device you are using do it, or do you disable the device's cache or set it to write through? Or, are you relying on the initiator side to send a SYNCHRONIZE_CACHE which calls tcmu_glfs_flush and for glfs_fdatasync you end up doing a flush/sync on brick's fs/device.

pranithk:
@mikechristie We want to rely on the initiator side to send a SYNCHRONIZE_CACHE. Should be fine right?

pranithk:
Sorry for the late response on this. @mikechristie Prasanna and I, did the following test but there are no fdatasync on glusterfs i.e. no SYNCHRONIZE_CACHE command was received by tcmu-runner. Prasanna further added that even after umount, this command is not received. How do you think we should proceed?

mikechristie:
What did you see for the write cache settings for the discovered disk in sysfs and/or its logs?

Is it write back or through? Were you using the patch I sent you guys:

https://github.com/open-iscsi/tcmu-runner/issues/149

Did you modify it so it reports that you are doing write back?

mikechristie:
I think there is actually a bug in the runner code. I saw this on the initiator

Jul 14 02:21:47 rh2 kernel: sd 11:0:0:0: [sdb] Cache data unavailable

so you should look into that.

mikechristie:
Hey @pranithk I am pretty sure our mode page handling code is just detecting that the allocated buffer is too small for the data we want to send and instead of just returning what we can we are returning a failure. Just fix up that code so it works more like the inquiry /evpd code where is sends what it can and it should work.

pranithk:
@mikechristie I found that what you said is what is happening using gdb. alloc_len is just 4 so it is not filling anything and as got_sense is false it says it is illegal request. 

Based on your comment, I added this dummy code just to check if things will work as expected, I am not sure this is what you are expecting, but this was just for experimenting:
```
diff --git a/api.c b/api.c
index 1b8b7aa..3b75b34 100644
--- a/api.c
+++ b/api.c
@@ -861,12 +861,6 @@ int tcmu_emulate_mode_sense(
                        if (ret <= 0)
                                break;
 
-                       if  (sense_ten && (used_len + ret >= 255))
-                               break;
-
-                       if (used_len + ret > alloc_len)
-                               break;
-
                        used_len += ret;
                }
        }
@@ -879,12 +873,6 @@ int tcmu_emulate_mode_sense(
                                if (ret <= 0)
                                        break;
 
-                               if  (!sense_ten && (used_len + ret >= 255))
-                                       break;
-
-                               if (used_len + ret > alloc_len)
-                                       break;
-
                                used_len += ret;
                                got_sense = true;
                                break;

```
With this change, write-cache is enabled:
[ 4051.242032] scsi host2: iSCSI Initiator over TCP/IP
[ 4051.252396] scsi 2:0:0:0: Direct-Access     LIO-ORG  TCMU device      0002 PQ: 0 ANSI: 5
[ 4051.253979] sd 2:0:0:0: Attached scsi generic sg1 type 0
[ 4051.255285] sd 2:0:0:0: [sda] 16777216 512-byte logical blocks: (8.59 GB/8.00 GiB)
[ 4052.772025] sd 2:0:0:0: [sda] Write Protect is off
[ 4052.772034] sd 2:0:0:0: [sda] Mode Sense: 2f 00 00 00
[ 4053.597726] sd 2:0:0:0: [sda] Write cache: enabled, read cache: enabled, doesn't support DPO or FUA
[ 4055.564098] sd 2:0:0:0: [sda] Attached SCSI disk

We receive a flush that got translated into fdatasync on gluster
(gdb) b tcmu_glfs_flush
Breakpoint 2 at 0x7ffff48d59a0
(gdb) c
Continuing.
[Switching to Thread 0x7fffdeefe700 (LWP 5226)]

Thread 12 "tcmu-runner" hit Breakpoint 2, 0x00007ffff48d59a0 in tcmu_glfs_flush ()
   from /usr/local/lib64/tcmu-runner/handler_glfs.so
(gdb) c

 %-latency   Avg-latency   Min-Latency   Max-Latency   No. of calls         Fop
 ---------   -----------   -----------   -----------   ------------        ----
      0.00    3821.00 us    3821.00 us    3821.00 us              1       FSYNC
    100.00  138978.19 us   16173.00 us  329099.00 us            644       WRITE
 
So it is working as we expected it to :-)
Thanks a lot Mike. Sorry it took a while to respond back to you, I am not well versed with this code or what it does, so had to read the scsi reference guide(Still not completely sure I understood it :-) ).

I am not completely sure about what exactly you are asking in the patch though.

pkalever:
On a Quick glance at the code, I'm thinking the 'alloc_len' seems to be getting a wrong representation of data, coz of a bug in tcmu_get_xfer_length() ?

```
diff --git a/api.c b/api.c
index 1b8b7aa..d9b0acf 100644
--- a/api.c
+++ b/api.c
@@ -85,7 +85,7 @@ uint32_t tcmu_get_xfer_length(uint8_t *cdb)
 {
        switch (tcmu_get_cdb_length(cdb)) {
        case 6:
-               return cdb[4];
+               return be16toh(*((uint16_t *)&cdb[4]));
        case 10:
                return be16toh(*((uint16_t *)&cdb[7]));
        case 12:

```
IMO, something like above should solve, @mikechristie @pranithk Thoughts ?

pranithk:
Oh, where will we have details about this representation?

mikechristie:
MODE SENSE 6's allocation len is only 1 byte, so I do not think the patch to tcmu_get_xfer_length would not be correct.

You can find details in the SCSI specs on t10.org. Specifically in the SPC 3 spec in the MODE SENSE and Mode parameters sections.

For pranithk's patch we do not want to return error when we hit that used_len check. However, I am not sure what to return if the used_len would overflow the mode data length field. For example, mode data length is only 1 byte for MODE SENSE 6. Do you just set it to 255 if used_len > 255? I could not find anything in the spec. Do you guys see anything?

I am also not sure why the sense_ten checks are different for the different page_codes. I am trying to find that info the spec.

pranithk:
I googled a bit but didn't find anything useful yet :-(

pkalever:
In File Optical handler https://github.com/open-iscsi/tcmu-runner/blob/master/file_optical.c#L431 
I have noticed 
```
        else {                                                                   
                   [...]
                                if (used_len + ret > alloc_len)                                                                                                                                                    
                                        break;    
                   [...]
        }                                                                        

```
is missing/not considered, in the *_emulate_mode_sense() function, but I'm not sure if it is intentional.

@bgly do you recollect why was this chunk not considered in the file optical handler ?

Thanks in advance :-)

bgly:
@pkalever It was not intentional, do you want to add it in or should i?

mikechristie:
> @pkalever It was not intentional, do you want to add it in or should i?

No. I think there was a miscommunication a couple comments ago.

We are removing it from the api.c code. SPC 4r37 says:

4.3.5.6 Allocation length:

> 
> The device server shall terminate transfers to the Data-In Buffer when the number of bytes or blocks specified by the ALLOCATION LENGTH field have been transferred or when all available data have been transferred, whichever is less. The allocation length is used to limit the maximum amount of variable length data (e.g., mode data, log data, diagnostic data) returned to an application client. If the information being transferred to the Data-In Buffer includes fields containing counts of the number of bytes in some or all of the data (e.g., the PARAMETER DATA LENGTH field, the PAGE LENGTH field, the DESCRIPTOR LENGTH field, the AVAILABLE DATA field), then the contents of these fields shall not be altered to reflect the truncation, if any, that results from an insuffi- cient ALLOCATION LENGTH value, unless the standard that describes the Data-In Buffer format states otherwise.

So file_optical is correct and we should not return a error for that type of issue.

However,  for the other part of the patch
```
-                       if  (!sense_ten && (used_len + ret >= 255))
-                               break;
```
I think this applies:

> If the amount of information to be transferred exceeds the maximum value that the ALLOCATION LENGTH field is capable of specifying, the device server shall transfer no data and terminate the command with CHECK CONDITION status, with the sense key set to ILLEGAL REQUEST, and the additional sense code set to INVALID FIELD IN CDB.

In the test above, because MODE SENSE 6's allocation len is 1 byte we are limited to 255 bytes. So if used_len + ret is greater than that we must return ILLEGAL REQUEST since it is impossible to handle.

The last issue we have been trying to figure out is why for the sense_ten check is different when the page code is 0x3f. I have not had found any reason in the spec.

pranithk:
@mikechristie Based on your comment, this is what I understood.
We should have used !sense_ten instead of sense_ten in both the branches, and
When page_code is not 0x3f these are the changes. 
```
diff --git a/api.c b/api.c
index 1b8b7aa..8ec4fe9 100644
--- a/api.c
+++ b/api.c
@@ -861,7 +861,7 @@ int tcmu_emulate_mode_sense(
                        if (ret <= 0)
                                break;
 
-                       if  (sense_ten && (used_len + ret >= 255))
+                       if  (!sense_ten && (used_len + ret >= 255))
                                break;
 
                        if (used_len + ret > alloc_len)
@@ -882,9 +882,6 @@ int tcmu_emulate_mode_sense(
                                if  (!sense_ten && (used_len + ret >= 255))
                                        break;
 
-                               if (used_len + ret > alloc_len)
-                                       break;
-
                                used_len += ret;
                                got_sense = true;
                                break;

```
But from the spec, it seems like we need to error out if the response is more than 255 for !sense_ten even when the page_code is 0x3f, so I think, this is what we should do? May be that's what you meant, but I am not completely sure. Just this last doubt if you could clarify, then I think I am all good to submit a new PR with the change.

```
diff --git a/api.c b/api.c
index 1b8b7aa..2ba95c5 100644
--- a/api.c
+++ b/api.c
@@ -855,15 +855,15 @@ int tcmu_emulate_mode_sense(
        /* Don't report block descriptors */
 
        if (page_code == 0x3f) {
-               got_sense = true;
                for (i = 0; i < ARRAY_SIZE(modesense_handlers); i++) {
                        ret = modesense_handlers[i].get(&buf[used_len], sizeof(buf) - used_len);
                        if (ret <= 0)
                                break;
 
-                       if  (sense_ten && (used_len + ret >= 255))
+                       if  (!sense_ten && (used_len + ret >= 255))
                                break;
 
+                        got_sense = true;
                        if (used_len + ret > alloc_len)
                                break;
 
@@ -882,9 +882,6 @@ int tcmu_emulate_mode_sense(
                                if  (!sense_ten && (used_len + ret >= 255))
                                        break;
 
-                               if (used_len + ret > alloc_len)
-                                       break;
-
                                used_len += ret;
                                got_sense = true;
                                break;
```

mikechristie:
1. we are executing a MODE_SENSE 6 and used_len + ret >= 255, we need to always return failure. This is for both the 0x3f and non 0x3f case.

2. for both the 0x3f case and non 3f case, if use_len + ret > alloc_len, then we do not return failure unless we hit also

3. there is also the case where our buffer, buf, is too small. In this case, we probably should be returning a ASC_INTERNAL_TARGET_FAILURE. This is hit when the mode sense handler get() function returns -1. We should probably just avoid this case and alloc the buf alloc_len bytes.

I attached a patch which is completely untested and probably does not compile to show what I mean.

[fix-mode-sense.txt](https://github.com/open-iscsi/tcmu-runner/files/1168292/fix-mode-sense.txt)

pranithk:
@mikechristie I will test it and let you know if there are any issues.

pranithk:
@mikechristie I had to make some small changes to the patch and I tested it and it worked fine. I am just finishing up testing with address sanitizer, just to make sure there aren't any other issues we might have introduced. Give me some more time and I will attach the modified patch for you to submit.

pranithk:
@mikechristie Done with my testing. Here is the patch. Let me know if you want to make any other changes. Happy to test. So it doesn't matter if we don't send the values handler popualtes in buf? Even then it is working fine :-). I was a bit surprised to find this actually.

Here is the diff between the patch you sent and the changes I made:

```
20:18:47⚡ diff ~/Downloads/fix-mode-sense.txt fix-mode-sense.txt
2c2
< index b53fd1c..cf4305c 100644
---
> index b53fd1c..11bba92 100644
26c26
< +	       to_len > from_len ? from_len : from_len - to_len);
---
> +	       to_len > from_len ? from_len : to_len);
28c28
< +
---
>  
32c32
<  
---
> +
41c41,42
<  int handle_cache_page(uint8_t *buf, size_t buf_len)
---
> -int handle_cache_page(uint8_t *buf, size_t buf_len)
> +int handle_cache_page(uint8_t *ret_buf, size_t ret_buf_len)
56c57,58
<  static int handle_control_page(uint8_t *buf, size_t buf_len)
---
> -static int handle_control_page(uint8_t *buf, size_t buf_len)
> +static int handle_control_page(uint8_t *ret_buf, size_t ret_buf_len)
80c82
< @@ -828,6 +848,43 @@ static struct {
---
> @@ -828,6 +848,42 @@ static struct {
86c88
< +				 uint8_t *used_len, bool sense_ten)
---
> +				 size_t *used_len, bool sense_ten)
88d89
< +	uint8_t *buf;
124c125
< @@ -846,13 +903,22 @@ int tcmu_emulate_mode_sense(
---
> @@ -846,13 +902,23 @@ int tcmu_emulate_mode_sense(
130a132
> +        uint8_t *orig_buf = NULL;
134c136
< +		return SAM_SAT_GOOD;
---
> +		return SAM_STAT_GOOD;
170c172
< +			if (ret < 0);
---
> +			if (ret < 0)
223c225
< +	tcmu_memcpy_into_iovec(iovec, iov_cnt, orig_buf, sizeof(orig_buf));
---
> +	tcmu_memcpy_into_iovec(iovec, iov_cnt, orig_buf, alloc_len);

```

[fix-mode-sense.txt](https://github.com/open-iscsi/tcmu-runner/files/1170180/fix-mode-sense.txt)

pkalever:
@mikechristie Thanks for the Patch.
@pranithk Thanks for testing this through.

Mike, If you think this patches are ready, will it be possible for you to merge your modesense patch along with O_SYNC patch as head ?

mikechristie:
@pkaleevr, yes I will merge everything at the same time.

@pranithk, could you send the final patch you tested? And yeah, it is weird, but the initiator side can just read the final length value that is calculated and then resend the command with a buffer/allocation length with the correct value. It does not have to continually guess until it finally works.

I just had one review comment for myself. I did not like the temp buf use. tcmu_memcpy_into_iovec would do the right thing so I can just use that directly. I am going to convert to using that. I will adapt that to your patch and test it.

pranithk:
@mikechristie the patch I attached in the last comment is the updated patch, I attached it with same name. Sorry for the confusion :-(

mikechristie:
@pranithk that was my fault. I did not check it :)

I made a PR with the mode sense patch here:
https://github.com/open-iscsi/tcmu-runner/pull/220

I will merge that PR with your patch here when it is all reviewed. I did not convert the patch to use the iovec directly instead of the tmp buffer. Because we need to put the use_len at the very front of all the other page's data, it became too much of a pain. I could just sneak a peak at the first iovec's buffer but that was too hacky. I want to fix up the buffer handling for all these types of commands, but will do that in another patchset, so we can get the protocol fixes in sooner.

PR #160, opened 2017-05-21
discard glfs_check_config + tcmu_err -> tcmu_dev_err

pkalever:


glfs.c
```diff
@@ -448,29 +449,29 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
...
 
 	if (st.st_size != tcmu_get_device_size(dev)) {
-		tcmu_err("device size and backing size disagree: "
-		       "device %lld backing %lld\n",
-		       tcmu_get_device_size(dev),
-		       (long long) st.st_size);
+		tcmu_dev_err(dev, "device size and backing size disagree: "
+		                  "device %lld backing %lld\n",
+		                  tcmu_get_device_size(dev),
+		                  (long long) st.st_size);
```
lxbsz: Shouldn't we align these to first characitor of the "dev," ?

glfs.c
```diff
@@ -430,58 +430,6 @@ static char* tcmu_get_path( struct tcmu_device *dev)
 	return config;
 }
 
-
-static bool glfs_check_config(const char *cfgstring, char **reason)
-{
```
mikechristie: Why do you want to remove it?

pkalever: For now we have these two stages to create a target using targetcli:
1. Create the backend with glfs storage module
    targetcli /backstores/user:glfs create glfsLUN 40G block@10.70.42.151/elastic-media.img
2. Set/Export LUN
    targetcli /iscsi/$IQN/tpg1/luns create /backstores/user:glfs/glfsLUN
    
glfs_check_config() checks, if the target file exists or not at stage 1, as anyway we are doing that as part of tcmu_glfs_open() and ensuring the size as-well, in case if file doesn't exist we fail in stage 2, and we don't succeed target creation.

In reality glfs_check_config() is expensive as it involves network operations :-)

glfs.c
pkalever: Waiting for mike ACK on below, which is important for this patch.
Once mike got convinced with my comment below, will change this for you :-)

glfs.c
mikechristie: All the handler's check_configs could be merged with their open callouts right? If we are going this route, then properly fix up the API and handlers for everyone.

mikechristie: Oh yeah, if you do the code, I can test all the other handlers.

pkalever: I'm not sure if it is a good Idea to remove check_config functionality for all handlers ?

IMO, its up to the handlers to choose, for example rbd did not choose to have check_config, may be because the way they ensure to push the target creation (from their wrapper, gwcli) commands will ensure to either create target or rewind the backend creation in case of failure. 

IMHO, for handlers whose fs calls are local, its not a bad idea to have config_check() as they ensure the config string at the first place.

The only reason that I intend to drop config_check for gluster is that, it involves additional network calls and the cost here might not be worthy when the wrapper tool take care of rewinding the backstore creation.

Let me know your thoughts ?

mikechristie: rbd's lack of implementation was more of a oversight I think. At the time of its implementation, we did not even know about the dbus API use and the person doing the implementation probably thought like you where it is a duplication of open (at least that is what I thought when I reviewed the code for merging). 

The person that did the initial implementation does not use gwcli. It is not required for use with rbd. gwcli also only comes into play on the initial setup. If the link to the rbd device is down during a restart then check_config would be needed to catch this.

> IMHO, for handlers whose fs calls are local, its not a bad idea to have config_check() as they ensure the config string at the first place.

I did not understand the comment about the first place and local handlers. For local handlers like file_* you can just do the equivalent of a check_config in targetcli/rtslib before making the create call like how you would in your rbd/gwcli example.

Agree with you on the cost comment. Here are my issues though:

1. The netlink interface to add device is broken. Runner cannot pass back add/create results to the kernel, so if the handler's open fails, you end up with a partial invalid device, and you do not know it until the initiator scans and sends IO.

2. It sounds like we were using check_config to validate the values that most commonly would cause failures and device addition time. Note: I thought it was supposed to be done in rtslib/targetcli, but I cannot seem to figure out where it is being called.

Ideally, we would just fix #1 and drop the check_config interface as it would then be a duplicate of open and it would work for all cases (initial setup and restart create time and for all apps).

If we are not going to fix it properly, then I think you would at least want to do this in a common place so it works for both the initial setup and restart case. To do that you need to do it in rtslib or runner.

BC-SDS: > rbd's lack of implementation was more of a oversight I think.

No, I just dropped it intentionally because I see it as a duplicate of open and nothing more for rbd. I didn't know why we need it because open could fail then bailout. If we pass open, then we pass .check_config...

liuy: > rbd's lack of implementation was more of a oversight I think

No, I just dropped it intentionally because I see it as a duplicate of open and nothing more for rbd. I didn't know why we need it because open could fail then bailout. If we pass open, then we pass .check_config...

mikechristie: @liuy see comment above yours for why it exists. The netlink interface cannot return failures. If open fails you end up with no device in runner, but you still get a device in the kernel. That then can get reported to initiators in the REPORT LUNS info.

So it seems like a hack. You run check_config first via a dbus call. If it fails, then you know not to call open and you avoid the problem.

Above, I was saying to just handle the problem at open time, we need to fix the netlink interface.

mikechristie: Oh yeah, above I was saying I thought it was supposed to be doing the dbus check_config call in rtslib, but I cannot find it. If it does not exists, I am fine with just saying delete all check_configs now and whoever has time first can fix the netlink interface so it works like how we all expected it to.  There would not be any type of regression then since it has always been busted.

If that is what you guys are saying then that I am ok with it.

mikechristie: Hey, pkalever I am assuming you did not find any uses ofd dbus check_config too right?

I started working on the nl fix. I will post the nl fix patches in another PR when I am done.

Just finish any other changes requested in this patchset like the ones requested from lxbsz and ping me when you have repushed.

Go ahead and assume I am going to finish the nl patches around the same time you repush and just include the glfs check_config removal. I will handle the other ones in my PR.

pkalever: Apologies for delay.

Mike, Yes I did not understand the intention of check_config may be because it is already broken.

Thanks for your very detailed input, as now it seems to make sense about netlink interface implementation for me.

IMO, it is very helpful to document some info about check_config/dbus/netlink intention somewhere in this repo.

> I started working on the nl fix. I will post the nl fix patches in another PR when I am done.

Cool :-)

> Go ahead and assume I am going to finish the nl patches around the same time you repush and just include the glfs check_config removal. I will handle the other ones in my PR.

I'm curious to cross check my understanding here, are you fixing the nl functionality using .open() now, rather it was with .check_config() before ?

mikechristie: I am going to fix the the nl code so runner can pass the kernel status about the nl command. So for TCMU_CMD_ADDED_DEVICE, runner will be able to return failure if the handler's open() fails. That error will then be propagated through the kernel so lio can clean itself up, and to rtslib and whatever tools is using that (targetcli, gwcli, gluster's tool) so they can return an error to the user.

So I am going to just kill check_config and in the open calls, make sure the same checks are done if needed.

glfs.c
mikechristie: @pkalever, were you ready to merge this? It think this is the only outstanding comment.

glfs.c
mikechristie: Just one update. targetcli was doing a CheckConfig call, but I did the kernel/runner changes to allow addition to be able to return errors, so we can still just drop check_config one day.

Here is the kernel patchset:
https://www.spinics.net/lists/target-devel/msg15657.html

Here is a targetcli fix:
https://github.com/open-iscsi/targetcli-fb/pull/87

glfs.c
pkalever: Done Now! Thanks

PR #161, opened 2017-05-21
load_module: defend on modules directory

pkalever:
In some environments (ex: containers) modules dir doesn't exist, in such cases
the load_module job is left to admin.

This patch checks if modules dir exist before a kmod lookup.

### DIFF
--- a/main.c
+++ b/main.c
@@ -40,6 +40,7 @@
 #include <scsi/scsi.h>
 
 #include <libkmod.h>
+#include <sys/utsname.h>
 #include <linux/target_core_user.h>
 #include "darray.h"
 #include "tcmu-runner.h"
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -459,11 +460,51 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+	char *modules_dir;
+
+	modules_dir = get_modules_dirpath();
+	if (!modules_dir) {
+		tcmu_err("get_modules_dirpath() failed\n");
+		return -1;
+	}
+
+	/* check if path entry exist */
+	if (stat(modules_dir, &sb))
```
mikechristie: Move the "{" like this:

if (something) {

main.c
```diff
@@ -459,11 +460,51 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+	{
+		free(modules_dir);
+
+		if(ENOENT == errno) {
```
mikechristie: space between "if" and opening "("

main.c
```diff
@@ -459,11 +460,51 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+	modules_dir = get_modules_dirpath();
+	if (!modules_dir) {
+		tcmu_err("get_modules_dirpath() failed\n");
+		return -1;
+	}
+
+	/* check if path entry exist */
+	if (stat(modules_dir, &sb))
+	{
+		free(modules_dir);
```
mikechristie: If you're not going to log the modules dir in the failure log message just free it once when your done with it:

ret = stat()
free(modules_dir);
if (ret) {
 ....
}

main.c
```diff
@@ -459,11 +460,51 @@ static void dbus_name_lost(GDBusConnection *connection,
...
+	struct stat sb;
+	char *modules_dir;
+
+	modules_dir = get_modules_dirpath();
+	if (!modules_dir) {
+		tcmu_err("get_modules_dirpath() failed\n");
+		return -1;
+	}
+
+	/* check if path entry exist */
```
mikechristie: This is a pretty useless comment. It is clear we are checking for its existing. In the code here, could you add a comment why we need to do this?  Why doesn't kmod lib not work if /lib/modules does not exist? Is it a bug or design decision? Why do we care?

In the git commit message, could you add similar info and why the kmod devs do not want to handle it? If there is a upstream discussion link to it.

This way, if kmod changes or the container setup does, we can drop hacks in our code.

pkalever:
Will make all the suggested changes to this patch :-)

main.c
pkalever: By doing this, errno might be overridden;
Of course even with my previous approach, will do the required.

pkalever:
Hopefully, this is much better approach than previous one :-)

mikechristie:
:) Thanks.

PR #162, opened 2017-05-22
tcmur: Covert XCOPY to async

lxbsz:
For the XCOPY the maximum of the lbas copied each time will be min(src's hw_max_sectors, dst's hw_max_sectors).

[edit]This is different from the FORMAT_UNIT/WRITE_SAME, which are 1MB as default. Because for the XCOPY we want to support and compatible with the case that the src device and dst device have different types, for example src(rbd) --> dst(glfs), they may have different hw_max_sectors size.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -495,10 +495,6 @@ out:
 #define XCOPY_MAX_SECTORS               1024
 
 struct xcopy {
-	pthread_cond_t cond;
-	pthread_mutex_t lock;
-	bool cbk_finished;
-
 	struct tcmu_device *origdev;
 	struct tcmu_device *src_dev;
 	uint8_t src_tid_wwn[XCOPY_NAA_IEEE_REGEX_LEN];
@@ -507,9 +503,15 @@ struct xcopy {
 
 	uint64_t src_lba;
 	uint64_t dst_lba;
-	unsigned long stdi;
-	unsigned long dtdi;
-	unsigned long lba_cnt;
+	uint32_t stdi;
+	uint32_t dtdi;
+	uint32_t lba_cnt;
+	uint32_t copy_lbas;
+
+	void *iov_base;
+	size_t iov_len;
+	struct iovec iovec;
+	size_t iov_cnt;
 };
 
 /* For now only supports block -> block type */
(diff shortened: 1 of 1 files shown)


tcmur_cmd_handler.c
```diff
@@ -1102,22 +1120,46 @@ static int handle_xcopy(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
-	pthread_cond_init(&xcopy->cond, NULL);
-	pthread_mutex_init(&xcopy->lock, NULL);
+	/* Nothing to do with BLOCK DEVICE NUMBER OF BLOCKS set to zero */
+	if (!xcopy->lba_cnt) {
+		ret = SAM_STAT_GOOD;
+		goto finish_err;
+	}
 
+	src_max_sectors = tcmu_get_attribute(xcopy->src_dev, "hw_max_sectors");
+	dst_max_sectors = tcmu_get_attribute(xcopy->dst_dev, "hw_max_sectors");
```
mikechristie: You can just use tcmu_get_dev_max_xfer_len now.

tcmur_cmd_handler.c
```diff
@@ -972,102 +974,118 @@ static int xcopy_parse_parameter_list(struct tcmu_device *dev,
...
+				  int ret)
 {
 	struct xcopy *xcopy = cmd->cmdstate;
+	struct tcmu_device *src_dev = xcopy->src_dev;
+
+	/* write failed - bail out */
+	if (ret != SAM_STAT_GOOD) {
+		tcmu_dev_err(src_dev, "Failed to write to dst device!\n");
+		ret = tcmu_set_sense_data(cmd->sense_buf, MEDIUM_ERROR,
+					  ASC_WRITE_ERROR, NULL);
```
mikechristie: Here and below there is not need to set the sense. The handler will have set it based on whatever error it hit.

tcmur_cmd_handler.c
```diff
@@ -972,102 +974,118 @@ static int xcopy_parse_parameter_list(struct tcmu_device *dev,
...
+	if (ret != TCMU_ASYNC_HANDLED)
+		goto finish_err;
+
+	return;
 
-	pthread_mutex_lock(&xcopy->lock);
-	xcopy->cbk_finished = true;
-	pthread_cond_signal(&xcopy->cond);
-	pthread_mutex_unlock(&xcopy->lock);
+finish_err:
```
mikechristie: This is both the success and failure path right? If so, let's rename the goto so it does not look like a failure path.

tcmur_cmd_handler.c
lxbsz: yes.

tcmur_cmd_handler.c
lxbsz: Okay, will remove this.

tcmur_cmd_handler.c
lxbsz: Right.

Rename it to "out" is okay ?

lxbsz:
Updated.

mikechristie:
Thanks.

PR #163, opened 2017-05-23
tcmur: Add WRITE SAME(10) support and remove the limitation of the max write length

lxbsz:
Since the WRITE SAME(10) and the WRITE SAME(16) are all pretty thesame code, this add the WRITE SAME(10) support.
    
Here just report full size limitation, which will be larger length than WS10(2-Bytes length) could support, and equals to WS16(4-Bytes length) could support. And the client will just figure out depend the specify write same commands and do the right thing.

### DIFF
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -63,7 +63,7 @@ struct tcmulib_cmd;
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
-#define VPD_MAX_WRITE_SAME_LENGTH 0xFFFF
+#define VPD_MAX_WRITE_SAME_LENGTH 0xFFFFFFFF
 
 typedef void (*cmd_done_t)(struct tcmu_device *, struct tcmulib_cmd *, int);
 
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -158,7 +158,7 @@ struct write_same {
 	size_t iov_len;
 };
 
-static int write_same_16_work_fn(struct tcmu_device *dev,
+static int write_same_work_fn(struct tcmu_device *dev,
 				 struct tcmulib_cmd *cmd)
 {
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
@@ -210,7 +210,7 @@ static void handle_write_same_cbk(struct tcmu_device *dev,
 			     write_same->cur_lba, write_lbas);
 	}
 
-	rc = async_handle_cmd(dev, cmd, write_same_16_work_fn);
+	rc = async_handle_cmd(dev, cmd, write_same_work_fn);
 	if (rc != TCMU_ASYNC_HANDLED) {
 		tcmu_dev_err(dev, "Write same async handle cmd failure\n");
 		ret = tcmu_set_sense_data(sense, MEDIUM_ERROR,
@@ -227,7 +227,7 @@ finish_err:
 	aio_command_finish(dev, cmd, ret);
 }
 
-static int handle_write_same_16(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
+static int handle_write_same(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 {
 	uint8_t *cdb = cmd->cdb;
 	uint8_t *sense = cmd->sense_buf;
@@ -335,7 +335,7 @@ static int handle_write_same_16(struct tcmu_device *dev, struct tcmulib_cmd *cmd
 	tcmu_dev_dbg(dev, "First lba: %llu, write lbas: %llu\n",
 		     start_lba, write_lbas);
 
-	return async_handle_cmd(dev, cmd, write_same_16_work_fn);
+	return async_handle_cmd(dev, cmd, write_same_work_fn);
 }
 
 /* async write verify */
@@ -1735,8 +1735,9 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	case WRITE_VERIFY_16:
 		ret = handle_write_verify(dev, cmd);
 		break;
+	case WRITE_SAME:
 	case WRITE_SAME_16:
-		ret = handle_write_same_16(dev, cmd);
+		ret = handle_write_same(dev, cmd);
 		break;
 	case FORMAT_UNIT:
 		ret = handle_format_unit(dev, cmd);


api.c
```diff
@@ -572,10 +571,6 @@ int tcmu_emulate_evpd_inquiry(
 		/* Optimal xfer length */
 		memcpy(&data[12], &val32, 4);
 
-		/* MAXIMUM WRITE SAME LENGTH */
-		val64 = htobe64(VPD_MAX_WRITE_SAME_LENGTH);
-		memcpy(&data[36], &val64, 8);
-
```
mikechristie: This does not seem to work for linux. Zero means no support or use lowest possible safe value it knows about which is 0xffff for both WS10 and WS16. If we report the max we support for WS16 and it is larger than 0xffff, it will just figure things out and do the right thing.

I wonder if the linux behavior is due to buggy targets like us that would report 0x3c for the b0 vpd page, but not actually support WS and so the MAX WS was zero in there.

lxbsz: Sorry for late, just came back from outside yesterday.

I tested this by using EXSI and sg3_utils and could work, and from the Kernel iblock/file emulator it set the limitation to 0xFFFF, the 2-Bytes from the WRITE SAME(10), and could be configurable.

BTW, what does "report 0x3c for the b0 vpd page, but not actually support WS and so the MAX WS was zero in there" mean ? Is the 0x3c value for b0 vpd page's PAGE LENGTH field is incorrect ?

mikechristie: > Sorry for late, just came back from outside yesterday.

No problem. I will try to get your other patchset reviewed this week now too.

For linux, I was not referring to sg utils. I was referring to the kernel block/scsi support, so what you see in drivers/scsi/sd.c and in /sys/block/sdXYZ/queue and /sys/block/sdXYZ/device/scsi_disk.

> BTW, what does "report 0x3c for the b0 vpd page, but not actually support WS and so the MAX WS was zero in there" mean ? Is the 0x3c value for b0 vpd page's PAGE LENGTH field is incorrect ?

It is ok.

I was just speculating on why linux does not support the MAXIMUM WRITE SAME LENGTH = 0 meaning no limit like defined in the spec. Is it a bug in drivers/scsi/sd.c or was it intentional to try and support devices that misreported values. For the latter for example, before your patch to implement WS, we reported the MAX WS LENGTH as zero, but we did not support WS.

lxbsz: From some linux commit logs, I think it may due to that some devices could only support 2-Bytes block counts even with WRITE SAME(16). So set the MAXIMUM WRITE SAME LENGTH to 0xFFFF as default, and each device it could be configurable via sysfs.

If that's why the linux won't setting MAXIMUM WRITE SAME LENGTH to zero as defined in the spec, then the tcmu-runner emulator won't have that limitation, so could it be setting to zero as spec defined ?

mikechristie: With your patch, when we set it to zero, we only get a max of 0xfff, because of this chunk in sd_config_write_same:

`        if (sdkp->max_ws_blocks > SD_MAX_WS10_BLOCKS)
                sdkp->max_ws_blocks = min_not_zero(sdkp->max_ws_blocks,
                                                   (u32)SD_MAX_WS16_BLOCKS);
        else if (sdkp->ws16 || sdkp->ws10 || sdkp->device->no_report_opcodes)
                sdkp->max_ws_blocks = min_not_zero(sdkp->max_ws_blocks,
                                                   (u32)SD_MAX_WS10_BLOCKS);
        else {
                sdkp->device->no_write_same = 1;
                sdkp->max_ws_blocks = 0;
        }`

I had to set it to linux's SD_MAX_WS16_BLOCKS value to get larger WS commands. So with linux you need to report a non zero max for WS10 and WS16. It will then do the right thing. In sd_setup_write_same_cmnd it will then do the right thing WRT lengths of commands and what WS command to use.

I have been testing by checking sysfs and then doing BLKZEROOUT ioctls to /dev/sdXYZ.

For VMware, if you return a zero max length does it to the right thing WRT to the spec?

If you set it to a value greater than 0xffff, does it do something similar to linux where it figures things out?

mikechristie: Oh yeah, for the first part of your comment I had seen the code and commit comments you are referring to. In the git commit it had said:

> We only enable transfers bigger than 0xFFFF if the device explicitly reports MAXIMUM WRITE SAME LENGTH in the BLOCK LIMITS VPD.

And so if the VPD length is 0x3c then the device is reporting a MAX WS LENGTH in my opinion, but maybe I am wrong. If it did not want to report a WS MAX LENGTH setting, then it would set the VPD data length to a shorter value. However, I was not sure if that is what they meant by "explicitly" reporting a value.

My question before was, if it does count as explicitly reporting a valid, did they mean to handle zero as a valid  reported value or did they just miss the part in the specs about zero == unlimited, or does zero not count as a valid reported value.

lxbsz: > Are you saying we should go with your patch where we return 0? If so, for the linux initiator side limitation we would be limited to 0xfff, so are you saying users would need to just adjust that in sysfs or are you saying we should modify the linux initiator side code?

Could also refer this linux commit:
[SCSI] sd: Implement support for WRITE SAME (5db44863b6eb)

We can see one log:
> +       /* Some devices can not handle block counts above 0xffff despite
> +        * supporting WRITE SAME(16). Consequently we default to 64k
> +        * blocks per I/O unless the device explicitly advertises a
> +        * bigger limit.
> +        */

This maybe backwarding compatibility with older or buggy device targets, and the targets could not wholly follow the spec, so the linux code couldn't just set the MAX to zero in vpd.

But I think the users won't have this limitation, so if every thing(such as the initiator) works as the spec defined  and also should be here.

From "SCSI Commands Reference Manual" 
100293068
Rev. D
December 2010

> PAGE LENGTH field
> The PAGE LENGTH field specifies the length of the following VPD page data. The relationship between the PAGE LENGTH field and the CDB ALLOCATION LENGTH field is defined in 2.2.6.
> If the device server supports thin provisioning, then the device server shall set the PAGE LENGTH field to 3Ch.
> If the device server does not support thin provisioning, then the device server may set the PAGE LENGTH field to 10h.

Here it using "may set", not a must or should. So I think the "explicitly" you mentioned should be non-zero and a larger WS MAX LENGTH than default value 0xFFFF.

mikechristie: > Here it using "may set", not a must or should. So I think the "explicitly" you mentioned should be non-zero and a larger WS MAX LENGTH than default value 0xFFFF.

Nice spec find. I agree with you. Let's just return a value larger than 0xffff then.

lxbsz: Full size like 0xFFFF FFFF ？ or just like the linux do by setting it to 8M(0x7FFFFF)?

mikechristie: 0xFFFF FFFF works on linux too right? It just does the min of reported limit and 0x7FFFFF.

Do you know the ESX limit from your testing?

lxbsz: Yes, it should work.

The esxi for my test I think it is 1M.

mikechristie:
Thanks.

PR #164, opened 2017-05-24
[RFC] tcmu: Enable reconfiguration

bgly:
This is a starting patch to begin reconfiguration code.

### DIFF
--- a/file_example.c
+++ b/file_example.c
@@ -76,6 +76,33 @@ static bool file_check_config(const char *cfgstring, char **reason)
 	return true;
 }
 
+static int file_reconfig(struct tcmu_device *dev, uint32_t cfgtype)
+{
+	int64_t dev_size;
+	uint32_t block_size;
+
+	switch (cfgtype) {
+	case No_reconfig:
+		break;
+	case Config_path:
+		tcmu_err("Device Path is an Unsupported Reconfiguration\n");
+		return -EINVAL;
+	case Config_size:
+		block_size = tcmu_get_attribute(dev, "hw_block_size");
+		if (block_size <= 0)
+			tcmu_dev_err(dev, "Could not get hw_block_size\n");
+		dev_size = tcmu_get_device_size(dev);
+		if (dev_size < 0)
+			tcmu_dev_err(dev, "Could not get device size\n");
+		tcmu_set_dev_num_lbas(dev, dev_size / block_size);
+		break;
+	case Config_writecache:
+		return -EINVAL;
+	}
+
+	return 0;
+}
+
 static int file_open(struct tcmu_device *dev)
 {
 	struct file_state *state;
@@ -200,6 +227,7 @@ static struct tcmur_handler file_handler = {
 	.cfg_desc = file_cfg_desc,
 
 	.check_config = file_check_config,
+	.reconfig = file_reconfig,
 
 	.open = file_open,
 	.close = file_close,
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -44,12 +44,15 @@
 static struct nla_policy tcmu_attr_policy[TCMU_ATTR_MAX+1] = {
 	[TCMU_ATTR_DEVICE]	= { .type = NLA_STRING },
 	[TCMU_ATTR_MINOR]	= { .type = NLA_U32 },
+	[TCMU_ATTR_TYPE]	= { .type = NLA_U32 },
 };
 
 static darray(struct tcmu_thread) g_threads = darray_new();
 
 static int add_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring);
 static void remove_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring);
+static int reconfig_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring,
+			   uint32_t cfgtype);
 
 static int handle_netlink(struct nl_cache_ops *unused, struct genl_cmd *cmd,
 			  struct genl_info *info, void *arg)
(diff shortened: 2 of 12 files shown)


libtcmu.c
```diff
@@ -325,6 +335,54 @@ static void close_devices(struct tcmulib_context *ctx)
...
+	if (!dev->handler) {
+		tcmu_err("could not find handler for %s\n", dev->dev_name);
+		goto err_free;
+	}
+
+	/* Handle all configuration changes in check_config */
+	if (dev->handler->check_config &&
+		!dev->handler->check_config(dev->cfgstring, &reason)) {
+		/* It may be handled by other handlers */
+		tcmu_err("check_config failed for %s because of %s\n", dev->dev_name, reason);
```
mikechristie: You can just do tcmu_dev_err(dev,

libtcmu.c
```diff
@@ -325,6 +335,54 @@ static void close_devices(struct tcmulib_context *ctx)
...
+
+	dev->handler = find_handler(ctx, dev->cfgstring);
+	if (!dev->handler) {
+		tcmu_err("could not find handler for %s\n", dev->dev_name);
+		goto err_free;
+	}
+
+	/* Handle all configuration changes in check_config */
+	if (dev->handler->check_config &&
+		!dev->handler->check_config(dev->cfgstring, &reason)) {
```
mikechristie: Just line up the !dev here

if (dev->handler...
    !device->handler

libtcmu.c
```diff
@@ -325,6 +335,54 @@ static void close_devices(struct tcmulib_context *ctx)
...
+		/* It may be handled by other handlers */
+		tcmu_err("check_config failed for %s because of %s\n", dev->dev_name, reason);
+		free(reason);
+		goto err_free;
+        }
+
+	return 0;
+
+err_free:
+	free(dev);
```
mikechristie: I always get the darray used mixed up, but we do not want to free this do we?

libtcmu.c
```diff
@@ -325,6 +335,54 @@ static void close_devices(struct tcmulib_context *ctx)
...
+	}
+
+	dev->handler = find_handler(ctx, dev->cfgstring);
+	if (!dev->handler) {
+		tcmu_err("could not find handler for %s\n", dev->dev_name);
+		goto err_free;
+	}
+
+	/* Handle all configuration changes in check_config */
+	if (dev->handler->check_config &&
```
mikechristie: So the idea is that in a future patch, each handler would detect if this was a check or recheck and for something like updating the cache settings decide if it is supported right?

If it is supported, where do we do the update? A new callout?

bgly: I think we can put all the new code in check_config? 

So if new configuration is found it would update based upon each individual handler, then flush if it needs to / reopen?

Reopen should probably be a new callout?

libtcmu.c
bgly: oops, you are right I dont think im supposed to free here.

bgly:
@mikechristie I think this patch should be ready now... atleast in terms of infrastructure for reconfig support. I added basic reconfig for just dev_size on file_example, but I am not entirely sure how to notify the client... I dont think we can send a UA since we dont have access to the cmd there, so as of now I know the client(atlaest in ubuntu) i can rescan scsi devices to pick up the change.

libtcmu.c
```diff
@@ -325,6 +343,58 @@ static void close_devices(struct tcmulib_context *ctx)
...
+			i++;
+		} else {
+			found = true;
+			break;
+		}
+	}
+
+	if (!found) {
+		tcmu_err("could not reconfigure device %s: not found\n",
+			 dev_name);
```
mikechristie: We can use tcmu_dev_err(dev, here and below.

libtcmu.c
```diff
@@ -325,6 +343,58 @@ static void close_devices(struct tcmulib_context *ctx)
...
+		tcmu_err("could not find handler for %s\n", dev->dev_name);
+		goto err_free;
+	}
+
+	if (!dev->handler->reconfig) {
+		tcmu_err("Reconfiguration is not supported with this device\n");
+		return -EPERM;
+	}
+
+        ret = dev->handler->reconfig(dev, cfgtype);
```
mikechristie: Extra space.

main.c
```diff
@@ -655,6 +656,9 @@ static int dev_added(struct tcmu_device *dev)
 	tcmu_dev_dbg(dev, "Got block_size %ld, size in bytes %lld\n",
 		     block_size, dev_size);
 
+	write_cache_enabled = tcmu_get_attribute(dev, "emulate_write_cache");
```
mikechristie: I think we need a check for if this failed in case it is a older kernel.

libtcmu_priv.h
```diff
@@ -61,6 +61,8 @@ struct tcmu_device {
 	uint64_t num_lbas;
 	uint32_t block_size;
 	uint32_t max_xfer_len;
+	unsigned int  write_cache_enabled:1;
+	unsigned int  read_cache_enabled:1;
```
mikechristie: Extra space between unsigned int and the name of the variable.

main.c
mikechristie: Let's actually drop the write cache handling from this patchset or break it out into separate patches at least.

The problem is if the user requested write cache disabled, for most likely safety reasons, there is only qcow able to handle it in these patches. I can convert glfs and rbd for us (they always want write cache disabled so we can just print an error if someone tries to enable it). I assume you can take file_optical. Either one of us can take file_example.

file_example.c
```diff
@@ -76,6 +76,33 @@ static bool file_check_config(const char *cfgstring, char **reason)
...
+		tcmu_err("Device Path is an Unsupported Reconfiguration\n");
+		return -EINVAL;
+	case Config_size:
+		block_size = tcmu_get_attribute(dev, "hw_block_size");
+		if (block_size <= 0)
+			tcmu_dev_err(dev, "Could not get hw_block_size\n");
+		dev_size = tcmu_get_device_size(dev);
+		if (dev_size < 0)
+			tcmu_dev_err(dev, "Could not get device size\n");
+		tcmu_set_dev_num_lbas(dev, dev_size / block_size);
```
mikechristie: We should make this and the similar code in dev_added into a libtcmu helper. We can then call it here. dev_added and other handlers can use it if that is all they need to do.

mikechristie:
>  I dont think we can send a UA since we dont have access to the cmd there,

You would normally send the UA when you get the next command then. iscsi has the ability to do async sense, but not upstream so ignore this.

Ignore UAs for now in your patches. It is a mess due to the kernel doing some UA work for TMRs, PRs, lun management and alua, but then runner have no support for it. It is a similar problem where we need to figure out how we are going to handle it and where.

main.c
bgly: write_cache is only an attribute in the kernel patch, so if a handler doesn't have implementation for the .reconfig, then it wont do anything. - None the less I will drop all the write_cache stuff for now and make it a separate patch.

PR #170, opened 2017-06-06
tcmu: Add Structure for Reconfiguration Support

bgly:
This patch adds the backbone for reconfiguration of any of the
backstores.

### DIFF
--- a/file_example.c
+++ b/file_example.c
@@ -42,6 +42,7 @@
 
 #include "scsi_defs.h"
 #include "tcmu-runner.h"
+#include "libtcmu.h"
 
 struct file_state {
 	int fd;
@@ -74,6 +75,25 @@ static bool file_check_config(const char *cfgstring, char **reason)
 	unlink(path);
 
 	return true;
+}
+
+static int file_reconfig(struct tcmu_device *dev, uint32_t cfgtype)
+{
+	int ret = 0;
+
+	switch (cfgtype) {
+	case CONFIG_PATH:
+		tcmu_dev_err(dev, "device path reconfiguration is not supported\n");
+		return -EINVAL;
+	case CONFIG_SIZE:
+		ret = tcmu_config_size(dev);
+		break;
+	case CONFIG_WRITECACHE:
+		tcmu_dev_err(dev, "write_cache reconfiguration is not supported\n");
+		return -EINVAL;
+	}
+
+	return ret;
 }
 
 static int file_open(struct tcmu_device *dev)
@@ -200,6 +220,7 @@ static struct tcmur_handler file_handler = {
 	.cfg_desc = file_cfg_desc,
 
 	.check_config = file_check_config,
+	.reconfig = file_reconfig,
 
 	.open = file_open,
 	.close = file_close,
--- a/file_example.c
+++ b/file_example.c
@@ -44,6 +44,10 @@
 #include "tcmu-runner.h"
 #include "libtcmu.h"
 
+#define TCMU_ATTR_DEV_CFG 	4
+#define TCMU_ATTR_DEV_SIZE	5
+#define TCMU_ATTR_WRITECACHE	6
+
 struct file_state {
 	int fd;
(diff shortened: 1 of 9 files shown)


bgly:
As of now rbd wont have a .reconfig thus itll return -EPERM and post tcmu_err("Reconfiguration is not supported with this device\n");

The rest have the handlers have atleast .reconfig function, but I have only Added resize for glfs/qcow/file. 

The next big one is write_cache. 
rbd/glfs - we want write_cache disabled, which is already disabled atm so no changes needed.
It seems like with file_example you can atleast call file_flush, then do you close and open the file again?
file_optical - wont support it
qcow - can we just implement basic flush with fsync then call that as the flush function + close/open the file?

mikechristie:
Let's take this in steps. For the initial merge, what you have is fine.

We do not need to add support for all options for all handlers.

Added some comments for this patch. It looks good. Just some minor nits.

file_example.c
```diff
@@ -76,6 +77,28 @@ static bool file_check_config(const char *cfgstring, char **reason)
 	return true;
 }
 
+static int file_reconfig(struct tcmu_device *dev, uint32_t cfgtype)
+{
+	int ret = 0;
+
+	switch (cfgtype) {
+	case NO_RECONFIG:
```
mikechristie: You can remove this since you check it for it before calling.

file_example.c
```diff
@@ -76,6 +77,28 @@ static bool file_check_config(const char *cfgstring, char **reason)
...
+		tcmu_err("Nothing to reconfigure\n");
+		break;
+	case CONFIG_PATH:
+		tcmu_err("Device Path is an Unsupported Reconfiguration\n");
```
mikechristie: You can drop the per handler error messages and just put it in reconfig_device. When yo do this, try to sync up your style sometimes you capitalize the first letter like above, but below for write cache you didn't.

Also use tcmu_dev_err, so we know what device its coming from.

libtcmu.c
```diff
@@ -325,6 +343,58 @@ static void close_devices(struct tcmulib_context *ctx)
...
+		} else {
+			found = true;
+			break;
+		}
+	}
+
+	if (!found) {
+		tcmu_dev_err(dev, "could not reconfigure device %s: not found\n",
+			     dev_name);
+		goto err_free;
```
mikechristie: You are not freeing anything so probably should modify the label to reflect that, or just drop the gotos and do returns like you do a couple times below. Just sync up on one or the other. Don't mix.

libtcmu.c
```diff
@@ -325,6 +343,58 @@ static void close_devices(struct tcmulib_context *ctx)
...
+	}
+
+	dev->handler = find_handler(ctx, dev->cfgstring);
+	if (!dev->handler) {
+		tcmu_dev_err(dev, "could not find handler for %s\n", dev->dev_name);
+		goto err_free;
+	}
+
+	if (!dev->handler->reconfig) {
+		tcmu_err("Reconfiguration is not supported with this device\n");
```
mikechristie: tcmu_dev_err.

libtcmu.c
```diff
@@ -815,3 +885,23 @@ void tcmulib_cleanup_all_cmdproc_threads()
 		cancel_thread(thread->thread_id);
 	}
 }
+
+int tcmu_config_size(struct tcmu_device *dev) {
```
mikechristie: Move "{" to the next line.

mikechristie:
Hey bgly,

I updated my sync-nl branch to support reconfigure:

https://github.com/mikechristie/tcmu-runner/tree/sync-nl

I did the lib and runner support, and only converted file_example dev_size handling for now. If you convert other handlers send me against the branch.

I also attaching updated kernel patches with reconfig done nl support here:

[sync-send-rfc2.zip](https://github.com/open-iscsi/tcmu-runner/files/1070985/sync-send-rfc2.zip)

PR #171, opened 2017-06-08
glfs: protect cache traffic from race

pkalever:
* fix indentation
* rename global variable cache -> glfs_cache
* protect cache traffic from race

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -25,6 +25,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <scsi/scsi.h>
+#include <pthread.h>
 #include <glusterfs/api/glfs.h>
 #include "darray.h"
 
@@ -38,6 +39,9 @@
 
 /* tcmu log dir path */
 extern char *tcmu_log_dir;
+
+/* cache protection */
+pthread_mutex_t glfs_lock;
 
 typedef enum gluster_transport {
 	GLUSTER_TRANSPORT_TCP,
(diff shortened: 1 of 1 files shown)


glfs.c
```diff
@@ -263,6 +267,38 @@ static void gluster_cache_refresh(glfs_t *fs, const char *cfgstring)
...
+	*fs = gluster_cache_query(entry, config);
+	if (*fs) {
+		*init = false;
+		ret = 0;
+		goto out;
+	}
+
+	*fs = glfs_new(entry->volname);
+	if (!*fs) {
+		tcmu_err("glfs_new failed: %m\n");
```
seroyer: Anything that you call that has a thread cancellation point needs to be taken into account when you lock a mutex.  http://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_09.html  Search in that page for Cancellation Points.  That includes most forms of I/O.  In this case, if this thread was cancelled it could hit a cancellation point here and never unlock the mutex.  Search for uses of pthread_cleanup_push in tcmur_aio.c for example.

Also, isn't tcmu_dev_err more appropriate now?

pkalever: That's a really valid point, thanks for the catch @seroyer, will make the changes and spin soon.
There is #160 for tcmu_dev_err changes, I want to include it there.

seroyer: That looks better.  Thanks.

pkalever:
ping :-)

glfs.c
```diff
@@ -641,5 +682,12 @@ struct tcmur_handler glfs_handler = {
...
 int handler_init(void)
 {
+	int ret;
+
+	ret = pthread_mutex_init(&glfs_lock, NULL);
+	if (ret != 0) {
+		return -1;
+	}
+
 	return tcmur_register_handler(&glfs_handler);
```
mikechristie: Hey, sorry just noticed this. If tcmur_register_handler fails, we leak whatever pthread_mutex_init allocated. Need a pthread_mutex_destroy in the error path.

pkalever: @mikechristie this is done now. Thanks!

PR #172, opened 2017-06-09
fixup fd leakage in libtcmu_config.c

forsterzhang:
We open a PID filedescriptor, but never close it while a read() is interrrupted 
after reading some data to retrun -1.
So close the PID fd properly.

Signed-off-by: kai zhang <zhang.kai16@zte.com.cn>

### DIFF
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -371,12 +371,11 @@ int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 	}
 
 	len = tcmu_read_config(fd, buf, TCMU_MAX_CFG_FILE_SIZE);
+	close(fd);
 	if (len < 0) {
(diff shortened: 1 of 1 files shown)


libtcmu_config.c
```diff
@@ -373,6 +373,7 @@ int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 	len = tcmu_read_config(fd, buf, TCMU_MAX_CFG_FILE_SIZE);
```
mikechristie: We do not need the fd open after this line.

Can you move the close(fd) here, and then delete the one right after the closing } below.

forsterzhang:
@mikechristie  Thanks. Updated,, please review.

mikechristie:
Thanks!

PR #174, opened 2017-06-09
lock: fix return value check of pthread_*() calls

pkalever:
Summary of manual pages of all pthread_*() functions,
On success, pthread_*() returns 0; on error, they returns an error number


lxbsz:
Yes, the code will be clearer and easier to read as what the manunal page says.

mikechristie:
Thanks.

PR #176, opened 2017-06-21
config: Add dynamic reloading support

lxbsz:
Using the inotify, if write to the unwritable or crashed config file
forcely, the vi/vim will try to move and delete the old config file
and then recreate it again via the *.swp. When deleting the old
config file, the inotify watching handler will be removed from the
monitor too, so needs to add it back again.

### DIFF
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -18,7 +18,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/inotify.h>
 #include <sys/stat.h>
+#include <sys/types.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <unistd.h>
@@ -73,8 +75,8 @@
  *	}
  *
  * Note: For now, if the options have been changed in config file, the
- * tcmu-runner, consumer and tcmu-synthesizer daemons should be restarted.
- * And the dynamic reloading feature will be added later.
+ * system config reload thread daemon will try to update them for all the
+ * tcmu-runner, consumer and tcmu-synthesizer daemons.
  */
 
 static darray(struct tcmu_conf_option) tcmu_options = darray_new();
@@ -147,6 +149,7 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg)
 	/* set log_level option */
 	TCMU_PARSE_CFG_INT(cfg, log_level);
 	TCMU_CONF_CHECK_LOG_LEVEL(log_level);
+	tcmu_set_log_level(cfg->log_level);
 
 	/* add your new config options */
 }
@@ -164,18 +167,26 @@ struct tcmu_config *tcmu_config_new(void)
 	return cfg;
 }
 
-void tcmu_config_destroy(struct tcmu_config *cfg)
+void tcmu_cancel_config_thread(struct tcmu_config *cfg)
 {
-	struct tcmu_conf_option *opt;
+	pthread_t thread_id = cfg->thread_id;
+	void *join_retval;
+	int ret;
 
-	darray_foreach(opt, tcmu_options) {
-		if (opt->type == TCMU_OPT_STR)
-			free(opt->opt_str);
+	ret = pthread_cancel(thread_id);
+	if (ret) {
+		tcmu_err("pthread_cancel failed with value %d\n", ret);
+		return;
 	}
 
-	darray_free(tcmu_options);
+	pthread_join(thread_id, &join_retval);
+	if (ret) {
+		tcmu_err("pthread_join failed with value %d\n", ret);
+		return;
+	}
 
-	free(cfg);
+	if (join_retval != PTHREAD_CANCELED)
+		tcmu_err("unexpected join retval: %p\n", join_retval);
 }
 
 #define TCMU_MAX_CFG_FILE_SIZE (2 * 1024 * 1024)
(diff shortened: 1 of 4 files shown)


libtcmu_config.c
```diff
@@ -356,11 +375,64 @@ static void tcmu_parse_options(struct tcmu_config *cfg, char *buf, int len)
...
+{
+	struct tcmu_config *cfg = arg;
+	struct stat m_stat;
+	int ret;
+
+	while (1) {
+		sleep(10);
+
+		ret = stat(cfg->path, &m_stat);
+		if (ret == -1) {
```
mikechristie: I am not too familiar with it, but why can't we use inotify?

lxbsz: Yes, for linux the inotify is a better way to do this. 

Not sure will we adapt this to other unix os, such as Mac OS, in future?

mikechristie: Do not worry about it. There is plenty of linux specifics like netlink and uio.

lxbsz: Yeah.

lxbsz:
Updated.

main.c
```diff
@@ -787,11 +790,9 @@ int main(int argc, char **argv)
 	struct tcmulib_context *tcmulib_context;
 	darray(struct tcmulib_handler) handlers = darray_new();
 	struct tcmur_handler **tmp_r_handler;
-	struct tcmu_config *cfg;
 
-	cfg = tcmu_config_new();
-	tcmu_load_config(cfg, NULL);
-	tcmu_set_log_level(cfg->log_level);
+	tcmu_cfg = tcmu_config_new();
+	tcmu_load_config(tcmu_cfg, NULL);
```
mikechristie: Just a heads up. I merged PR

https://github.com/open-iscsi/tcmu-runner/pull/181

which adds a null check so it will break this part of the patch.

libtcmu_log.c
```diff
@@ -99,7 +99,7 @@ void tcmu_set_log_level(int level)
 {
 	/* set the default log level to warning */
```
mikechristie: You should also update the comment.

libtcmu_config.c
```diff
@@ -381,5 +506,12 @@ int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 
 	tcmu_parse_options(cfg, buf, len);
 
+	cfg->path = strdup(path);
```
mikechristie: Add a check for failure.

libtcmu_config.c
```diff
@@ -164,18 +167,36 @@ struct tcmu_config *tcmu_config_new(void)
...
+
+void tcmu_cancel_config_thread(struct tcmu_config *cfg)
+{
+	cancel_config_thread(cfg);
+}
+
+void tcmu_config_destroy(struct tcmu_config *cfg)
+{
+	cancel_config_thread(cfg);
 }
```
mikechristie: Let's just drop the wrappers. Do tcmu_cancel_config_thread and have it do all the cancel work.

Do we need a check for if pthread_create failed and we are not going dynamic mode?

libtcmu_config.c
```diff
@@ -356,11 +377,115 @@ static void tcmu_parse_options(struct tcmu_config *cfg, char *buf, int len)
...
+
+	darray_foreach(opt, tcmu_options) {
+		if (opt->type == TCMU_OPT_STR)
+			free(opt->opt_str);
+	}
+
+	darray_free(tcmu_options);
+	free(cfg->path);
+	free(cfg);
+}
```
mikechristie: You could probably put this in a function like destroy_config or config_cleanup and call it from here and the case where we are not doing dynamic config support.

libtcmu_log.c
lxbsz: yes

libtcmu_config.c
lxbsz: sure

libtcmu_config.c
lxbsz: Okay I will add it later

libtcmu_config.c
lxbsz: If the dynamic thread fail to start, should it fall back to static mode ？ or just exit the runner ？

mikechristie: It does not matter to me. Unless you have a specific use case that needs it one way or another, I would say do what is easier.

lxbsz:
Updated, please review.
@mikechristie

libtcmu_config.c
```diff
@@ -164,18 +167,31 @@ struct tcmu_config *tcmu_config_new(void)
...
+		tcmu_err("pthread_join failed with value %d\n", ret);
+		return;
+	}
 
-	free(cfg);
+	if (join_retval != PTHREAD_CANCELED)
+		tcmu_err("unexpected join retval: %p\n", join_retval);
+}
+
+void tcmu_cancel_config_thread(struct tcmu_config *cfg)
```
mikechristie: I am just going to rename cancel_config_thread to tcmu_cancel_config_thread and drop this wrapper since there does not seem to be much use.

libtcmu_config.c
mikechristie: It looks like you left support in for the non-dynamic case. For the cleanup of that case, were you going to add some code to cleanup the resources used?

The zte developers have been cleaning up those code paths, so I do not want to add a new one.

libtcmu_config.c
lxbsz: Yes, this has updated.

libtcmu_config.c
lxbsz: Sorry, Updated of this.

mikechristie:
Thanks.

PR #177, opened 2017-06-24
Perform netlink operations synchronously and add reconfig support

mikechristie:
These patches fix the issue where the kernel does not wait for userspace responses to add/delete/reconfigure, so if runner failed to execute them, the kernel will still end up updating its state. This can result in the kernel returning LUNs that are not setup in userspace, a race where close is called on the uio device after the kernel has done unregister, and the userspace and kernel device state out of sync.

These patches along with these kernel changes:
https://www.spinics.net/lists/target-devel/msg15706.html
fix the problem by having the kernel  wait for runner's response to these operations.

### DIFF
--- a/file_example.c
+++ b/file_example.c
@@ -41,6 +41,7 @@
 #include <scsi/scsi.h>
 
 #include "scsi_defs.h"
+#include "libtcmu.h"
 #include "tcmu-runner.h"
 
 struct file_state {
@@ -193,6 +194,18 @@ done:
 	return 0;
 }
 
+static int file_reconfig(struct tcmu_device *dev, struct tcmulib_cfg_info *cfg)
+{
+	switch (cfg->type) {
+	case TCMULIB_CFG_DEV_SIZE:
+		return tcmu_update_num_lbas(dev, cfg->data.dev_size);
+	case TCMULIB_CFG_DEV_CFGSTR:
+	case TCMULIB_CFG_WRITE_CACHE:
+	default:
+		return -EOPNOTSUPP;
+	}
+}
+
 static const char file_cfg_desc[] =
 	"The path to the file to use as a backstore.";
 
@@ -200,6 +213,7 @@ static struct tcmur_handler file_handler = {
 	.cfg_desc = file_cfg_desc,
 
 	.check_config = file_check_config,
+	.reconfig = file_reconfig,
 
 	.open = file_open,
 	.close = file_close,
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -48,6 +48,9 @@ static struct nla_policy tcmu_attr_policy[TCMU_ATTR_MAX+1] = {
 	[TCMU_ATTR_MINOR]	= { .type = NLA_U32 },
 	[TCMU_ATTR_CMD_STATUS]	= { .type = NLA_S32 },
 	[TCMU_ATTR_DEVICE_ID]	= { .type = NLA_U32 },
+	[TCMU_ATTR_DEV_CFG]	= { .type = NLA_STRING },
+	[TCMU_ATTR_DEV_SIZE]	= { .type = NLA_U64 },
+	[TCMU_ATTR_WRITECACHE]	= { .type = NLA_U8 },
 	[TCMU_ATTR_SUPP_KERN_CMD_REPLY] = { .type = NLA_U8 },
 };
 
@@ -69,6 +72,13 @@ static struct genl_cmd tcmu_cmds[] = {
 	{
 		.c_id		= TCMU_CMD_REMOVED_DEVICE,
 		.c_name		= "REMOVED DEVICE",
+		.c_msg_parser	= handle_netlink,
+		.c_maxattr	= TCMU_ATTR_MAX,
+		.c_attr_policy	= tcmu_attr_policy,
+	},
+	{
+		.c_id		= TCMU_CMD_RECONFIG_DEVICE,
+		.c_name		= "RECONFIG DEVICE",
 		.c_msg_parser	= handle_netlink,
 		.c_maxattr	= TCMU_ATTR_MAX,
 		.c_attr_policy	= tcmu_attr_policy,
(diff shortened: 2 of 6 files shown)


file_example.c
```diff
@@ -193,13 +194,26 @@ static int file_flush(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return 0;
 }
 
+static int file_reconfig(struct tcmu_device *dev, struct tcmulib_cfg_info *cfg)
+{
+	switch (cfg->type) {
+	case TCMULIB_CFG_DEV_SIZE:
+		return tcmu_update_num_lbas(dev, cfg->data.dev_size);
```
bgly: Dont you need to send a UA to notify capacity has changed?

diff --git a/libtcmu.c b/libtcmu.c
index aaecf27..712d949 100644
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -724,6 +724,8 @@ int tcmu_update_num_lbas(struct tcmu_device *dev, uint64_t new_size)
        if (!new_size)
                return -EINVAL;

+       dev->flags |= TCMU_CAPACITY_CHANGED;
+
        tcmu_set_dev_num_lbas(dev, new_size / tcmu_get_dev_block_size(dev));
        return 0;
 }
@@ -1004,3 +1006,17 @@ void tcmulib_cleanup_all_cmdproc_threads()
                cancel_thread(thread->thread_id);
        }
 }

Here you probably want to get more complex and check:
cdb[0] != INQUIRY && cdb[0] != GPCMD_GET_EVENT_STATUS_NOTIFICATION && REPORT LUNS(this one doesnt matter since tmcu doesnt process it), Request_sense? - This one we for some reason dont process but I believe you can copy the FBO emulate request sense and make it generic.

+
+int tcmulib_check_state(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
+{
+       int ret;
+
+       if(dev->flags & TCMU_CAPACITY_CHANGED) {
+               ret = tcmu_set_sense_data(cmd->sense_buf, UNIT_ATTENTION,
+                                         ASC_CAPACITY_DATA_HAS_CHANGED,
+                                         NULL);
+               dev->flags &= ~TCMU_CAPACITY_CHANGED;
+               return ret;
+       }
+       return 0;
+}
diff --git a/libtcmu.h b/libtcmu.h
index 5d041dc..4469b9d 100644
--- a/libtcmu.h
+++ b/libtcmu.h
@@ -32,6 +32,8 @@ extern "C" {
 #include "libtcmu_common.h"
 #include "libtcmu_log.h"

+#define TCMU_CAPACITY_CHANGED   0x01
+
 enum tcmulib_cfg_type {
        TCMULIB_CFG_DEV_CFGSTR,
        TCMULIB_CFG_DEV_SIZE,
@@ -142,6 +144,9 @@ void tcmulib_cleanup_cmdproc_thread(struct tcmu_device *dev);
 /* cleanup all (devices) command processing threads */
 void tcmulib_cleanup_all_cmdproc_threads();

+/* Checks state of the device to see if any reconfiguration happening */
+int tcmulib_check_state(struct tcmu_device *dev, struct tcmulib_cmd *cmd);
+
 #ifdef __cplusplus
 }
 #endif
diff --git a/libtcmu_priv.h b/libtcmu_priv.h
index d5d9078..6e9992e 100644
--- a/libtcmu_priv.h
+++ b/libtcmu_priv.h
@@ -72,6 +72,11 @@ struct tcmu_device {

        void *d_private; /* private ptr for the daemon */
        void *hm_private; /* private ptr for handler module */
+
+       /* Flag for state of the device for example if
+        * Capacity has changed, or path, etc.
+        */
+       uint32_t flags;
 };

 struct tcmu_thread {
diff --git a/main.c b/main.c
index ae80f3a..6f865a3 100644
--- a/main.c
+++ b/main.c
@@ -576,6 +576,10 @@ static void *tcmur_cmdproc_thread(void *arg)
                        if (tcmu_get_log_level() == TCMU_LOG_DEBUG_SCSI_CMD)
                                tcmu_cdb_debug_info(cmd);

+                       ret = tcmulib_check_state(dev, cmd);
+                       if(ret != 0)
+                               goto unit_attention;
+
                        if (tcmur_handler_is_passthrough_only(rhandler))
                                ret = tcmur_cmd_passthrough_handler(dev, cmd);
                        else
@@ -584,6 +588,7 @@ static void *tcmur_cmdproc_thread(void *arg)
                        if (ret == TCMU_NOT_HANDLED)
                                tcmu_warn("Command 0x%x not supported\n", cmd->cdb[0]);

+unit_attention:
                        /*
                         * command (processing) completion is called in the following
                         * scenarios:
diff --git a/scsi_defs.h b/scsi_defs.h
index ae2fb9f..42c2479 100644
--- a/scsi_defs.h
+++ b/scsi_defs.h
@@ -112,8 +112,10 @@ enum scsi_protocol {
 #define ASC_LBA_OUT_OF_RANGE                    0x2100
 #define ASC_INVALID_FIELD_IN_CDB                0x2400
 #define ASC_INVALID_FIELD_IN_PARAMETER_LIST     0x2600
+#define ASC_MEDIUM_CHANGED                     0x2800
 #define ASC_UNSUPPORTED_SEGMENT_DESC_TYPE_CODE  0x2609
 #define ASC_UNSUPPORTED_TARGET_DESC_TYPE_CODE   0x2607
+#define ASC_CAPACITY_DATA_HAS_CHANGED          0x2A09
 #define ASC_CANT_WRITE_INCOMPATIBLE_FORMAT      0x3005
 #define ASC_SAVING_PARAMETERS_NOT_SUPPORTED     0x3900
 #define ASC_INTERNAL_TARGET_FAILURE             0x4400


You will probably need something like this.

bgly: [0001-tcmu-Add-UA-to-reconfig.txt](https://github.com/open-iscsi/tcmu-runner/files/1103725/0001-tcmu-Add-UA-to-reconfig.txt)

bgly: You probably want to edit the check_state in the txt file to check for INQUIRY/GET_EVENT_STATUS_NOTIFICAITON/REPORT_LUNS(tcmu doesnt handle this atm but w/e), request sense....
Something like this:
        if(dev->flags & TCMU_CAPACITY_CHANGED &&
           cdb[0] != INQUIRY &&
           cdb[0] != GPCMD_GET_EVENT_STATUS_NOTIFICATION) {
                ret = tcmu_set_sense_data(cmd->sense_buf, UNIT_ATTENTION,
                                          ASC_CAPACITY_DATA_HAS_CHANGED,
                                          NULL);
                dev->flags &= ~TCMU_CAPACITY_CHANGED;
                return ret;
        }

mikechristie: Yeah, I was thinking when the kernel code is merged we could merge the 2 patches in this PR to add the basic infrastructure. Then do your UA patches in another PR on top.

mikechristie:
Kernel support posted here
http://www.spinics.net/lists/target-devel/msg15706.html
was merged in Nick's for-next branch. I think we missed the feature window for 4.13, so it should be in 4.14. There is compat code, so if using a kernel without the above patches then you just get the old behavior.

PR #178, opened 2017-06-26
tcmur-runner: increase max open file descriptors limit

pkalever:
In some environments default 'Max open files' will be 1024. As the number of
block create increases, open FDs also increases in daemon process quota and
in case if open fd limit is reached, then bock create will result in open() call
failures with EMFILE.

Solution is to set the 'Max open files' limit to higher value in tcmu-runner

### DIFF
--- a/main.c
+++ b/main.c
@@ -38,6 +38,8 @@
 #include <getopt.h>
 #include <poll.h>
 #include <scsi/scsi.h>
+#include <sys/time.h>
+#include <sys/resource.h>
 
 #include <libkmod.h>
 #include <sys/utsname.h>
@@ -52,6 +54,9 @@
 #include "version.h"
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
+
+#define TCMUR_MIN_OPEN_FD 65536
+#define TCMUR_MAX_OPEN_FD 1048576
 
 static char *handler_path = DEFAULT_HANDLER_PATH;
 /* tcmu log dir path */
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -53,6 +55,9 @@
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
 
+#define TCMUR_MIN_OPEN_FD 65536
+#define TCMUR_MAX_OPEN_FD 1048576
```
mikechristie: What about making this a tcmu.conf setting? Checkout the libtcmu_config.c code for examples on how to add new settings.

mikechristie:
How about just setting it in the tcmu-runner.service file. systemd will then just handle for us and users can override in that config file. We do not need to do anything.

Made a PR for that here:
https://github.com/open-iscsi/tcmu-runner/pull/246

mikechristie:
Closing.Merged #246.

lxbsz:
@mikechristie @pkalever 

I think this patch is needed and must.
I have encountered this limitation bug when running the runner in terminal with about more than 100 targets.

And this should be very useful.

main.c
lxbsz: @mikechristie @pkalever 
If this PR will be merged, since I wrote the configure code I can do this later based on it.

pkalever:
@lxbsz This can be solved in an other way with https://github.com/open-iscsi/tcmu-runner/pull/246, I think you might have missed this patch or not using systemd for starting your service.

Or Is that you mean for running tcmu-runner daemon without using systemd or non systemd env, we need this patch ?

lxbsz:
Yes, I know that PR works with systemd, but I need run it from the terminal.

mikechristie:
Just to make sure I get what you said, do you mean you have to do

# tcmu-runner

And cannot do

# systemctl start tcmu-runner

?

If so what is the use case?

If you need to do the non systemd start up then go ahead and fix up the patch and I will merge.

Just a warning, librbd seems to run out of FDs sometimes and t just hangs. You do not get a error. Just something to watch out for in case you are seeing the issue.

lxbsz:
These days we are debugging the runner in different environments with hundreds of backend devices , we'd like to see all the logs from the current shell terminal and to generate core dump when the runner crashed.
And for some the max open files is 1024.

Yes, I think I have encountered the similiar issue you mentioned above, but not very sure they are the same.

Thanks.

lxbsz:
@pkalever I will just pull your branch without changing your commit and then add new ones to do the fix and enhancemment.

Thanks

PR #181, opened 2017-07-03
fix: judge the function tcmu_config_new return value

tangwenji:
If the function tcmu_config_new return NULL，the process shoud exit，otherwise  it may cause coredump。


mikechristie:
Thanks!

PR #182, opened 2017-07-06
fix the syslog message ident

tangwenji:
tcmu-runner message in system log display as:
Jul  7 03:21:55 localhost ,o|: load_our_module:533 : Inserted module 'target_core_user'
The reason is that the first parameter 'ident' that calls openlog in the open_syslog function is a local variable. When calling syslog to print a specific message, the address pointed to by the 'ident' pointer may already be used by other programs, so there may be Garbled.
In the call openlog parameter should be a string constant or NULL (the system automatically add program name).


mikechristie:
Thanks.

PR #183, opened 2017-07-06
[RFC] runner: add local target_core_user.h

mikechristie:
It is kind of a pain to coordinate our upstream releases with
kernel headers releases. To allow users to be able to not have
to update kernel headers, but still use newer runner versions, this adds a
local target_core_user.h which will stay in sync with the upstream
kernel.

### DIFF
--- a/consumer.c
+++ b/consumer.c
@@ -36,7 +36,7 @@
 #include <stdint.h>
 #include <scsi/scsi.h>
 #define _BITS_UIO_H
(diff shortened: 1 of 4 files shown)


bgly:
works for me, we built a dkms version of LIO to get ours updated

mikechristie:
Merging this and if people have issues like we must use the kernel headers package then I will revert. Until then we can at least avoid user issues.

PR #184, opened 2017-07-07
fix:the dev->cfgstring is modifid in tcmu_rbd_open

tangwenji:
The dev->cfgstring is rbd/pool/image before call tcmu_rbd_open function,but it is modifid to rbd/pool after called,the cause is strtok function can modify the input paramter.


mikechristie:
Thanks.

PR #185, opened 2017-07-08
fix memory leak in gluster_cache_add

tangwenji:
When memory allocation failed, the allocated memory has to be freed

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -171,7 +171,7 @@ static int gluster_cache_add(gluster_server *dst, glfs_t *fs, char* cfgstring)
 
 	entry->server = calloc(1, sizeof(gluster_hostdef));
 	if (!entry->server)
-		goto error;
+		goto free_entry;
 
 	entry->server->type = dst->server->type;
 
@@ -192,6 +192,10 @@ static int gluster_cache_add(gluster_server *dst, glfs_t *fs, char* cfgstring)
 
 	return 0;
 
+free_entry:
+	if (entry->volname)
+		free(entry->volname);
+	free(entry);
  error:
 	return -1;
 }


mikechristie:
Looks good to me.

Will give gluster devs until Monday. @pkalever @pkalever please review.

glfs.c
```diff
@@ -192,6 +192,12 @@ static int gluster_cache_add(gluster_server *dst, glfs_t *fs, char* cfgstring)
 
 	return 0;
 
+free_entry:
+	if (entry->volname)
+		free(entry->volname);
+	if (entry->server)
+		free(entry->server);
```
pkalever: IMO, line 199 will never be executed, as the condition will fail always
Something similar like, 
```
diff --git a/glfs.c b/glfs.c
index 1a7fff3..fe3a525 100644
--- a/glfs.c
+++ b/glfs.c
@@ -170,8 +170,11 @@ static int gluster_cache_add(gluster_server *dst, glfs_t *fs, char* cfgstring)
        entry->volname = strdup(dst->volname);
 
        entry->server = calloc(1, sizeof(gluster_hostdef));
-       if (!entry->server)
+       if (!entry->server) {
+               free(entry->volname);
+               free(entry);
                goto error;
+       }
 
        entry->server->type = dst->server->type;
```
will be good.

@tangwenji Thanks for the catch :-)

tangwenji: updated，thanks

tangwenji:
updated，thanks

pkalever:
Looks good to me. thanks!

pkalever:
In File Optical handler https://github.com/open-iscsi/tcmu-runner/blob/master/file_optical.c#L431 
I have noticed 
```
        else {                                                                   
                   [...]
                                if (used_len + ret > alloc_len)                                                                                                                                                    
                                        break;    
                   [...]
        }                                                                        

```
is missing/not considered, in the *_emulate_mode_sense() function, but I'm not sure if it is intentional.

@bgly do you recollect why was this chunk not considered in the file optical handler ?

Thanks in advance :-)

PR #186, opened 2017-07-08
fix rados_connect failed in tcmu_rbd_image_open

tangwenji:
For clean up rados_shutdown()   is only necessary after rados_connect() has succeeded


mikechristie:
Nice. Thanks.

PR #189, opened 2017-07-10
Fbo media change

bgly:
To change dev_config run:

set attribute dev_size=661370880 dev_config=fbo//home/neo/ubuntu-16.10-server-ppc64el1.iso


file_optical.c
```diff
@@ -1670,13 +1670,60 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	int ret = -EOPNOTSUPP;
+
+	switch (cfg->type) {
+	case TCMULIB_CFG_DEV_CFGSTR:
+		ret = fbo_update_cfgstring(dev, cfg);
+		break;
+	case TCMULIB_CFG_DEV_SIZE:
+		tcmu_update_state(dev, TCMU_CAPACITY_CHANGE_FOR_CONFIG_CHANGE);
+		return 0;
+		break;
```
mikechristie: Do a break or return. Not both.

file_optical.c
```diff
@@ -1670,13 +1670,60 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	tcmu_remove_state(dev, TCMU_CAPACITY_CHANGE_FOR_CONFIG_CHANGE);	
+
+	tcmu_update_state(dev, TCMU_CONFIG_CHANGING);
+	
+	fbo_close(dev);
+
+	oldptr = cfg->data.dev_cfgstring;
+	tcmu_set_dev_cfgstring(dev, oldptr);
+
+	tcmu_err("path is %s\n", oldptr);
```
mikechristie: Don't forget to remove in the final patches.

main.c
```diff
@@ -576,6 +576,10 @@ static void *tcmur_cmdproc_thread(void *arg)
 			if (tcmu_get_log_level() == TCMU_LOG_DEBUG_SCSI_CMD)
 				tcmu_cdb_debug_info(cmd);
 
+			ret = tcmulib_check_state(dev, cmd);
+			if(ret)
```
mikechristie: Missing space between if and (.

libtcmu.h
```diff
@@ -41,7 +46,11 @@ enum tcmulib_cfg_type {
 struct tcmulib_cfg_info {
 	enum tcmulib_cfg_type type;
 
+	struct tcmulib_context *ctx;
+	struct genl_info *info;
```
mikechristie: I didn't see these get used in the patches.

libtcmu.h
```diff
@@ -41,7 +46,11 @@ enum tcmulib_cfg_type {
...
+
 	union {
+		char *dev_name;
```
mikechristie: Same as above.

libtcmu.h
```diff
@@ -32,6 +32,11 @@ extern "C" {
 #include "libtcmu_common.h"
 #include "libtcmu_log.h"
 
+#define TCMU_CAPACITY_CHANGED   0x01
+#define TCMU_CONFIG_CHANGING	0x02
+#define TCMU_CONFIG_CHANGED	0x04
+#define TCMU_CAPACITY_CHANGE_FOR_CONFIG_CHANGE	0x08
```
mikechristie: Let's rename to make it more clear what they are for. Maybe something like TCMU_DEV_STATE_something_sommething

file_optical.c
```diff
@@ -1670,13 +1670,60 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return 0;
 }
 
+static int fbo_update_cfgstring(struct tcmu_device *dev, struct tcmulib_cfg_info *cfg)
+{
+	char *oldptr;
+
+	if (!tcmu_check_state(dev, TCMU_CAPACITY_CHANGE_FOR_CONFIG_CHANGE)) {
+		tcmu_err("You need to include size change to be able to change dev_config\n");
```
mikechristie: tcmu_dev_err

file_optical.c
```diff
@@ -1670,13 +1670,60 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	tcmu_update_state(dev, TCMU_CONFIG_CHANGING);
+	
+	fbo_close(dev);
+
+	oldptr = cfg->data.dev_cfgstring;
+	tcmu_set_dev_cfgstring(dev, oldptr);
+
+	tcmu_err("path is %s\n", oldptr);
+
+	fbo_open(dev);
```
mikechristie: Why not return an error if this fails?

Maybe at least an error message if this fails.

bgly:
@mikechristie So far this testing has been approved by our internal IBMi OS team in terms of our use case and how we want to use it, let me know if you have any other questions or comments.

file_optical.c
```diff
@@ -1670,13 +1670,68 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return 0;
 }
 
+static int fbo_update_cfgstring(struct tcmu_device *dev,
+				struct tcmulib_cfg_info *cfg)
+{
+	char *oldptr;
+	int ret;
+
+	if (!tcmu_check_state(dev, TCMU_DEV_STATE_CAPACITY_CHANGE_FOR_CONFIG_CHANGE)) {
```
mikechristie: I did not get why you need to change the size and cfg string?

Is it just to enforce that if the size did change it would always need to come before the cfg string part part of the update?

file_optical.c
```diff
@@ -1670,13 +1670,68 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	oldptr = cfg->data.dev_cfgstring;
+	tcmu_set_dev_cfgstring(dev, oldptr);
+
+	ret = fbo_open(dev);
+	if (ret != 0) {
+		tcmu_dev_err(dev, "Failed opening of new media\n");
+		goto fail_open;
+	}
+
+	tcmu_remove_state(dev, TCMU_DEV_STATE_CONFIG_CHANGING);
```
mikechristie: Between these 2 state changes is there a race? It seems like a command could be executed at the time and would not get the UA right?

file_optical.c
bgly: Yes, it is to enforce that if you swap the media and if it is different then the size is absorbed first. At first I didn't change the size and on media changes where the sizes change it causes a reset and the client isn't happy on either Linux or AIX.

file_optical.c
bgly: I mean it is possible, so a possible fix would be to move the update state so that is before the remove state. Then if a cmd does come in then the check for TCMU_DEV_STATE_CONFIG_CHANGED would be executed.

file_optical.c
```diff
@@ -1670,13 +1670,68 @@ static int fbo_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	tcmu_set_dev_cfgstring(dev, oldptr);
+
+	ret = fbo_open(dev);
+	if (ret != 0) {
+		tcmu_dev_err(dev, "Failed opening of new media\n");
+		goto fail_open;
+	}
+
+	tcmu_update_state(dev, TCMU_DEV_STATE_CONFIG_CHANGED);
+	tcmu_remove_state(dev, TCMU_DEV_STATE_CONFIG_CHANGING);
```
mikechristie: I think you will hit a similar problem as before. But now, if you get a command after the CONFIG_CHANGED change but before the CONFIG_CHANGING you will get an extra medium not present UA.

Maybe we want to do the state changes under a lock or make the state an atomic.

bgly: I went ahead and did the state changes under a lock

mikechristie:
I have not forgot this patchset.

I have some patches to make reopening a device safer. Currently, for some handlers it might cause problems if commands are in flight when we ere changing internal structs like changing the fd being used. I am going to adapt your patches to mine. Will cc you on the PR.

bgly:
Okay thanks! I figured you were reworking something

mikechristie:
I am making a 1.3.0-rc1 build tonight and I was not able to get this in.

I am still hoping to get it into the final. I have added __tcmu_reopen_dev which can be used to reopen a device with the new settings. I am just fixing up the UA code. I am going to make it work under the state_lock added.

I am just trying to make the update nicer in that it is kind of klunky to have to do multiple update calls. I am debating about just passing all the needed info in the cfg string then libtcmu parsing out values like the size and then passing the handler just the handler info like the file path. I am still testing some things though.

bgly:
Yeah, it is kind of klunky... But I had to do it incase IO comes in during different states of the reconfiguration... 

I guess if you can make size and string handled in one go then you wouldn't need to have the two states for a media change.

mikechristie:
Added UA support here:

https://github.com/open-iscsi/tcmu-runner/pull/309

Going forward for the multiple settings (size and dev_config string) type of operation we need for fbo backing change, I am thinking maybe we just want to pass all the new values in the dev_config or add a new configfs file that can be used to update multiple settings.

I think we want to add a new file for this, so dev_config can be used for that specific string and there will be no compat issues.

bgly:
Hi Mike, 

I have some free time to get this patch updated, since your last comment do you have a specific way you want this implemented?

bgly:
@mikechristie

mikechristie:
Sorry for the late reply. I have been debating with myself on this the last couple of days.

I think just adding a new configfs attr that is for updating multiple options will be best. It can be a string of multiple options similar to what is used for the initial setup.

What do you think?

bgly:
@mikechristie I think that is probably best so that a media change can be processed all at once and not one option at a time

bgly:
@mikechristie How do you want the new attribute to be delimited? I was thinking a ":" ? 

set attribute dev_reconfig=dev_config:blahblah:dev_size=2222:etc?

mikechristie:
":" is probably best looking but some tools/libs like rtslib does not allow ":" in some inputs.

We have been using ";".

PR #191, opened 2017-07-13
libtcmu: fix unregister genl ops in setup_netlink

tangwenji:
After genl_register_family succeeds，the following operations failed should call genl_unregister_family。


mikechristie:
Thanks. Hopefully, you are finding all these issues by just review and not actually hitting all these bugs :)

PR #192, opened 2017-07-13
tcmu-runner: fix the resource released when exception exits in main

tangwenji:
The cfg should released when main function abnormal exit；
also the tcmulib_context should closed after tcmulib_initialize successed whether main function' is normal or abnormal  exit。


mikechristie:
Thanks.

PR #193, opened 2017-07-14
Various security and functionality related bugfixes (multiple DoS, memory leaks)

mgerstner:
While doing a security audit of tcmu-runner I've come across these issues.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -544,7 +544,7 @@ static void remove_device(struct tcmulib_context *ctx,
 static int is_uio(const struct dirent *dirent)
 {
 	int fd;
-	char tmp_path[64];
+	char tmp_path[NAME_MAX];
 	char buf[256];
 	ssize_t ret;
 
@@ -587,7 +587,7 @@ static int open_devices(struct tcmulib_context *ctx)
 		return -1;
 
 	for (i = 0; i < num_devs; i++) {
-		char tmp_path[64];
+		char tmp_path[NAME_MAX];
 		char buf[256];
 		int fd;
 		int ret;
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -358,7 +358,8 @@ static void tcmu_parse_options(struct tcmu_config *cfg, char *buf, int len)
 
 int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 {
-	char buf[TCMU_MAX_CFG_FILE_SIZE];
+	char *buf = malloc(TCMU_MAX_CFG_FILE_SIZE);
+	int res = 0;
 	int fd, len;
 
 	if (!path)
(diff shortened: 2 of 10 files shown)


mikechristie:
One general comment for all these patches.

For coding style and ifs, we try to follow the linux kernel coding style. There is some code that does not follow it, but your style is way too different :)

1. Put a space between the "if" and opening "(".
2. Do not put a space between the opening "(" and the first test.
3. Put the opening "{" on the same line as the "if" if possible.

Something like:

`
if (x == y) {
`

4. For multiple line tests don't start it on a new line so:

```
if(
        g_unix_signal_add(SIGINT, sighandler, loop) <= 0 ||
        g_unix_signal_add(SIGTERM, sighandler, loop) <= 0
) {

```

would be

```

if (g_unix_signal_add(SIGINT, sighandler, loop) <= 0 ||
    g_unix_signal_add(SIGTERM, sighandler, loop) <= 0) {

```

mgerstner:
Okay I tried to accommodate your requests. Hope this helps.

libtcmu_config.c
```diff
@@ -358,7 +358,8 @@ static void tcmu_parse_options(struct tcmu_config *cfg, char *buf, int len)
 
 int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 {
-	char buf[TCMU_MAX_CFG_FILE_SIZE];
+	char *buf = malloc(TCMU_MAX_CFG_FILE_SIZE);
```
mikechristie: This change is ok with me. What do you recommend for the limit when to switch to dynamic allocation should be?

mgerstner: Given that the currently shipped config file is 286 bytes large I think a maximum size of 64KiB or something would suffice. As long as it stays on the heap the 2MiB are also okay for the moment. Just when it's on the stack strange things can happen like with valgrind failing to understand what's going on.

In the end the best solution would be to process the config file line wise using a buffer of LINE_MAX bytes. But that change would have been out of the scope of my security oriented review / pull request.

pkalever:
No doubt the patches are really helpful, over all looks good to me.

But my highest concern is that the manageability of the patches, we better squash them at required levels. This helps in reviewing better, easy back portability and easy to tracking as well.

For example:
patch "silenced warnings introduced by memory leak fix"  is a fix for "fixed a number of memory leaks with (de)registering of dbus handlers"

patch "adjusted coding style to upstream requirements" fixes the code style violation in patches introduced within the pull request, so can be distributed across.

patch "use boolean in struct tcmur_handler for keeping track of external handlers"  ->  "only allow dynamic UnregisterHandler for external handlers, thereby fixing DoS"

HTH

libtcmu_config.c
```diff
@@ -367,19 +368,23 @@ int tcmu_load_config(struct tcmu_config *cfg, const char *path)
 	fd = open(path, O_RDONLY);
 	if (fd < 0) {
 		tcmu_err("Failed to open file '%s'\n", path);
-		return -1;
+		res = -1;
```
pkalever: This is just a matter of coding style, but If I write this, I would have initialized 'res=-1;' and at the end, before label 'out:' I would have assigned it to 'res=0;'

But as of today it looks okay, in the future if someone introduces more check in this function has to duplicate "res=-1" in every { }

libtcmu.c
```diff
@@ -544,7 +544,7 @@ static void remove_device(struct tcmulib_context *ctx,
 static int is_uio(const struct dirent *dirent)
 {
 	int fd;
-	char tmp_path[64];
+	char tmp_path[NAME_MAX];
```
mgerstner: Yes, you're right. I was so focused on d_name, which is of length NAME_MAX, because the gcc warning was about that. It should be PATH_MAX. I will adjust.

libtcmu_config.c
mgerstner: Agreed. I will adjust.

mgerstner:
I've opened up a new pull request from a separate branch in my fork repo, because of all the rebasing I did. Avoids more confusion I guess. Closing this one.

New pull request is here:

https://github.com/open-iscsi/tcmu-runner/pull/200

PR #199, opened 2017-07-18
build: fix compiler warnings

pkalever:
fixed warnings: uninitialized format-truncation unused-results

Signed-off-by: Prasanna Kumar Kalever <prasanna.kalever@redhat.com>

Issue #194, opened 2017-07-14
Security information leak in internal qcow handler's check_config callback


mgerstner:
The implementation of `qcow_check_config` presents an information leak, because it allows any user with access to the dbus system bus to check for the existance of files and directories, even if the user shouldn't have access to those files.

For example issuing this command as a regular user allows to get the knowledge, whether */root/.bash_history* exists:

    dbus-send --system --print-reply --dest=org.kernel.TCMUService1 /org/kernel/TCMUService1/qcow org.kernel.TCMUService1.CheckConfig string://root/.bash_history

These kind of information leaks in combination with other security issues often leverage more serious security issues.

I'd have suggested a bugfix for this but it is difficult to get right. If we knew the peer credentials from dbus we could do a comparison of those credentials against the target file credentials. This still would be difficult for corner cases, where the file itself would be accessible to the user, but not some of the upper components of the path.

The `check_config` callback does currently not provide a possibility to pass peer credential information in, so this would affect public API in some way. Also I've not found a simple way to get the dbus peer credentials via libgio, maybe by using a `GDBusAuthObserver`.

All quite complicated, might be easier to drop this config callback altogether.

mikechristie:
We are actually in the process of removing the check_config callback, so just send a patch to remove them all (file_example.c will have a similar issue I think).

mgerstner:
Okay I did as you suggested. Good hint with the *file_example.c*. It contains an even worse vulnerability, allows to temporarily create arbitrary files in the system, possibly deleting other processes' files after winning a race.

mikechristie:
Patches for these were merged in PR above, so closing issue.

PR #200, opened 2017-07-18
Rebased: Various security and functionality related bugfixes (multiple DoS, memory leaks)

mgerstner:
While doing a security audit of tcmu-runner I've come across these issues.

Rebased and squashed version of previous pull request:

https://github.com/open-iscsi/tcmu-runner/pull/193

### DIFF
--- a/main.c
+++ b/main.c
@@ -935,6 +935,7 @@ int main(int argc, char **argv)
 	g_main_loop_unref(loop);
 	tcmulib_close(tcmulib_context);
 	tcmu_config_destroy(tcmu_cfg);
+	darray_free(handlers);
 
 	return 0;
 
--- a/main.c
+++ b/main.c
@@ -91,6 +91,12 @@ int tcmur_register_handler(struct tcmur_handler *handler)
 	return 0;
 }
 
+static int tcmur_register_dbus_handler(struct tcmur_handler *handler)
+{
+	assert(handler->_is_dbus_handler == true);
+	return tcmur_register_handler(handler);
+}
+
 bool tcmur_unregister_handler(struct tcmur_handler *handler)
 {
 	int i;
@@ -101,6 +107,16 @@ bool tcmur_unregister_handler(struct tcmur_handler *handler)
 		}
 	}
 	return false;
+}
+
+static bool tcmur_unregister_dbus_handler(struct tcmur_handler *handler)
+{
+	bool ret = false;
+	assert(handler->_is_dbus_handler == true);
+
+	ret = tcmur_unregister_handler(handler);
+
+	return ret;
 }
 
 static int is_handler(const struct dirent *dirent)
@@ -315,7 +331,7 @@ on_handler_appeared(GDBusConnection *connection,
 
 	if (info->register_invocation) {
 		info->connection = connection;
-		tcmur_register_handler(handler);
+		tcmur_register_dbus_handler(handler);
 		dbus_export_handler(handler, G_CALLBACK(on_dbus_check_config));
 		g_dbus_method_invocation_return_value(info->register_invocation,
 			    g_variant_new("(bs)", TRUE, "succeeded"));
@@ -340,7 +356,7 @@ on_handler_vanished(GDBusConnection *connection,
 			    g_variant_new("(bs)", FALSE, reason));
 		g_free(reason);
 	}
-	tcmur_unregister_handler(handler);
+	tcmur_unregister_dbus_handler(handler);
 	dbus_unexport_handler(handler);
 }
 
@@ -366,6 +382,8 @@ on_register_handler(TCMUService1HandlerManager1 *interface,
 	handler->handle_cmd   = dbus_handler_handle_cmd;
 
 	info = g_new0(struct dbus_info, 1);
+	handler->opaque = info;
+	handler->_is_dbus_handler = 1;
 	info->register_invocation = invocation;
 	info->watcher_id = g_bus_watch_name(G_BUS_TYPE_SYSTEM,
 					    bus_name,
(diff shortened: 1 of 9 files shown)


pkalever:
Hi @mgerstner ,
I was discussing the criticality of this issue with @sidhax and we are wondering if you have raised any CVE for the security issue for this ?

mgerstner:
Actually that was something I wanted to ask you about, too. I have not requested any CVE yet. 

I think commit 8cf8208775022301adaa59c240bb7f93742d1329 justifies a CVE, because of the information leak as described in issue #194.

The memory leaks and denial of service issues from commits

7a78eda52d973d3edc06fea84ad874678d6055f0
bb80e9c7a798f035768260ebdadffb6eb0786178
e2d953050766ac538615a811c64b34358614edce

are minor issues, but mostly they're also assigned CVEs for easier tracking by other distros.

Do you have a workflow for this? I can request the CVEs for you if you like.

sidhax:
@mgerstner can you please file request for CVEs ?

mgerstner:
@sidhax: Okay I will request CVEs and report back when they're assigned.

No you can't read file content but it still gives you some information that would otherwise not be available like what files are existing in /root. There could be situations, combined with other security issues in other software, when this can lead to exploits.

Worse is the example code in file_example_c, by the way. It allows to create arbitrary empty files for a short time, making it possible for example to create /etc/suid-debug just in time for some suid binary to run in debug mode. There could be other possibilities like creating things via /sys. I hope nobody used this example code for their own handlers.

The DoS are only local yes. But a non-root user can DoS a service running as root. Or leak memory.

Nothing critical altogether but let's give users of tcmu-runner a chance to recognize these issues and backport them if they deem it necessary.

pkalever:
@mgerstner Make sense to me, Thanks :-)

pkalever:
@mgerstner sorry! you may have to drop https://github.com/open-iscsi/tcmu-runner/pull/200/commits/b0beac3a490092c7c202dc4fb0bbaa7af09bb10d as https://github.com/open-iscsi/tcmu-runner/pull/199 is closed

mgerstner:
Okay I've rebased, dropped the conflicting commit

mikechristie:
> @mikechristie
Should we wait till open-iscsi/targetcli-fb#87 gets merged and available ?

I don't know.

I have been debating that. It will also require users to upgrade to the 4.13 kernel and that is around rc1 stage right now.

The thing is that with the removal of the check_config code, I think we can just delete all the dbus code since it does nothing now as far as I can tell. For this we would also need to sync up with targetcli releases. It's code to connect to dbus will fail during device setup, so existing setups would fail if they updated runner, but not targetcli.

Maybe, we could say for the next runner release you must use 4.13 and targetcli-fb XYZ?

main.c
```diff
@@ -935,6 +935,7 @@ int main(int argc, char **argv)
 	g_main_loop_unref(loop);
 	tcmulib_close(tcmulib_context);
 	tcmu_config_destroy(tcmu_cfg);
+	darray_free(handlers);
```
mikechristie: This was just fixed in commit, so will drop this patch.

commit aaf121b7441a939b1f60fcddeacad7c69f0c2b1e
Author: Xiubo Li <lixiubo@cmss.chinamobile.com>
Date:   Thu Jul 20 10:11:41 2017 +0800

    tcmur: free useless darray type handlers memory.

mikechristie:
Here is one additional fix up we will need for these patches. When we would exit() from the signal handler, the code after the g_main_run would not get run normally. With the patch to use g_main_quit, we cannot use the logger after the signal handler has killed the logger threads. It also moves the cancel_log_thread call to after the config cancel incase the config shutdown code tries to log something.

[0001-runner-fix-log-use-during-shutdown.txt](https://github.com/open-iscsi/tcmu-runner/files/1163654/0001-runner-fix-log-use-during-shutdown.txt)

mikechristie:
One update:

I have emailed the targetcli maintainer about trying to coordinate check_config and dbus removal. When that is done I will merge these patches.

mgerstner:
@mikechristie: do you want me to add the logging patch to this pull request? I can add an additional commit but want to avoid another rebase, because I've already used the commit URLs in CVE requests and bugzilla tickets.

mgerstner:
@mikechristie: we could also think about cancelling the log thread not in the signal handler but at the end of the main() function. This would make for a smaller patch and logging could be used without worries about shutdown situations.

mikechristie:
No, you do not need to carry the patch. I will merge it when I merge you PR.

As far as doing the canceling in the sig handler or end of the main function, it does not matter to me where it goes. Is there any place where it is more common? It would be best to do both the config and log cancel calls in the same place, just so it is consistent.

mikechristie:
Here's the plan.

1. I am going to merge this set as is even though some patches will not be necessary when the dbus code is removed.
2. Andy (targetcli/rtslib -fb maintainer) will help us coordinate a release.
3. In another PR we will remove the dbus code.

mgerstner:
FTR:

- CVE requests are still pending
- I've posted a summary on the oss-security mailing list: http://www.openwall.com/lists/oss-security/2017/07/24/1

mgerstner:
I've got most of the CVEs via [DWF](https://github.com/distributedweaknessfiling/) now:

- CVE-2017-1000198 for DoS in the GLFS handler which was removed in commit 61bd03e600d2abf309173e9186f4d465bb1b7157
- CVE-2017-1000190 for information leak in qcow handler as reported in issue #194
- CVE-2017-1000200 for DoS fixed in commit bb80e9c7a798f035768260ebdadffb6eb0786178
- CVE-2017-1000201 for DoS fixed in commit e2d953050766ac538615a811c64b34358614edce

a CVE for the memory leaks fixed in commit 7a78eda52d973d3edc06fea84ad874678d6055f0 is still pending.

PR #201, opened 2017-07-20
tcmur: free useless darray type handlers memory.

lxbsz:


mikechristie:
Thanks.

PR #202, opened 2017-07-20
Libtcmu library update

lxbsz:
1, Move alua/tcmur_aio/tcmur_cmd-handler to libtcmu library.
2, Rename tcmur_aio to libtcmu_aio
3, Rename tcmur_cmd_handler to libtcmu_cmd_handler
4, Update some helpers name to make the code tidy and more readable.

These are all preparing for adding .handle_cmd support for handlers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ add_library(tcmu
   api.c
   alua.c
   tcmur_cmd_handler.c
-  tcmur_aio.c
+  libtcmu_aio.c
   libtcmu.c
   libtcmu-register.c
   tcmuhandler-generated.c
@@ -70,7 +70,7 @@ add_library(tcmu_static
   api.c
   alua.c
   tcmur_cmd_handler.c
-  tcmur_aio.c
+  libtcmu_aio.c
   libtcmu.c
   libtcmu-register.c
   tcmuhandler-generated.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ add_library(tcmu
   configfs.c
   api.c
   alua.c
-  tcmur_cmd_handler.c
+  libtcmu_cmd_handler.c
   libtcmu_aio.c
   libtcmu.c
   libtcmu-register.c
@@ -69,7 +69,7 @@ add_library(tcmu_static
   configfs.c
   api.c
   alua.c
-  tcmur_cmd_handler.c
+  libtcmu_cmd_handler.c
   libtcmu_aio.c
   libtcmu.c
   libtcmu-register.c
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,9 @@ add_library(tcmu
   SHARED
   configfs.c
   api.c
+  alua.c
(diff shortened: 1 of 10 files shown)


mikechristie:
Are you going to make the libtcmu_aio.c and libtcmu_cmd_handler.c code be usable from non tcmu-runner daemons or just kill the abstraction?

The libtcmu code is supposed to be code that is common to all daemon implementations. The aio and cmd_handler code is currently tied to tcmu-runner because it is tied to the tcmu runner handler and tcmu runner device.

If you are going to just kill the libctmu outside daemon support, I think I am ok with this. So I mean merging the libtcmu and runner handlers and the runner device and libtcmu device.

I am worried about breaking libtcmu users, because I cannot contact all of them. Making them recompile because we change a struct or function signiture is not too bad. I have been trying to not change behavior of existing functions in a way that might break someone without them knowing.

I think I would be ok with just saying in the release we are working on outside daemons are just not supported anymore.  Thoughts?

lxbsz:
For adding handle_cmd support in handlers, as we discussed before, the generic scsi processing work should be in some where like tcmur_cmd_handler.c as traditional. So for all handlers they need a way to access the callout.

There are two ways to do this:
1, as this patch series does, and then adds the callouts in tcmur_cmd_handler.c.
2, as tcmur_register_handler() does.

Which one is better ?

IMO the current one is better, but this will introduce the breaking libtcmu users issue mentioned above. Since the runner is still in developing process(?), so I agree declaring about this through release note.

mikechristie:
I am ok with breaking libtcmu and saying the last version released by Andy was the last version to support that API.

However, another option is that we can also just add a new callout for the passthrough you are adding.

- handle_cmd will only be used for fbo/file_optical type of handlers where they handle all SCSI details. It can also be used for the libtcmu type of handlers that did their own daemon.

We let them do whatever they want and only provide basic support like running it in a thread for them if they need it like we do today for file_optical.

- New handler, something_something_cmd(). It can be whatever you want. It will give you a clean slate and we handlers like file_optical can still have a the level of flexibility they wanted and we do not have to worry about them.

lxbsz:
If my understanding is correct, the .handle_cmd will only be used for passthrough only type handlers like file_optical.c.

Adding a new callout .something_something_cmd to handle the current cmds we are discussing?

Or sorry for not correctly getting the New handler idea.

lxbsz:
@mikechristie 
Please see the PR #151's newest update. 
It moved all the generic scsi and aio code back to tcmur_cmd_handler and temporarily export the tcmur_handle_unmap() from tcmur_cmd_handler.c through main-syms.txt.

What the rbd handler should do is to provide its own unmap_fn hook to tcmur_cmd_handler.c.

mikechristie:
Hey,

You do not need this patch anymore right?

It looks like your current unmap patches do not use this.

I also posted a patch to begin removing the stable libtcmu API. We can then kill all that funky code.

mikechristie:
Closing. Just reopen if we need it again.

lxbsz:
@mikechristie
Sorry for missing this.
Many times I couldn't receive any replied e-mails like these ones.

PR #204, opened 2017-07-20
tcmu: fix SCCS bit of standard inquiry response

tangwenji:
Set SCCS for RTPG and STPG 

### DIFF
--- a/api.c
+++ b/api.c
@@ -293,6 +293,15 @@ int tcmu_emulate_std_inquiry(
 	 * Enable the XCOPY
 	 */
 	buf[5] = 0x08;
+	/*
+	  * Set SCCS for RTPG and STPG.spc4r22 6.27 6.35
+	  * The rtpg command is a service action of the MAINENANCE IN command,
+	  * and the stpg command is a service action of the MAINENANCE OUT command.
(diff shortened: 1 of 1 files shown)


api.c
```diff
@@ -293,6 +293,15 @@ int tcmu_emulate_std_inquiry(
 	 * Enable the XCOPY
 	 */
 	buf[5] = 0x08;
+	/*
+	  * Set SCCS for RTPG and STPG.spc4r22 6.27 6.35
+	  * The rtpg command is a service action of the MAINENANCE IN command,
+	  * and the stpg command is a service action of the MAINENANCE OUT command.
```
mikechristie: Agree with the above chunk.

api.c
```diff
@@ -293,6 +293,15 @@ int tcmu_emulate_std_inquiry(
...
+	  * The MAINTENANCE IN and MAINTENANCE OUT service actions defined
+	  * only in SCC-2 apply only to logical units that return a device type of 0Ch
+	  * or the SCCS bit set to one in their standard INQUIRY data.
+	  */
```
mikechristie: I am not sure how/why this is related to RTPG/STPG. RTPG and STPG are defined in SPC and I do not see them defined or even redefined in SCC. I only have access to SCC 2 r4 though.

PR #206, opened 2017-07-21
tcmu: fix SCCS bit of standard inquiry response

tangwenji:
Set SCCS for RTPG and STPG

### DIFF
--- a/api.c
+++ b/api.c
@@ -293,6 +293,14 @@ int tcmu_emulate_std_inquiry(
 	 * Enable the XCOPY
 	 */
 	buf[5] = 0x08;
+	/*
+	  * Set SCCS for RTPG and STPG.spc4r22 6.27 6.35
+	  * The rtpg command is a service action of the MAINENANCE IN command,
+	  * and the stpg command is a service action of the MAINENANCE OUT command.
+	  * The MAINTENANCE IN + RTPG  and MAINTENANCE OUT + STPG need
+	  * the SCCS bit set to one in their standard INQUIRY data.
+	  */
+	buf[5] |= 0x80;
 	if (port)
 		buf[5] |= port->grp->tpgs;
 


tangwenji:
@mikechristie 
Sorry， due to operational errors， the pull request #204  is closed by me。
I modified the comment and update。

mikechristie:
My issue with the patch is not the comment. I am saying the patch and comment are incorrect.

The is what I see in spc4r37:

> If the PERIPHERAL DEVICE TYPE field is set to 0Ch (i.e., storage array controller device) or the SCCS bit is set to one in the standard INQUIRY data (see 6.6.2), then the MAINTENANCE IN service action definition in SCC-2 for the specified service action, if any, applies.

There is something similar for MAINTENANCE OUT.

That does not mean that you need to set the SCCS bit for RTPG/STPG commands. It us just saying that if the dev type of 0Ch or SCCS is set, then you use the definitions in SCC-2. In SCC-2, I do not see any definition for RPTG or STPG, so as far as I can tell we use the SPC definitions and those do not require SCCS to be set for a disk device type.

If you see something else in a older version of spc4 could you quote it here? I do not have the version you used.

mikechristie:
Oh yeah, for SCCS, I only have scc 2 r4. If you have a version where it defines RTPG/STPG please quote it here for me. The company I work for is not a t10 member, so I do not have access to all versions of specs.

tangwenji:
In SPC4r22:6.27
> The REPORT TARGET PORT GROUPS command is a service action of the MAINTENANCE IN command.
> Additional MAINTENANCE IN service actions are defined in SCC-2 and in this standard. The MAINTENANCE IN service actions defined only in SCC-2 apply only to logical units that return a device type of 0Ch or the SCCS bit set to one in their standard INQUIRY data.

tangwenji:
in spc4r22 defined MAINTENANCE OUT service actions not in scc2：
stpg 0Ah
change aliases 0Bh
set priority 0Eh
set timestamp 0Fh

in spc4r22 defined MAINTENANCE IN service actions not in scc2
rtpg 0Ah
report aliases 0Bh
report supported operation codes  0Ch
report supported task management function 0Dh
report priority 0Eh 
report timestamp 0Fh

maybe our scc2 version is too old。

mikechristie:
> maybe our scc2 version is too old。

I am not sure what you mean. In my copy I said RTPG and STPG are not defined in my version of SCC2.

I think we are interpreting the spec differently.

From your quote:

> 
> In SPC4r22:6.27
> 
> The REPORT TARGET PORT GROUPS command is a service action of the MAINTENANCE IN command.
> Additional MAINTENANCE IN service actions are defined in SCC-2 and in this standard. The MAINTENANCE IN service actions defined only in SCC-2 apply only to logical units that return a device type of 0Ch or the SCCS bit set to one in their standard INQUIRY data.

As you you wrote, and we agree on, STPG and RTPG are not in SCC. I am saying because of that, we do not need to set the SCCS bit. The part you quoted says the SCCS bit just means that if the device is not type 0Ch, the device can still implement the MAINTENANCE SAs  in SCC. STPG and RTPG are not in SCC, so we do not need to set the SCCS bit.

As a disk type device, if we set the SCCS bit it means we implement the MAITENANCE SAs defined in SSC. RTPG and STPG are not in there, and we are not implementing any of the SAs in SSC.

mikechristie:
Hey,

If you disagree with my interpretation of the spec, then just send a patch to the upstream target/lio inquiry code spc_emulate_inquiry_std, or to a project like the scst's vdisk code to do add the same behavior. If they merge the patch, then I will merge the patch in this PR.

tangwenji:
I think the device type is 0Ch，we implement the MAINTENANCE SAs in SCC，as a disk type device，we implement some MAINTENANCE SAs（such as RTPG、STPG etc），standard inquiry data SCCS bit must set to one。

tangwenji:
In kernel tcm code，
```
spc_emulate_inquiry_std(struct se_cmd *cmd, unsigned char *buf)
{
	********
	/*
	 * Enable SCCS and TPGS fields for Emulated ALUA
	 */
	spc_fill_alua_data(lun, buf);
        *******
}
static void spc_fill_alua_data(struct se_lun *lun, unsigned char *buf)
{
	struct t10_alua_tg_pt_gp *tg_pt_gp;

	/*
	 * Set SCCS for MAINTENANCE_IN + REPORT_TARGET_PORT_GROUPS.
	 */
	buf[5]	= 0x80;
       ******
}

```

tangwenji:
Hey, @mikechristie 
I checked the spc4r36q，the RTPG command is only related to the TPGs field in standard inquiry response data。
I give up this change and close this pr ?
I apologize for the distress。

mikechristie:
It's no problem. Go ahead and close.

PR #211, opened 2017-07-24
runner: use ppoll instead of poll

pkalever:
Since we do not want to restart poll from kernel

Suggested-by: Mike Christie <mchristi@redhat.com>

### DIFF
--- a/consumer.c
+++ b/consumer.c
@@ -18,7 +18,8 @@
  * userspace passthrough devices.
  */
 
-#define _DEFAULT_SOURCE
+#define _GNU_SOURCE
+#define _BITS_UIO_H
 #include <stdlib.h>
 #include <stdarg.h>
 #include <stdio.h>
@@ -31,11 +32,11 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <sys/mman.h>
+#include <signal.h>
 #include <poll.h>
 
 #include <stdint.h>
 #include <scsi/scsi.h>
-#define _BITS_UIO_H
 #include "target_core_user_local.h"
 #include "libtcmu.h"
 #include "scsi_defs.h"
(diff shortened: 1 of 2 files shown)


mikechristie:
Hey just debating this with myself.

I know using ppoll was my idea, but I am still not sure if it is uio bug we are hacking around or if we are supposed to be using ppoll for our use. The kernel engineer in me wants to fix the uio/poll code. The lazy userspace engineer in me wants to say merge the patch :)

I think we should at least add a comment in the code and/or better patch description. Maybe:

Use ppoll instead poll to avoid poll call reschedules during signal handling. If we were removing a device, then the uio device's memory could be freed, but the poll would be rescheduled and end up accessing the released device.

pkalever:
> Hey just debating this with myself.
> 
> I know using ppoll was my idea, but I am still not sure if it is uio bug we are hacking around or if we are supposed to be using ppoll for our use. The kernel engineer in me wants to fix the uio/poll code. The lazy userspace engineer in me wants to say merge the patch :)

IMHO, we should let this go as it is a easy/quick fix, and in any point using ppoll is not a bad idea at all. And once you get sometime, may be you can fix it in kernel side ? 

> 
> I think we should at least add a comment in the code and/or better patch description. Maybe:
> 
> Use ppoll instead poll to avoid poll call reschedules during signal handling. If we were removing a device, then the uio device's memory could be freed, but the poll would be rescheduled and end up accessing the released device.

Sure. I can do that.

Thanks :-)

mikechristie:
Ok. Add the requested comment to the code and git commit message, so we can one day figure out why we used ppol. I will merge it for now.

pkalever:
I have refreshed the patch, by adding the comment to git commit. May be you also wanted comment in the code ? Let me do that as well then.

pkalever:
Here is the refreshed version of patch.

mikechristie:
Ok. Thanks.

PR #213, opened 2017-07-24
defend for failures in lookup_dev_by_name

pkalever:


libtcmu.c
```diff
@@ -138,6 +138,12 @@ lookup_dev_by_name(struct tcmulib_context *ctx, char *dev_name, int *index)
 	darray_foreach(dev_ptr, ctx->devices) {
 		dev = *dev_ptr;
 		size_t len = strnlen(dev->dev_name, sizeof(dev->dev_name));
+		if (len == sizeof(dev->dev_name)) {
+			char err_dev[256];
```
mikechristie: Where did 256 come from?

I think the max len it could be is going to be strlen(uio) + max number of chars to represent an int which I think is 10 + plus null char.

Maybe add a define for the magic number we use, so we can change it later.

I am not in favor of adding extra checks because we think there is a bug and we can't root cause it. The input we get at the entry points should be checked and once it passes, the rest of the code does not need to double check it. It will get crazy with tests if we have to double check everything.

For example, I think you should add checks in open_devices that the sysfs name we are using is the format we expect, and in handle_netlink we should verify we got what we expected. Is nla_get_u32 using unsigned but other places are using a singed int an issue.

pkalever: > Where did 256 come from?
```
(gdb) bt
#0  0x00007ffff7bc7b72 in open_devices (ctx=0x62b8d0) at /root/tcmu-runner/libtcmu.c:619
#1  0x00007ffff7bc7e42 in tcmulib_initialize (handlers=0x62b880, handler_count=1) at /root/tcmu-runner/libtcmu.c:660
#2  0x000000000040f977 in main (argc=1, argv=0x7fffffffe538) at /root/tcmu-runner/main.c:917

(gdb) list
619                     ret = add_device(ctx, dirent_list[i]->d_name, buf);
620                     if (ret < 0)
621                             continue;

(gdb) p dirent_list[i]->d_name
$1 = "uio0\000\361u", '\000' <repeats 248 times>

(gdb) p sizeof(dirent_list[i]->d_name)
$2 = 256

```
Have noticed 32/16 in some other places as well, so just wanted to make the buf length constant, to avoid lose of data while cpy.

> 
> I think the max len it could be is going to be strlen(uio) + max number of chars to represent an int which I think is 10 + plus null char.
I understand this part, except the max uio dev string size.
So the max uio devices possible will be only sizeof(int) ?

> 
> Maybe add a define for the magic number we use, so we can change it later.
> 
> I am not in favor of adding extra checks because we think there is a bug and we can't root cause it. The input we get at the entry points should be checked and once it passes, the rest of the code does not need to double check it. It will get crazy with tests if we have to double check everything.

Agree.
It look likes with open_devices() path is_uio() is defending in a best possible way and
handle_netlink() path we already have a check on info->attrs[TCMU_ATTR_MINOR]
Both of them use snprintf which ensure null byte. So the Entry of dev_name looks okay.

> 
> For example, I think you should add checks in open_devices that the sysfs name we are using is the format we expect, and in handle_netlink we should verify we got what we expected.
IMO already taken care now.

> Is nla_get_u32 using unsigned but other places are using a singed int an issue.

Which other places ? was there a signed version of nla_get_u32 ? At least I did not find one.
Or are you saying at https://github.com/open-iscsi/tcmu-runner/blob/master/libtcmu.c#L220 we should use "%u" ? (IMO depends on the max devs numbers)

PR #214, opened 2017-07-24
rbd: register tcmu-runner handler with Ceph cluster

dillaman:
This will allow layered products to detect the RBD TMCU device

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -25,6 +25,7 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/utsname.h>
 #include <fcntl.h>
 #include <endian.h>
 #include <errno.h>
@@ -74,6 +75,10 @@ struct tcmu_rbd_state {
 
 	pthread_spinlock_t lock;	/* protect state */
 	int state;
+
+#ifdef LIBRADOS_SUPPORTS_SERVICES
+	bool lock_owner;
+#endif /* LIBRADOS_SUPPORTS_SERVICES */
 };
 
 struct rbd_aio_cb {
@@ -83,6 +88,22 @@ struct rbd_aio_cb {
 	int64_t length;
 	char *bounce_buffer;
 };
+
+#ifdef LIBRADOS_SUPPORTS_SERVICES
+static void tcmu_rbd_service_status_update(struct tcmu_device *dev)
+{
+	struct tcmu_rbd_state *state = tcmu_get_dev_private(dev);
+	int ret;
+	char status_buf[1024];
+	snprintf(status_buf, sizeof(status_buf), "%s%c%s%c",
+		 "lock_owner", '\0', state->lock_owner ? "true" : "false", '\0');
+	ret = rados_service_update_status(state->cluster, status_buf);
+	if (ret < 0) {
+		tcmu_dev_err(dev, "Could not update service status. (Err %d\n",
+			     ret);
+	}
+}
+#endif /* LIBRADOS_SUPPORTS_SERVICES */
 
 static void tcmu_rbd_image_close(struct tcmu_device *dev)
 {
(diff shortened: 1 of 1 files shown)


dillaman:
This is bare-bones registration that includes the static image spec details along with the dynamic lock owner status. It might be nice if we could provide status like the target iqn (if connected to iSCSI and if that is exposed down to tcmu-runner somewhere).

rbd.c
```diff
@@ -71,6 +75,22 @@ struct rbd_aio_cb {
...
 
+static void tcmu_rbd_service_status_update(struct tcmu_device *dev)
+{
+#ifdef LIBRADOS_SUPPORTS_SERVICES
+	struct tcmu_rbd_state *state = tcmu_get_dev_private(dev);
+
+	char status_buf[1024];
+	snprintf(status_buf, sizeof(status_buf), "%s%c%s%c",
+		 "lock_owner", '\0', state->lock_owner ? "true" : "false", '\0');
+	int ret = rados_service_update_status(state->cluster, status_buf);
```
dillaman: Note: this is a non-blocking call

rbd.c
```diff
@@ -290,9 +328,11 @@ static int tcmu_rbd_lock(struct tcmu_device *dev)
 	 * TODO: Add retry/timeout settings to handle windows/ESX.
 	 * Or, set to transitioning and grab the lock in the background.
 	 */
+	state->lock_owner = false;
```
lxbsz: This should also be nested in:
#ifdef LIBRADOS_SUPPORTS_SERVICES
...
#endif

rbd.c
```diff
@@ -290,9 +328,11 @@ static int tcmu_rbd_lock(struct tcmu_device *dev)
...
 	while (attempts++ < 5) {
 		ret = tcmu_rbd_has_lock(dev);
 		if (ret == 1) {
+			state->lock_owner = true;
```
lxbsz: And here.

rbd.c
```diff
@@ -330,6 +371,10 @@ static int tcmu_rbd_lock(struct tcmu_device *dev)
 static int tcmu_rbd_unlock(struct tcmu_device *dev)
 {
 	struct tcmu_rbd_state *state = tcmu_get_dev_private(dev);
+
+	state->lock_owner = false;
```
lxbsz: And here.

rbd.c
```diff
@@ -134,6 +154,24 @@ static int tcmu_rbd_image_open(struct tcmu_device *dev)
 		goto set_cluster_null;
 	}
 
+#ifdef LIBRADOS_SUPPORTS_SERVICES
+	char daemon_buf[1024];
```
lxbsz: This could be dynamically allocated as #200 has fix a similar issue to defence the stack clashing attack.

rbd.c
```diff
@@ -134,6 +154,24 @@ static int tcmu_rbd_image_open(struct tcmu_device *dev)
...
+	snprintf(daemon_buf, sizeof(daemon_buf), "%s/%s",
+		 state->pool_name, state->image_name);
+	char metadata_buf[1024];
```
lxbsz: And here.

mikechristie:
> This is bare-bones registration that includes the static image spec details along with the dynamic lock owner status. It might be nice if we could provide status like the target iqn (if connected to iSCSI and if that is exposed down to tcmu-runner somewhere).

It is exposed, but the problem is when we create the device, we do not know the target to LUN/device mapping yet. We are not guaranteed a target mapping until the first IO is received right now.

A SUSE engineer was looking into being able to notify runner for events like transport session creation (basically login for iscsi), and that would be a good time to get the info. We could also list what initiators are connected. I will ping them to see where they are.

If the format we return can change or we can just add more info by appending to the existing format, then I think it would be ok to add the info you have now and modify it as we are able to get more info.

dillaman:
@mikechristie Yeah -- it's free-form, so we can always add additional key/value pairs to the status dictionary in the future.

mikechristie:
I reworked this patch here:

https://github.com/open-iscsi/tcmu-runner/pull/248/commits/2d1c28d8263f188ac8c26ce76ca4fa905a3b24c9

@lxbsz  I handled your review comments.

One issue I guess we need to iron out is how large a buffer can we allocate on the stack. I am thinking up to a page is ok. An MB or more is too much, but to try and keep things simple and also because our stack does not get too deep 4K is ok. What do you think?

For my patch above, I did switch to dynamic allocations though. That was because I thought the rbd pool and image names could get long one day (the kernel seems to support really long names), and I did not want to have to deal with it plus if you disagree with 4k stack allocations then I have it handled too :)

PR #217, opened 2017-07-25
libtcmu: fix dev_added return check

pkalever:


mikechristie:
Thanks.

PR #218, opened 2017-07-25
libtcmu: mop-off cfgstring from remove_device

pkalever:

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -57,7 +57,7 @@ static struct nla_policy tcmu_attr_policy[TCMU_ATTR_MAX+1] = {
 static darray(struct tcmu_thread) g_threads = darray_new();
 
 static int add_device(struct tcmulib_context *ctx, char *dev_name, char *cfgstring);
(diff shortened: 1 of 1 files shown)


pkalever:
This is not needed, unless we wanted to use cfgstring for device lookup
i.e lookup_dev_by_name -> lookup_dev_by_cfgstring or club both of them for better defense.

mikechristie:
What does club in "club both of them for better defense" mean? Does it mean check/test both of them?

If you want to be extra safe, I think if you are using a kernel without the sync netlink support, then you could hit a issue like this:

1. tcmu sends remove netlink event for uio1. netink drops it.
2. tcmu does not wait for reply and cannot handle failures here so it still removes uio device which frees minor value 1.
3. tcmu gets add device request. It creates device and uio gets minor value 1.
4. The addition netlink event is able to reach userspace.
5. If there is a successful remove then you would end up removing the first one found.

I think you would then have 2 devices with the same dev_name, but possibly for different lio/tcmu devices. The lio device name is represented by {tcm_dev_name and tcm_hba_name}.

I think this would normally be rare, but it could happen if tcmu-runner was not running when you run the device removal command.

PR #219, opened 2017-07-25
libtcmu: fix cfgstring len

pkalever:
cfgstring will look like,
test@192.168.124.208/block-store/c2d55250-9d9d-4965-bfa1-43ce1b3aa5fc (glfs)

The string can exceed 256 characters.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -545,7 +545,7 @@ static int is_uio(const struct dirent *dirent)
 {
 	int fd;
 	char *tmp_path;
-	char buf[256] = {'\0'};
+	char buf[CFGSTRING_MAX] = {'\0'};
 	ssize_t ret = 0;
 
 	if (strncmp(dirent->d_name, "uio", 3))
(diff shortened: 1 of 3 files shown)


pkalever:
The cfgstring is handler specific, what works best for us ? What can be the MAX value ? 
We can make the CFGSTRING_MAX considering all handlers.
I have made this based on glfs cfgstring for now.

libtcmu.c
```diff
@@ -591,7 +591,7 @@ static int open_devices(struct tcmulib_context *ctx)
 
 	for (i = 0; i < num_devs; i++) {
 		char *tmp_path;
-		char buf[256] = {'\0'};
+		char buf[CFGSTRING_MAX] = {'\0'};
```
mikechristie: We get the cfgstring in those 2 functions from a sysfs file so it is limited to a page (4k on most archs). I think that is the only tcmu/tcmu-runner restriction. I think LIO is going to be limited by a page from its configfs use.

Could you move the the common file open/read related parts of open_devices and is_uio to a helper and call it. We will just have the one place to worry about it for the file reading case then.

For the max cfgstring name,  would it be easier to just strdup whatever we get from the kernel and find in the sysfs file? runner at least never has to worry about it. For the max sysfs file size, we could just do stat on "/sys/class/uio/%s/name" to get the initial max size for the read buffer.

pkalever:
Oops! 
Sorry for deleting this branch, it was by mistake.
Made a new pull at https://github.com/open-iscsi/tcmu-runner/pull/278

PR #220, opened 2017-07-26
Mode sense

mikechristie:
Fixes for mode sense and the cache page:

1. Patch 1 fixes several issues in the mode sense handling of the allocation length field and short buffers.
2. Patch 2 allows handlers to report different WCE values.
3. The rest of the patches covert handlers to #2 so they report a WCE value they support.

### DIFF
--- a/api.c
+++ b/api.c
@@ -740,34 +740,52 @@ int tcmu_emulate_read_capacity_16(
 	return SAM_STAT_GOOD;
 }
 
-int handle_rwrecovery_page(uint8_t *buf, size_t buf_len)
+static void copy_to_response_buf(uint8_t *to_buf, size_t to_len,
+				 uint8_t *from_buf, size_t from_len)
 {
-	if (buf_len < 12)
-		return -1;
+	if (!to_buf)
+		return;
+	/*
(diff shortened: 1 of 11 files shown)


api.c
```diff
@@ -846,72 +913,72 @@ int tcmu_emulate_mode_sense(
 	int i;
 	int ret;
 	size_t used_len;
-	uint8_t buf[512];
-	bool got_sense = false;
+	uint8_t *buf;
+        uint8_t *orig_buf = NULL;
```
mikechristie: Giving myself a review comment. Use tab here instead of spaces.

Also for a alloc_len of 0 I think I should have returned an ILLEGAL REQUEST instead of returning success.

mikechristie: Actually, I was right the first time, alloc_len of zero is not a failure.

mikechristie:
Adding a testing note:

MODE_SELECT support is broken. I do not think it ever worked, because it is checking for the page and subpage codes at the wrong offsets.. It is looking for them where MODE_SENSE sets them when it needs to find them in the parameter list data. 

We do not support setting any values, so will fix this in a separate PR.

PR #221, opened 2017-07-26
tcmur: add handle_try_passthrough helper support

lxbsz:
For the passthrough partially type handlers, likes the rbd handler,
will only passthrough some commands like XCOPY/CAW/WS..., the
.handle_cmd will be support in rbd's handler.

Sipmlify the code to let the developers wouldn't have to worry about
the handler is async or sync.


mikechristie:
Sorry for the delay.

PR #222, opened 2017-07-27
Alua misc fix

lxbsz:


alua.c
```diff
@@ -319,6 +323,8 @@ tcmu_get_tgt_port_grp(struct tcmu_device *dev, const char *name)
 	free(orig_str_val);
 	return group;
```
mikechristie: These patches are nice for future kernel changes but this is not possible right now is it?

The kernel always returns one of the checked values? If not then, we are missing some checks for valid values.

mikechristie:
Thanks.

alua.c
mikechristie: Oh yeah, I merged as is assuming it was for future proofing the code.

If there are other strings being returned let me know the kernel version and I can fix that.

lxbsz: Yes, from the current kernel code, this usually should not happen till now.

PR #224, opened 2017-07-31
alua: Fix MANAGEMENT PROTOCOL IN command sense code handling

lxbsz:
The error check here is actually the mistake by CDB data from scsi
client, should not be locally in target.


mikechristie:
Thanks.

PR #225, opened 2017-07-31
Add flush support for qcow

bwarrum-ibm:
Adds flush support to the qcow handler.
Removes O_SYNC flag and unconditionally sets WCE=1 for qcow.

I've tested the flush support for qcow using the script I've linked below.
There is a comment block at the top of the script that should explain its functionality.
I've tested this with and without the SYNCHRONIZE_CACHE command after a write (sg_sync in the test script), and have also tested with and without the O_SYNC flag set in qcow_open.
[test-qcow-sync.txt](https://github.com/open-iscsi/tcmu-runner/files/1188560/test-qcow-sync.txt)

### DIFF
--- a/qcow.c
+++ b/qcow.c
@@ -1407,11 +1407,11 @@ static int qcow_open(struct tcmu_device *dev)
 	tcmu_dbg("%s\n", config);
 
 	/*
-	 * We do not implement flush, so set WCE=0 and do sync IO.
+	 * Force WCE=1 until we support reconfig for WCE
 	 */
-	tcmu_set_dev_write_cache_enabled(dev, 0);
+	tcmu_set_dev_write_cache_enabled(dev, 1);
 
-	if (bdev_open(bdev, AT_FDCWD, config, O_RDWR | O_SYNC) == -1)
+	if (bdev_open(bdev, AT_FDCWD, config, O_RDWR) == -1)
 		goto err;
 	return 0;
 err:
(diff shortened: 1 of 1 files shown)


bgly:
@mikechristie So reviewing this and the mode sense stuff, that you had pushed up before that would default devices to a specific "write_cache" shouldn't we reflect this back to kernel? For example if we get attribute emulate_write_cache it will always default to the 0 that is preset in kernel. So when the netlink to call TCMU_CMD_ADDED_DEVICE occurs, at the end after it calls ret = dev->handler->added(dev); we can probably add something to also send back the emulate_write_cache attribute that is being set per device. Then put it back in the send_netlink_reply.

qcow.c
```diff
@@ -1479,6 +1479,23 @@ static int qcow_write(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 	return 0;
 }
 
+static int qcow_flush(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
+{
+	struct bdev *bdev = tcmu_get_dev_private(dev);
+	int ret;
+
+	if (fsync(bdev->fd)) {
+		tcmu_err("sync failed\n");
```
mikechristie: Do a tcmu_dev_err(dev, ) so we know which device failed.

mikechristie:
@bgly,

I wanted to just have runner set the settings it ended up using by writing to the attribute config fs files.

The problem was the tcmu_dev_configured test. Right now it returns true during addition because it checks for the uio device which is setup before we do the netlink command to runner. Runner then fails if the handler does not have a reconfiguration handler for the value being changed.

I was thinking we could just change tcmu_dev_configured to test for DF_CONFIGURED because that is set after the target_core_user->configure_device callout which does TCMU_CMD_ADDED_DEVICE.

bgly:
@mikechristie 

okay, I think it should work. I think this patch can go in and another PR to do the reconfig of emulate write cache to allow for toggling it to off, which would essentially reopen the device but as OSYNC, which solves our problems.

PR #226, opened 2017-08-02
UNMAP: set VPD MAX UNMAP LBA COUNT to device hw_max_sectors

lxbsz:


mikechristie:
Thanks.

mikechristie:
Just adding a reminder. We might will also need to set the unmap alignment and grain values on some OSs. Will test some more.

PR #229, opened 2017-08-04
Implicit failover support

mikechristie:
These patches add support for implicit failover support with rbd.

If using the ceph-iscsi-cli/ansible tools this patch is required:
https://github.com/mikechristie/ceph-iscsi-config/tree/ano

If doing manual setup, the alua_access_type = "Implicit" (configfs int val 1) should be set for your alua groups, and  the local group state would be active-optimized (AO) and the remote port would be active-non-optimized (ANO)

When using this with ESX, iscsi initiator recovery timeout must be = LIO's (2 * nopin_response_timeout + nopin_timeout).

Note the LIO nopin settings require a client ACL and can be found:
/sys/kernel/config/target/iscsi/target_name/tpgt_$N/acls/client_acl_name/attrib

For linux, the multipath.conf path_grouping_policy must be group_by_prio or failover and the fast_io_fail_tmo  (on older kernels the iscsi node.session.timeo.replacement_timeout must also be set) = LIO's (2 * nopin_response_timeout + nopin_timeout).

> devices {
>         device {
>                 vendor                          "LIO-ORG"
>                 hardware_handler        "1 alua"
>                 path_grouping_policy   group_by_prio
>                 path_selector                "round-robin 1 1"
>                 failback                          immediate
>                 path_checker                 tur
>                 prio                                alua
>                 no_path_retry                120
>                 fast_io_fail_tmo              25
>         }
> }

Windows has not yet been fully tested.

Note:

This also allows active/active to be setup, but there are still issues if IO is in flight so it is not safe for production.

ESX and linux are not yet production ready. This patchset does not handle the case where the iscsi connection is ok, but the ceph cluster is not reachable. Linux will failover but flip flop between paths, and newer versions of ESX will stay stuck on that path. This will be fixed in the next patches.


rbd.c
```diff
@@ -475,6 +475,19 @@ static void tcmu_rbd_close(struct tcmu_device *dev)
 	tcmu_rbd_state_free(state);
 }
 
+static int tcmu_rbd_handle_blacklisted_cmd(struct tcmu_device *dev,
+					   struct tcmulib_cmd *cmd)
+{
+       alua_lock_lost_notify(dev);
```
mikechristie: Note. This is a temp hack.

We are going to redo the command completion code so instead of the handlers setting the sense we will just have them return a TCMUR_CMD_ERR_XYZ. runner will then update the lock state. This will be done in the next patchset to support the case where the iscsi/target path is ok, but the link to the cluster has gone bad. That will add code to handle command time outs and forced completion.

scsi_defs.h
```diff
@@ -114,12 +115,13 @@ enum scsi_protocol {
 #define ASC_INVALID_FIELD_IN_PARAMETER_LIST     0x2600
 #define ASC_UNSUPPORTED_SEGMENT_DESC_TYPE_CODE  0x2609
 #define ASC_UNSUPPORTED_TARGET_DESC_TYPE_CODE   0x2607
+#define ASC_STATE_CHNAGED			0x2A06
```
dillaman: Nit: typo

rbd.c
```diff
@@ -254,7 +262,8 @@ static int tcmu_rbd_lock_break(struct tcmu_device *dev, char **orig_owner)
 
 	if (ret < 0) {
 		tcmu_dev_err(dev, "Could not get lock owners %d\n", ret);
-		return -EAGAIN;
+		if (ret != -ETIMEDOUT)
```
dillaman: If you time-out attempting to get the lock owners, I'd imagine you'd either need to retry or fail out of this function instead of continuing like out-params are valid.

rbd.c
```diff
@@ -490,6 +506,32 @@ static int tcmu_rbd_handle_blacklisted_cmd(struct tcmu_device *dev,
...
+ */
+static int tcmu_rbd_handle_timedout_cmd(struct tcmu_device *dev,
+					struct tcmulib_cmd *cmd)
+{
+	tcmu_dev_err(dev, "Timing out cmd.\n");
+	tcmu_notify_conn_lost(dev);
+
+	/*
+	 * TODO: For AA, we will want to kill the ceph tcp connections
+	 * with LINGER on and set to 0, so there are no TCP retires,
```
dillaman: Nit: typo

rbd.c
```diff
@@ -490,6 +506,32 @@ static int tcmu_rbd_handle_blacklisted_cmd(struct tcmu_device *dev,
...
+	 * and we need something on the OSD side to drop requests
+	 * that end up reaching it after the initiator's failover/recovery
+	 * timeout. For implicit and explicit FO, we will just disable
+	 * the iscsi port, and let the initiator switch paths which will
+	 * result in us getting blacklisted, so fail where with a retryable
```
dillaman: Nit: typo

rbd.c
mikechristie: Oops yeah a bug on my part. Fixed.

rbd.c
mikechristie: Both typos also fixed.

PR #230, opened 2017-08-07
rbd: Add write same support

lxbsz:
1, rbd add write same support and with the UNMAP bit support.
2, write same emulator add UNMAP bit support.
3, limit the MAX WRITE SAME LENGTH to "hw_max_sectors" when both the unmap and write same are supported.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -229,12 +229,14 @@ finish_err:
 
 static int handle_writesame_check(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 {
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
 	uint8_t *cdb = cmd->cdb;
 	uint8_t *sense = cmd->sense_buf;
 	uint32_t lba_cnt = tcmu_get_xfer_length(cdb);
 	uint32_t block_size = tcmu_get_dev_block_size(dev);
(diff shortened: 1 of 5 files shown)


lxbsz:
The next will add UNMAP bit support in WRITE SAME cmd.

tcmur_cmd_handler.c
```diff
@@ -330,12 +392,51 @@ static int handle_write_same(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	tcmur_writesame_fn_t write_same_fn = cmd->cmdstate;
+	uint32_t block_size = tcmu_get_dev_block_size(dev);
+	uint8_t *cdb = cmd->cdb;
+	uint64_t off = block_size * tcmu_get_lba(cdb);
+	uint32_t len = block_size * tcmu_get_xfer_length(cdb);
+	bool unmap = false;
+
+	if (rhandler->unmap && (cmd->cdb[1] & 0x08)) {
+		tcmu_dev_dbg(dev, "Do UNMAP in WRITE_SAME cmd!\n");
+		unmap = true;
```
mikechristie: It looks like you can drop the unmap code in rbd_aio_writesame, move the setting of the done function to handle_generic_cbk before the "if" above and then call rhandler->unmap here.

lxbsz: Yes, have updated.

mikechristie:
Thanks.

PR #231, opened 2017-08-07
api: VPD 0xB0 add OPTIMAL UNMAP GRANULARITY support

lxbsz:


mikechristie:
Thanks.

Why no granularity alignment though?

lxbsz:
Till now not very sure the use case about the alignment.

I have ever saw one VMWare bug about the UNMAP granularity alignment by Windows Guest OS:
when a Windows NTFS filesystem is formatted with 4KB blocks, Automated UNMAP is not working, but it is not caused by scsi proto, just the VMWare's bug. 
(http://cormachogan.com/2017/03/15/vsphere-6-5-p01-important-patch-users-automated-unmap)

mikechristie:
There is an example in the linux kernel in __blkdev_issue_discard.

On the device side, if the discard/unmap is not aligned then it could cause blocks to not be freed or cause it to hit a slow path depending on the implementation.

PR #232, opened 2017-08-08
Fix cluster allocation for nearly-full qcow images

bwarrum-ibm:
The qcow handler's block allocation function originally
checked to make sure the host offset of the cluster being allocated
would not exceed the size specified in the qcow image header.
This is improper behavior, because it counted qcow metadata clusters
towards the virtual image size. This change removes the improper
check from the block allocation function and moves the bound checking
into pwritev.

### DIFF
--- a/qcow.c
+++ b/qcow.c
@@ -501,6 +501,11 @@ static int qcow_image_open(struct bdev *bdev, int dirfd, const char *pathname, i
 		tcmu_err("Image size too big\n");
 		goto fail;
 	}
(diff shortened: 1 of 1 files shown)


mikechristie:
@bgly  Did you also review this already too?

Do you and @bwarrum-ibm work in the same group?

bgly:
@mikechristie Yes I have reviewed it and it looks good. Yes, he works in my team.

mikechristie:
Thanks.

PR #233, opened 2017-08-08
api: remove extra useless code

lxbsz:
Since the INQUIRY cmd handler is always called by tcmur_cmdproc_thread,
created by dev->handler->added(), just after rhandler->open(). And
before the thread is alive all the possible dev->max_xfer_len setting
is done. So the code here is useless.


mikechristie:
It was for non tcmu-runner libtcmu users.

I guess we broke that with the unmap changes which added the tcmur_handler references.

PR #235, opened 2017-08-08
[RFC] libtcmu: do not install headers and drop libtcmu stable API

mikechristie:
When libtcmu was created it was really early in the project and
missing lots of features. It is now difficult to maintain that API
and there may not be any users considering how many things
it was missing and we have not got any patches from non
tcmu-runner users in a long time.

The project will of course not change license and remain Apache
Version 2.0, so any one can just use the code in their own
project, but in future versions the stable libtcmu will not be
maintained.

If there are users, please speak up. It is a RFC patch, so I am open
to working something out. Are you using the netlink and uio ib functions,
but not the scsi and iovec ones in api.c?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ add_library(tcmu
   )
 set_target_properties(tcmu
   PROPERTIES
-  SOVERSION "1"
+  SOVERSION "2"
   )
 target_include_directories(tcmu
   PUBLIC ${LIBNL_INCLUDE_DIR}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -256,8 +256,6 @@ configure_file (
   )
 install(SCRIPT tcmu.conf_install.cmake)
 
-install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
-  DESTINATION include)
 install(FILES org.kernel.TCMUService1.service
   DESTINATION /usr/share/dbus-1/system-services)
 install(FILES tcmu-runner.conf DESTINATION /etc/dbus-1/system.d)


bgly:
@mikechristie I was planning on putting resources into moving qemu-tcmu forward. From what I recall I believe the patch utilizes the libtcmu api and also requires the headers. @famz 

https://patchwork.kernel.org/patch/9412171/

I know it hasn't been worked on in awhile but, I have been debating whether to fully port qcow into tcmu or to get qemu-tcmu working and that seems more beneficial and better support long term. What do you think Mike?

mikechristie:
@famz, was your issue with using tcmu-runner that you found bugs? If that was it, I think it would be best to fix the bugs in tcmu-runner and rewrite the qemu-tcmu code as runner handler. You will get so many things for free.

We have not been adding libtcmu features that work without tcmu-runner, so you will have to cut and paste the code between runner and your daemon. For example, it will be missing vmware VAAI support, HA (failover, distributed PGRs when it is done, etc), dynamic reconfigure, and more complete SCSI spec support. If your handler is in our repo, we will update it with all the other handlers.

We also have a lot of common bugs that we could clean up in one place like daemon restart when IO is in flight.

If you are planning on using some other scsi command emulation support like mentioned in your patch (scsi-disk use reference), then maybe the idea where we only have a lib that just handles the kernel interaction (uio and netlink) would be best.

We could also come up with a new lib interface since we have a better understanding of what commands we need to support and LIO/tcmu's design issues.

famz:
Thanks @bgly. I haven't been able to allocate my time on pushing qemu-tcmu forward. but qemu-tcmu still makes sense to me for tcmu to incorperate qcow2, gluster, ceph and all the other block features of QEMU. Red Hat has been productionizing QEMU and upstreaming all the work, and qcow2 is one of the most important functionality of it.

famz:
> If you are planning on using some other scsi command emulation support like mentioned in your patch (scsi-disk use reference), then maybe the idea where we only have a lib that just handles the kernel interaction (uio and netlink) would be best.

Yes, this sounds fine to me.

famz:
> We could also come up with a new lib interface since we have a better understanding of what commands we need to support and LIO/tcmu's design issues.

@mikechristie FWIW, it won't be a big problem for me, because the patch is till RFC.

bgly:
@famz So is the plan to copy paste the code to add the missing VAAI, HA, etc support.

Then to use scsi-disk emulation code instead? If so I can start looking into doing that

famz:
@bgly I haven't checked the current status of libtcmu/tcmu-runner for months, and have little clue about VAAI,HA, etc. in qemu-tcmu.

The point of qemu-tcmu is avoid duplicated work of qcow2, gluster, etc. So if the code is non-trivial, copy paste is not a great idea.

bgly:
@mikechristie I think it might be easier to have it the way it is now and utilize libtcmu. QEMU doesn't have any libraries and they run everything via executables.... 

@famz I don't know what went into having Vmware VAAI support but for ALUA stuff all that can be copied into INQUIRY handling/Maintenance In/Out. Reconfiguration is a trickier... 

I will start with getting the RFC patch updated to work with current level of TCMU and see what we need.

famz:
> I will start with getting the RFC patch updated to work with current level of TCMU and see what we need.

Sounds good to me, thanks @bgly

mikechristie:
I am merging this patch and bumping the .so version.

For the next release we can design something that will work well for us.

PR #239, opened 2017-08-10
build: don't overwrite tcmu.conf

mikechristie:
If there is a tcmu.conf do not overwrite it.

Signed-off-by: Mike Christie <mchristi@redhat.com>


lxbsz:
Yes, this is needed.

PR #243, opened 2017-08-13
unmap runner: fix error handling

mikechristie:
If we have done N unmap descr calls, but then N + 1 fails, we need
    to return TCMU_ASYNC_HANDLED, so we do the aio tracking properly
    and so libtcmu does not complete the command while there are
    unmap calls in flight.

    Also, if a descr unmap call fails we need to save the result and
    return it at the end.


mikechristie:
@lxbsz please review when you get a chance.

mikechristie:
@lxbsz I am going to close this with the assumption you will handle these errors in your patchset since you started to add code to handle it, and I do not want my patch to break yours.

lxbsz:
Sorry again for missing this PR yesterday.

You're so kindly and I will fix this and may will use "status" in your patch instead of "error".

Thanks

PR #245, opened 2017-08-13
WRITE SAME UNMAP add align and split feature support

lxbsz:
1, add handle_split_align_unmap helper support
2, write same and write same emulator's UNMAP add align and split feature support

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -704,6 +704,57 @@ static int ws_unmap_work_fn(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return rhandler->unmap(dev, cmd, offset, length);
 }
 
+static int handle_unmap_in_writesame(struct tcmu_device *dev,
+				     struct tcmulib_cmd *cmd)
+{
+	uint8_t *sense = cmd->sense_buf;
+	uint8_t *cdb = cmd->cdb;
+	uint64_t lba = tcmu_get_lba(cdb);
+	uint64_t nlbas = tcmu_get_xfer_length(cdb);
+	struct unmap_state *state;
+	unsigned int refcount;
+	int ret;
+
+	tcmu_dev_dbg(dev, "Do UNMAP in WRITE_SAME cmd!\n");
+
+	state = calloc(1, sizeof(*state));
+	if (!state) {
+		tcmu_dev_err(dev, "Failed to calloc memory for unmap_state!\n");
+		ret = tcmu_set_sense_data(sense, HARDWARE_ERROR,
+					  ASC_INTERNAL_TARGET_FAILURE,
+					  NULL);
+		return ret;
+	}
+
+	ret = pthread_mutex_init(&state->lock, NULL);
+	if (ret == -1) {
+		tcmu_dev_err(dev, "Failed to init spin lock in state!\n");
+		ret = tcmu_set_sense_data(sense, HARDWARE_ERROR,
+					  ASC_INTERNAL_TARGET_FAILURE,
+					  NULL);
+		free(state);
+		return ret;
+	}
+
+	state->refcount = 0;
+	state->error = false;
+	cmd->cmdstate = state;
+
+	pthread_mutex_lock(&state->lock);
+	ret = handle_split_align_unmap(dev, cmd, lba, nlbas);
+	if (ret != TCMU_ASYNC_HANDLED)
+		state->error = true;
+
+	refcount = state->refcount;
+	pthread_mutex_unlock(&state->lock);
+
+	/* Or will let the cbk to do the release */
+	if (!refcount)
+		free(state);
+
+	return ret;
+}
+
 static int handle_writesame(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 {
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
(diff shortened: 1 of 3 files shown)


tcmur_cmd_handler.c
```diff
@@ -181,6 +181,340 @@ struct write_same {
...
+		bddl -= 16;
+	}
+state_unlock:
+	refcount = state->refcount;
+	pthread_mutex_unlock(&state->lock);
+
+	/* Or will let the cbk to do the release */
+	if (!refcount)
+		free(state);
+
```
mikechristie: I think this still has the bug fixed in my patch

https://github.com/open-iscsi/tcmu-runner/pull/243

where if we failed after we have done a couple unmap calls, then we will return failure here, track_aio_request_finish will be called and the this code patch will complete the command. But refcount will be non-zero and so handle_unmap_cbk will also call the aio track function and complete the command.

tcmur_cmd_handler.c
```diff
@@ -181,6 +181,340 @@ struct write_same {
...
+	free(desc);
+	free(ucmd);
+
+	pthread_mutex_lock(&state->lock);
+	if (--state->refcount > 0) {
+		pthread_mutex_unlock(&state->lock);
+		return;
+	}
+	error = state->error;
```
mikechristie: See PR 243. I think this will hit a issue where if the Nth unmap call failed, but the N + 1 completed ok, ret will be SAM_STAT_GOOD and the sense will not get copied over and the error will not be returned to the initiator.

tcmur_cmd_handler.c
mikechristie: Actually the above comment is not completely correct.

I think the bug is:

 If handle_split_align_unmap fails in something like a allocation check then it returns SAM_STAT_CHECK_CONDITION and does not set state->error. In this case if we had done some unmap handler calls already, this code path would return the error and call track_aio_request_finish and also handle_unmap_cbk  would also do this since state->error is not set.

Also for the case where state->error is set, I do not think it is a good idea to call the libtcmu complete function while the handler is still processing the command. The kernel and runner's cmd counters will be off. For example, the kernel will have decremented the refcount on the device and will allow device deletions. tcmu and runner had assumed that this could not happen so we could end up trying to delete the device while the handler is still processing commands.

tcmur_cmd_handler.c
mikechristie: Oh yeah, for this case where a handler->unmap call may fail and/or a handle_split_align_unmap call may fail or both might, I think as long as one of the sense or status codes is returned it will be ok for now. We should probably prioritize them, but that looks like it will be a mess. In the next release I want to move the sense handling to runner, and then we can handle this.

mikechristie: Sorry, one more comment.

There is a bug in my patch in PR 243 and here.

If the handler->unmap call fails, then it copies sense to the ucmd->sense_buf. In this case we will need to copy that sense to origcmd->sense_buf.

tcmur_cmd_handler.c
lxbsz: Sorry for missing your previous PR #243.

Yes, you are right, no matter success or fail to handle all the ucmds, let the call back to do the clean up work is a good idea.

tcmur_cmd_handler.c
lxbsz: Just forgot and mindly ignored the errors  may encounter in ->unmap.

lxbsz:
Test output:

[root@wzy-2 ~]# sg_write_same --lba=2 --num=81923 /dev/sdh -v -S -U
Default data-out buffer set to 512 zeros
[root@wzy-2 ~]# sg_write_same --lba=2 --num=81923 /dev/sdh -v -R -U
Default data-out buffer set to 512 zeros
do_write_same: use WRITE SAME(16) instead of 10 byte cdb
[root@wzy-2 ~]#


[DEBUG] handle_writesame_check:673 block0: Start lba: 2, number of lba:: 16387, last lba: 81924                                                                                                                   
[DEBUG] handle_unmap_in_writesame:689 block0: Do UNMAP in WRITE_SAME cmd!
[DEBUG] handle_split_align_unmap:267 block0: OPTIMAL UNMAP GRANULARITY: 8192, UNMAP GRANULARITY ALIGNMENT: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 0: start lba: 2, end lba: 8191, lbas: 8190
[DEBUG] handle_split_align_unmap:307 block0: Split 1: start lba: 8192, end lba: 16383, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 2: start lba: 16384, end lba: 24575, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 3: start lba: 24576, end lba: 32767, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 4: start lba: 32768, end lba: 40959, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 5: start lba: 40960, end lba: 49151, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 6: start lba: 49152, end lba: 57343, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 7: start lba: 57344, end lba: 65535, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 8: start lba: 65536, end lba: 73727, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 9: start lba: 73728, end lba: 81919, lbas: 8192
[DEBUG] handle_split_align_unmap:307 block0: Split 10: start lba: 81920, end lba: 81924, lbas: 5

lxbsz:
Resolved the conflicts.

mikechristie:
Hey @lxbsz I just had some small cleanups to this patchset. Made a PR here:

https://github.com/open-iscsi/tcmu-runner/pull/250

PR #246, opened 2017-08-13
tcmu-runner service: increase default file limit

mikechristie:
It looks like we are hitting the fd limit with setups having
lots of devices and with ceph/gluster setups with a medium/high
amount (only 40 with ceph and 1000 with gluster). This patch
just increases the default limit.


pkalever:
Looks good to me :-)

PR #247, opened 2017-08-14
runner: make the usage visible without intalling tcmu.conf

lxbsz:
When new user who are not very familiar about the tcmu-runner,
his/her mostly will print the usage of this like:

./tcmu-runner -h
Failed to open file '/etc/tcmu/tcmu.conf', No such file or directory

This patch will make the usage visible even without intalling
the tcmu.conf file.

Signed-off-by: Zhuoyu Zhang <zhangzhuoyu@cmss.chinamobile.com>

### DIFF
--- a/main.c
+++ b/main.c
@@ -847,10 +847,6 @@ int main(int argc, char **argv)
 	guint reg_id;
 	int ret;
 
-	tcmu_cfg = tcmu_setup_config(NULL);
-	if (!tcmu_cfg)
-		exit(1);
-
 	while (1) {
 		int option_index = 0;
 		int c;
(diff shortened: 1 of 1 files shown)


mikechristie:
It seems ok to me, but how are you guys using runner without doing a make install on it?

lxbsz:
She was newer here and tring to help me to do the test about the runner. She just wanted to know how to use this.

I am thinking should we must install it when just doing a simple test? How about just using one tmp conf file instead if the /etc/tcmu/tcmu.conf is absent?

mikechristie:
The patch is fine.

Do not worry about it. I was just  wondering if you were doing some sort of non standard type of setup where config files and other stuff is not going to be available.

lxbsz:
Yes, if not use tmp conf file instead, I think we'd better to tell users that the correct and last step should include "make install" in README or where the /etc/tcmu/tcmu.conf comes from.

mikechristie:
Ok.

mikechristie:
Here is a PR for the make install step and also CMAKE_INSTALL_PREFIX use because it seems people are getting tripped by that too.

https://github.com/open-iscsi/tcmu-runner/pull/249

PR #248, opened 2017-08-14
misc fixes

mikechristie:
The following just are misc fixes and a rebased patch for PR 214.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -25,6 +25,7 @@
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/utsname.h>
 #include <fcntl.h>
 #include <endian.h>
 #include <errno.h>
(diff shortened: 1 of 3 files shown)


mikechristie:
@dillaman I reworked your services patch in this PR here

https://github.com/open-iscsi/tcmu-runner/pull/248/commits/2d1c28d8263f188ac8c26ce76ca4fa905a3b24c9

I had 2 questions though.

1. Does the service get unregistered when we do a rbd_close()?

2. Can you do service updates if we are blacklisted? One funny thing about the current patch is that all it does now is do a status update if we got the lock or if a lock call failed. There is no unlock call anymore because the initiators never really used it for explicit failover and implicit does not need it. We only know if we lost the lock if we get a ESHUTDOWN/EBLACKLIST error from a lock call or from a IO request.

If we can send service updates while blacklisted then I can add another service update call indicating we do not have the lock anymore.

dillaman:
@mikechristie The service is associated w/ the librados connection, not the image. It just so happens that tcmu-runner always opens a connection per-image.  I honestly don't know if the service updates can be sent after being blacklisted, but I am going to presume that they can be sent (since they are sent to the ceph-mgr, not the ceph-mon/ceph-osd).

dillaman:
@mikechristie The blacklist is entirely an OSD-level concept, so the client->manager connection should remain open after the primary is blacklisted.

rbd.c
```diff
@@ -76,6 +77,89 @@ struct rbd_aio_cb {
...
+	free(daemon_buf);
+	return ret;
+}
+
+#else
+
+static int tcmu_rbd_service_register(struct tcmu_device *dev)
+{
+	/* Ignorable. Just log in dbg mode just in case. */
+	tcmu_dev_dbg(dev, "Could not register service to ceph cluster.\n");
```
dillaman: Nit: should this message be something like "service registration not supported"?

mikechristie: Yeah, will fix.

mikechristie:
> @mikechristie The blacklist is entirely an OSD-level concept, so the client->manager connection should remain open after the primary is blacklisted.

Ok. Thanks.

I am going to merge with just the string fix. I will have to add the service update call in the blacklist detection code in the next go around just in case it changes timing  in a way that breaks failover/failbakck. I do not have time to retest everything tonight.

rbd.c
```diff
@@ -292,11 +380,39 @@ static int tcmu_rbd_lock(struct tcmu_device *dev)
...
+	int ret;
+
+	ret = rbd_get_features(state->image, &features);
+	if (ret) {
+		tcmu_dev_warn(dev, "Could not get rbd features. HA may not be supported. Err %d.\n", ret);
+		return;
+	}
+
+	if (!(features & RBD_FEATURE_EXCLUSIVE_LOCK)) {
+		tcmu_dev_warn(dev, "exclusive-lock not enabled for image. HA not be supported.\n");
```
mikechristie: Will fix this string too.

PR #249, opened 2017-08-15
doc: add build info

mikechristie:
Document:

1. CMAKE_INSTALL_PREFIX=/ will normally need to be used to install
tcmu-runner to where the systemd service requires it.

2. Add "make install"


lxbsz:
And also DSUPPORT_SYSTEMD=ON should also be passed sometimes, like:

cmake3 -DSUPPORT_SYSTEMD=ON -DCMAKE_INSTALL_PREFIX=/usr/ .

Or it could work on my case: CentOS Linux release 7.2.1511 (Core)

mikechristie:
> cmake3 -DSUPPORT_SYSTEMD=ON -DCMAKE_INSTALL_PREFIX=/usr/ .

Thanks. Added this and dropped my "DCMAKE_INSTALL_PREFIX=/" because yours is correct.

mikechristie:
I did not add "cmake3". That should be fixed now and older versions of cmake should work since this commit:

https://github.com/open-iscsi/tcmu-runner/pull/83

lxbsz:
Yes, it work with cmake 2.8 with here.

lxbsz:
This look good to me.

PR #250, opened 2017-08-15
Unmap split fixes

mikechristie:
Some trivial changes for the split and align changes.

1. Just some code comment cleanups and function rename.
2. Fix for possible mutex leak.


lxbsz:
This look good to me.

Test it and work fine.

PR #258, opened 2017-08-18
file_zbc: ZBC disk emulation handler

damien-lemoal:
file_zbc.c implements a ZBC disk emulation using a file as backend
storage. The handler is compiled as handler_file_zbc.

handler_file_zbc can emulate both host aware and host managed disks.
In both cases, the LBA size, zone size and the number of conventional
zones can be configured through options in the configuration string.
The configuration, together with zone information (zone type, zone
condition, write pointer position, etc) are stored as metadata at the
beginning of the backstore file, invisibly to the user. The
configuration and state of the emulated device is thus retained
accross reboots.

The metadata in the backstore file is reinitialized (i.e. the device
configuration and state reset) if the specified backstore file does
not exist, if the metadata in the file is incorrect or if configuration
options are specified in the configuration string. Otherwise, the
metadata is reused as is, restoring the emulated device state from its
last use.

The primary goal of handler_file_zbc is kernel features development
and tests. It is especially useful for people without access to SMR
disks or on systems on which those disks cannot be used
(e.g. a laptop).

The emulated device, used either through the loopback or iscsi fabrics,
successfully passes libzbc conformance tests (using kernel 4.13-rc as
many target patches fixing various problems with sense code and sense
keys reporting are needed). Testing was also performed using f2fs
(which natively supports host managed zoned block devices) and the
dm-zoned device mapper. Additionally, the emulated device is also
currently being used for btrfs zoned block device support development
and can also be very helpful in general to test Linux block and SCSI
stack support of zoned block devices.

The blktets framework would also be another area that can benefit from
ZBC emulation.

Signed-off-by: Damien Le Moal <damien.lemoal@wdc.com>

### DIFF
--- a/file_zbc.c
+++ b/file_zbc.c
@@ -1,0 +1,2202 @@
+/*
+ * Copyright 2017, Western Digital Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain
+ * a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+*/
+
+/*
+ * ZBC device emulation with a file backstore.
+ */
+
+#define _GNU_SOURCE
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <endian.h>
+#include <errno.h>
+#include <scsi/scsi.h>
+#include <linux/blkzoned.h>
+
+#include "scsi_defs.h"
+#include "libtcmu.h"
+#include "tcmu-runner.h"
+
+/*
+ * SCSI commands.
+ */
+#define ZBC_OUT					0x94
+#define ZBC_IN					0x95
+
+/*
+ * ZBC IN/OUT Sevice Actions
+ */
+#define ZBC_SA_REPORT_ZONES			0x00
+#define ZBC_SA_CLOSE_ZONE			0x01
+#define ZBC_SA_FINISH_ZONE			0x02
+#define ZBC_SA_OPEN_ZONE			0x03
+#define ZBC_SA_RESET_WP				0x04
+
+/*
+ * ZBC related additional sense codes.
+ */
+#define ASC_UNALIGNED_WRITE_COMMAND		0x2104
+#define ASC_WRITE_BOUNDARY_VIOLATION		0x2105
+#define ASC_ATTEMPT_TO_READ_INVALID_DATA	0x2106
+#define ASC_READ_BOUNDARY_VIOLATION		0x2107
+#define ASC_INSUFFICIENT_ZONE_RESOURCES		0x550E
+
+
+/*
+ * Device zone model.
+ */
+enum zbc_dev_model {
+	ZBC_HA = 0x00,
+	ZBC_HM = 0x14,
+};
+
+/*
+ * Reporting options.
+ */
+enum zbc_reporting_options {
+
+	/* List all of the zones in the device */
+	ZBC_RO_ALL		= 0x00,
+
+	/* List the zones with a Zone Condition of EMPTY */
+	ZBC_RO_EMPTY		= 0x01,
+
+	/* List the zones with a Zone Condition of IMPLICIT OPEN */
+	 ZBC_RO_IMP_OPEN	= 0x02,
+
+	/* List the zones with a Zone Condition of EXPLICIT OPEN */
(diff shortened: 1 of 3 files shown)


mikechristie:
Sorry for the delay. I have been on vacation. I will have comments by next week.

damien-lemoal:
Hi Mike,

Any comment regarding this pull request ?
Is there anything I can help clarify ?

Thank you.

mikechristie:
Sorry for the late response.

My only major comment is just about if we can combine this with file_example.c, so we only have the one handler that does IO to a file. I have been going back and forth with myself, and my first thought is that it is not really worth it as the zbc handler will be mostly used for testing and development, so let it be separate so you can do whatever crazy/interesting things you want to do.

I guess one question I had is do you think users will ever use this in production? Will users want to use it in some sort of setup where they do other things like multipating? Will you need stuff like ALUA? Will you need to support persistent reservations?

damien-lemoal:
Mike,

I initially also thought of having a ZBC mode with the file handler, but there was enough differences and changes that making a new handler ended up being easier, both development wise but also for maintenance as any change for the "regular" disk emulation would not necessitate also test the "zbc" side.

As you correctly state, the intent of the ZBC handler is testing and development. I would never ever recommend this for production. That would be nonsensical anyway. The premise for using SMR/ZBC drives is capacity. So emulating that on top of a file on a regular drive would give an application developer all the headaches of dealing with sequential writes while not providing the benefits of higher capacity. Production use makes no sense to me and I will always discourage it.
So features like multipathing/ALUA are not a requirement at all. The kernel does not support that for SMR drives anyway (think of the nightmare that these introduce if one wants to maintain a correct sequential write pattern to the disk sequential zones).

For the test and development use case, there is however a strong need from the kernel folks (at least my team), as well as application developers too (some of our customers are very interested). The biggest headache with SMR drives is the ecosystem. Many BIOSes and HBAs are simply not yet supporting those disks, in particular the host managed flavor. Things are slowly getting better and there are now HBAs supporting ZBC/ZAC. But while a system developer readies a production system for SMR, application and kernel developers can still use tcmu ZBC handler to develop and test their software readiness for SMR. Tcmu-runner with the ZBC handler is really a great tool for that as it can even also give you a bus analyzer level information. Some kernel maintainers have requested such a tool as a condition for accepting ZBC related code (e.g. BtrFS community as we are working on BtrFS support for ZBC).

Last but not the least, standards are evolving and ZBC-2.0 proposals are already being worked on, as well as a dynamic SMR version (where zone types can be changed dynamically to trade capacity vs ease of use). The ZBC handler will also be a great tool for validating standard choices *before* these choices make it to the final standard (somewhat the same model as was used for NVMe development with the standard and kernel support being developed simultaneously and cross validated).

Best regards.

file_zbc.c
```diff
@@ -0,0 +1,2240 @@
...
+	"  will create a host-managed disk with 128 MiB zones and 100\n"
+	"  conventional zones, stored in the file /var/local/zbc.raw\n";
+
+static struct tcmur_handler zbc_handler = {
+	.cfg_desc = zbc_cfg_desc,
+
+	.name = "ZBC Emulation Handler",
+	.subtype = "zbc",
+
+	.check_config = zbc_check_config,
```
mikechristie: The check_config callout was a hack around the kernel/runner not being able to propagate errors at open() time. We are removing this callout, so you should make sure these checks are done at open() time and a -EXYZ error code is returned.

file_zbc.c
```diff
@@ -0,0 +1,2240 @@
...
+	"                      The default is 512\n"
+	"  zsize-<size(MiB)> : Zone size in MiB. The default is 256 MiB.\n"
+	"  conv-<num>        : Number of conventional zones at LBA 0 (can be 0)\n"
+	"                      The default is 1%% of the device capacity\n"
+	"  open-<num>        : Optimal (HA) or maximum (HM) number of open zones\n"
+	"                      The default is 128\n"
+	"Ex:\n"
+	"  cfgstring=model-HM/zsize-128/conv-100@/var/local/zbc.raw\n"
+	"  will create a host-managed disk with 128 MiB zones and 100\n"
+	"  conventional zones, stored in the file /var/local/zbc.raw\n";
```
mikechristie: Add this info to the README in the section "Creating a LIO user-backed storage object with targetcli-fb or configfs" with the formats for the other handlers.

file_zbc.c
```diff
@@ -0,0 +1,2240 @@
...
+		return ret;
+
+	/* For writes, check zone boundary crossing */
+	zone = zbc_get_zone(zdev, lba, false);
+	if (lba + nr_lbas > zone->start + zone->len) {
+		tcmu_dev_err(dev, "Write boundary violation lba %llu, xfer len %lu\n",
+			     lba, nr_lbas);
+		return tcmu_set_sense_data(cmd->sense_buf,
+					   ILLEGAL_REQUEST,
+					   ASC_WRITE_BOUNDARY_VIOLATION,
```
mikechristie: Don't forget to include this addition to scsi_def.h.

mikechristie: Ignore this comment. I was going to ask you to move the asc/ascq defs to scsi_def.h, but then flip flopped and was ok just leaving them in this file for now since you will be the only user.

mikechristie:
Sounds good. Just handle the basic review comments and I will merge it up.

damien-lemoal:
Mike,

Just updated the pull according to your comments. Please check.
In particular, the fixes to the README.md file are more extended than you requested. I fixed the markup in different places to improve the display in the top page.
Thank you.

Best regards.

mikechristie:
Thanks, and sorry again for the delay.

PR #263, opened 2017-08-21
tcmulib dbus registration changes

bwarrum-ibm:
Fixes issues with dbus registration of tcmulib handlers.
tcmu-runner can now handle tcmulib processes that are killed or release
the dbus name for a specific subtype.
tcmulib daemon processes can now handle situations where the tcmu-runner
service becomes unavailable.

Signed-off-by: Bryant G. Ly bryantly@linux.vnet.ibm.com

Prior to this change, the TCMUService1.HandlerManager1 DBus interface exposed two methods - RegisterHandler and UnregisterHandler.  The UnregisterHandler method did not have a corresponding tcmulib function.  When a tcmulib process died without calling the UnregisterHandler method, tcmu-runner would continue watching the bus name for all subtypes registered by the tcmulib process (on_register_handler() in main.c).  When the tcmulib process restarted, tcmu-runner watched the same bus name again.  This caused all of the watch_name callbacks (on_handler_vanished, on_handler_appeared) to be called more than once for each corresponding dbus event.  This caused an assortment of problems - one of which caused tcmu-runner to segfault in glib due to a null pointer dereference with the following error messages after restarting the tcmulib daemon process 3 times in a row:
```
Aug 15 18:56:32 systemd[1]: Started LIO Userspace-passthrough daemon.
Aug 15 18:57:04 tcmu-runner[39324]: (process:39324): GLib-GIO-CRITICAL **: g_dbus_object_manager_server_unexport_unlocked: assertion 'g_variant_is_object_path (object_path)' failed
Aug 15 18:57:12 tcmu-runner[39324]: (process:39324): GLib-GIO-CRITICAL **: g_dbus_object_skeleton_new: assertion 'g_variant_is_object_path (object_path)' failed
Aug 15 18:57:12 tcmu-runner[39324]: (process:39324): GLib-GIO-CRITICAL **: g_dbus_object_skeleton_add_interface: assertion 'G_IS_DBUS_OBJECT_SKELETON (object)' failed
Aug 15 18:57:12 systemd[1]: tcmu-runner.service: Main process exited, code=dumped, status=11/SEGV
Aug 15 18:57:12 systemd[1]: tcmu-runner.service: Unit entered failed state.
Aug 15 18:57:12 systemd[1]: tcmu-runner.service: Failed with result 'core-dump'.
```
Additionally, tcmulib-register was incapable of handling situations where tcmu-runner was not running when the tcmulib process was started, or where tcmu-runner was restarted while the tcmulib process was running (causing the TCMUService name to vanish and reappear).  The result was that the subtypes for the tcmulib process would not reappear in targetcli until the tcmulib process was restarted.

This change drops the UnregisterHandler method from the HandlerManager DBus interface.  tcmu-runner now stops watching the DBus name for a tcmulib subtype when the name vanishes. The RegisterHandler call is now triggered every time the TCMUService name becomes available and a subtype bus name becomes owned by the tcmulib process.

I was unsure of whether or not this would require a version number change to HandlerManager1.  Since the UnregisterHandler method was never used, I assumed it was not necessary.  I'm also not entirely sure how to handle/report DBus errors during the registration process.

### DIFF
--- a/libtcmu-register.c
+++ b/libtcmu-register.c
@@ -66,7 +66,7 @@ tcmulib_register_handler(struct tcmulib_context *ctx,
 
 static void tcmulib_reg_fail(struct tcmulib_context *ctx)
 {
-	ctx->reg_count_down = 0;
+	// TODO: Report failures back to process performing registration
 }
 
 static gboolean
@@ -129,15 +129,11 @@ tcmulib_reg_name_acquired(GDBusConnection *connection,
 	struct tcmulib_handler *handler = user_data;
 	struct tcmulib_context *ctx = handler->ctx;
 
-	if (!ctx->reg_count_down || --ctx->reg_count_down)
-		return;
-	/* We've acquired all needed buses, now register each handler to
-	 * org.kernel.TCMUService1.HandlerManager1. */
-	darray_foreach(handler, ctx->handlers) {
-		if (!tcmulib_register_handler(ctx, handler)) {
-			tcmulib_reg_fail(ctx);
-			break;
-		}
+	handler->connection = connection;
+
+	if (ctx->connection) {
+		/* The primary service is already available.  Register immediately. */
+		tcmulib_register_handler(ctx, handler);
 	}
 }
 
@@ -147,8 +143,7 @@ tcmulib_reg_name_lost(GDBusConnection *connection,
 		      gpointer         user_data)
 {
 	struct tcmulib_handler *handler = user_data;
-
-	tcmulib_reg_fail(handler->ctx);
+	handler->connection = NULL;
 }
 
 static void tcmulib_handler_own_bus(struct tcmulib_handler *handler)
@@ -176,12 +171,12 @@ tcmulib_reg_name_appeared(GDBusConnection *connection,
 	struct tcmulib_handler *handler;
 
 	ctx->connection = connection;
-	if (!ctx->reg_count_down)
-		return;
+	/* Primary TCMU service is now available. None of the handlers are
+	   registered, so register all handlers that have acquired their bus name */
 	darray_foreach(handler, ctx->handlers) {
-		/* XXX: Set this at initalize time. */
-		handler->ctx = ctx;
-		tcmulib_handler_own_bus(handler);
+		if (handler->connection) {
+			tcmulib_register_handler(ctx, handler);
+		}
 	}
 }
 
@@ -190,15 +185,21 @@ tcmulib_reg_name_vanished(GDBusConnection *connection,
 			  const gchar     *name,
 			  gpointer         user_data)
 {
-	tcmu_err("Failed to get bus %s\n", name);
+	struct tcmulib_context *ctx = user_data;
+
+	ctx->connection = NULL;
 }
 
 void tcmulib_register(struct tcmulib_context *ctx)
 {
-	assert(!ctx->reg_count_down);
-	ctx->reg_count_down = darray_size(ctx->handlers);
-	if (!ctx->reg_count_down)
-		return;
+	struct tcmulib_handler *handler;
+
+	/* Start acquiring buses for each subtype owned by this context. */
+	darray_foreach(handler, ctx->handlers) {
+		tcmulib_handler_own_bus(handler);
+	}
+
+	/* Start waiting for the primary service to become available */
 	g_bus_watch_name(G_BUS_TYPE_SYSTEM,
 			 "org.kernel.TCMUService1",
 			 G_BUS_NAME_WATCHER_FLAGS_NONE,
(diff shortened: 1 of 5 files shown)


mikechristie:
We were actually trying to drop this code. The only user is targetcli to get the handler names. Patch is ok for now. Just revert the coding style changes.

bgly:
For qemu-tcmu to work we still need this section of the code

mikechristie:
What does qemu tcmu use dbus for? The discovery part or initializing the daemon because the systemd script is not being run?

bgly:
As of now qemu-tcmu uses these functions:

tcmulib_command_complete
tcmulib_processing_complete
tcmulib_get_next_command
tcmulib_master_fd_ready
tcmulib_get_master_fd
tcmulib_initialize
tcmulib_register

So it basically needs tcmulib_initialize/tcmulib_register/tcmulib_get_master_fd to initialize the daemon.

mikechristie:
Ok it's kind of a pain to look at the code on my phone but it looks like those don't use dbus but rather set up things that dbus calls use later.

They are also all lib functions where the internals are not exposed so we can change them. I was more comcerned about the open dbus calls like targetcli fb does.

bgly:
@famz It seems like you aren't allowed to link apache into GPLv2 unless its GPLv3? So did you know of something I do not know?

famz:
@bgly Hmm, yes, TBH I have no idea, maybe we could ask libtcmu for a re-license under GPL?

bgly:
@famz Well i have a working copy of qemu-tcmu on latest qemu/tcmu-runner. The only problem is if there is license incompatibility; I don't see a point in continuing to get it to work

mikechristie:
I have no preference for the license type. Andy preferred apache so that is why it is licensed as that.

To relicense we would have to go through the git commits and ask everyone. If we are making a new lib though, I am open to whatever.

famz:
OK, so it seems the options for the qemu-tcmu work to move on are:

 1. relicensing current libtcmu as a GPLv2 compatible library
 2. making a new lib licensed such
 3. (re)implement whatever is needed in QEMU so it doesn't require libtcmu

All in all I think option 2 is the best solution.

libtcmu-register.c
```diff
@@ -190,21 +185,27 @@ tcmulib_reg_name_vanished(GDBusConnection *connection,
...
-			 tcmulib_reg_name_appeared,
-			 tcmulib_reg_name_vanished,
-			 ctx,
-			 NULL);
+		"org.kernel.TCMUService1",
+		G_BUS_NAME_WATCHER_FLAGS_NONE,
+		tcmulib_reg_name_appeared,
+		tcmulib_reg_name_vanished,
+		ctx,
+		NULL);
```
mikechristie: Sorry for the delay on this one. Just noticed my comment never went through (mobile interface I was using on vacation is different from normal browser and put my comment in pending).

I am fine with this patch for now. My only comment was to just revert this coding style chance as the existing is more standard for the tcmu-runner code.

bwarrum-ibm: This was unintentional, fixed this and a couple of white space problems.

mikechristie:
Thanks.

amarts:
@bgly @famz is there any update on licencing for qemu-tcmu ?

We are very much interested in getting qemu-tcmu efforts being successfully merged in the mainstream projects, for [gluster-block](https://github.com/gluster/gluster-block) project.

Wanted to know what is the latest on licencing issues for the project? and is there anyway we can participate in taking this to resolution.

bgly:
I have a sandbox somewhere that has qemu-tcmu working with some modifications. - I don't have time to dedicate much effort into this since we have other pressing issues at the moment. 

It seems like Apache v2 can be included in GPLv3. I know qemu is largely licensed under GPLv2, but if this qemu-tcmu stuff can be licensed under GPLv3 and put into qemu then that can work?

baihuahua:
FYI, these days i worked on making qemu-tcmu move forward based on qemu-tcmu v2 with fam's help and it looks pretty ready for upstreaming. As for license incompatibility between qemu-tcmu and libtcmu, i'm also  trying to re-implement libtcmu in qemu. But there still need some work in targetcli-fb and/or rtslib-fb...

bgly:
@baihuahua  What kind of work and can you create a repo for access?

baihuahua:
@bgly @famz @mikechristie @amarts  [Qemu-tcmu v3](https://github.com/baihuahua/qemu/tree/qemu-tcmu) is here. It supports multiple targets in the same process and you can configure it through parameters or targetcli. Qemu-tcmu doesn't relys on tcmu-runner now and you just need 1. start qemu-tcmu 2. configure targets. That's all! But there're still some issues:
1. As a prototype we just copy tcmu-runner's libtcmu directly with some necessary modifications. It needs more cleanups  and re-implementation.
2. As we droped libtcmu dbus's registration, we can use lio-utils to configure targets for now.
3. More code cleanups.

baihuahua:
FYI, i just bring libtcmu dbus registration back. Let's use targetcli instead of lio-utils. 
qq 20180309155456

PR #265, opened 2017-08-22
rbd: Add timers check for cluster osd

lxbsz:
The rados osd op timeout must be longer than the timeouts to detect
unreachable OSDs (osd heartbeat grace + osd heartbeat interval) or
we will end up failing the transport connection when we just needed
to try a different OSD.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -176,6 +176,70 @@ static void tcmu_rbd_image_close(struct tcmu_device *dev)
 	state->image = NULL;
 }
 
+static int timer_check_and_set_def(struct tcmu_device *dev)
+{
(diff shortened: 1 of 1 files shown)


mikechristie:
In timer_check_and_set_def is the use of atof needed or should it be am integer call like atol or atoi, etc?

I don't have easy access to the ceph code or docs so I was not sure if it is possible to do a float/double.

lxbsz:
Sorry for late, yes from the ceph code we can see that the type of "rados_osd_op_timeout" is float, and i meant to s/atoi/atof, but just forgot to commit that change.

PR #266, opened 2017-08-23
Add -Werror for C_FLAGS

lxbsz:
This will make the code more save, because sometimes some warnings must be resolved or it will introduce bugs, like:

char *p;

if (cond)
    p=q;

foo (p) 
    .....


mikechristie:
Thanks.

PR #269, opened 2017-08-24
tcmu:fix return data length of rtpg response data

tangwenji:
When the actual length is greater than the allocation length，the return data length filed of the rtpg response data is incorrect。

### DIFF
--- a/alua.c
+++ b/alua.c
@@ -374,7 +374,7 @@ int tcmu_emulate_report_tgt_port_grps(struct tcmu_device *dev,
 		int next_off = off + 8 + (group->num_tgt_ports * 4);
 
 		if (next_off > alloc_len) {
-			ret_data_len += next_off;
+			ret_data_len += 8 + (group->num_tgt_ports * 4);
 			continue;
 		}
 


lxbsz:
Yes, it looks good to me.

BTW, from spc4r31 section 4.3.5.6 Allocation Length:
If the amount of information to be transferred exceeds the maximum value that the ALLOCATION LENGTH field is capable of specifying, shouldn't we terminate the command with CHECK CONDITION status, with the sense key set to ILLEGAL REQUEST, and the additional sense code set to INVALID FIELD IN CDB ?

tangwenji:
thanks for reviews.

From spc4r36q section 6.37 REPORT TARGET PORT GROUPS command
The RETURN DATA LENGTH field indicates the length in bytes of the list of target port group descriptors. The contents of the RETURN DATA LENGTH field are not altered based on the allocation length.
The actual transferred length is not exceeds the allocation length in rtpg response data,but the RETURN DATA LENGTH may exceed the allocation length. I think the initiator should resend rtpg command with longer alloc_length in this case。At least the code in multipath-tools is doing this。

lxbsz:
Nice!

mikechristie:
For this lxbsz's comment is also correct. If return data length is greater than what 32 bits (I'm assuming this is the max size because I do both have access to the spec) can handle then you should fail with an illegal request.

The part tangwenji quoted says if the allocated length is too small to fit the data the target should set the return data length to the value that would have been returned if the allocated length was large enough. This allows the initiator to get the correct size and then send a commmand with a buffer that is large enough.

If ret data length is greater than 32 bits can handle we would overflow. The initiator would read that value, send a rtpg with that len and then hit the same issue again and again.

mikechristie:
Oh yeah this is a separate issue and we have this bug in a lot of code. I think maybe we maybe only handle it in the mode sense code.

We need to come up with some nice helpers to copy data, check lens and set return data length correctly.

mikechristie:
One more oh yeah, I do not like the mode sense code even though it was my commit.

I think we can make it so we copy to the iovec and not have to use a temp buffer. Some commands needed to put a final length value at the beginning of the buffer/descriptor and some at the end.

tangwenji:
> If ret data length is greater than 32 bits can handle we would overflow. The initiator would read that value, send a rtpg with that len and then hit the same issue again and again.

I think the initiator will not be so endless cycle, under normal circumstances, the initiator will only re-issued once. I have seen multipath-tools before the code, for rtpg processing as follows:

> 
> rc = do_rtpg(fd, buf, buflen, timeout);
> 	if (rc < 0)
> 		goto out;
> 	scsi_buflen = (buf[0] << 24 | buf[1] << 16 | buf[2] << 8 | buf[3]) + 4;
> 	if (scsi_buflen > UINT_MAX)
> 		scsi_buflen = UINT_MAX;
> 	if (buflen < scsi_buflen) {
> 		free(buf);
> 		buf = (unsigned char *)malloc(scsi_buflen);
> 		if (!buf) {
> 			PRINT_DEBUG ("malloc failed: could not allocate"
> 				"%u bytes\n", scsi_buflen);
> 			return -RTPG_RTPG_FAILED;
> 		}
> 		buflen = scsi_buflen;
> 		memset(buf, 0, buflen);
> 		rc = do_rtpg(fd, buf, buflen, timeout);
> 		if (rc < 0)
> 			goto out;
> 	}
> 

But in the kernel scsi_dh_alua.c, there may indeed be repeated cases：

> 
>  retry:
> 	err = 0;
> 	retval = submit_rtpg(sdev, buff, bufflen, &sense_hdr, pg->flags);
> 
> 	if (retval) {
.......
> 	}
> 
> 	len = get_unaligned_be32(&buff[0]) + 4;
> 
> 	if (len > bufflen) {
> 		/* Resubmit with the correct length */
> 		kfree(buff);
> 		bufflen = len;
> 		buff = kmalloc(bufflen, GFP_KERNEL);
> 		if (!buff) {
> 			sdev_printk(KERN_WARNING, sdev,
> 				    "%s: kmalloc buffer failed\n",__func__);
> 			/* Temporary failure, bypass */
> 			pg->expiry = 0;
> 			return SCSI_DH_DEV_TEMP_BUSY;
> 		}
> 		goto retry;
> 	}
> 

I agree with mikechristie，If return data length is greater than what 32 bits can handle then you should fail with an illegal request.

PR #270, opened 2017-08-25
tcmu:double free tgt_port in tcmu_get_alua_grp

tangwenji:
The function tcmu_release_tgt_ports is called by function tcmu_free_alua_grp，so there is no need for explicit call tcmu_release_tgt_ports in tcmu_get_alua_grp()。


lxbsz:
Good catch and looks good to me.

PR #271, opened 2017-08-25
tcmu-runner:fix the exception return value for function tcmu_get_alua_grps()

tangwenji:
When the function tcmu_get_alua_grp() return NULL,the function tcmu_get_alua_grps should return exception value ,that can avoid the upper call function re-release grp list。


lxbsz:
Looks good to me.

mikechristie:
Thanks for the patches tangwenji and reviews lxbsz.

PR #275, opened 2017-08-29
alua:fix untrack when implicit transition not success

zhuozh:
when client login, alua_implicit_transition always return with lock state 2
this patch just fix it.


lxbsz:
Good catch and looks good to me.

mikechristie:
Thanks and thanks for the review lxbsz.

Issue #255, opened 2017-08-16
Kernel hung task when aio tracking gets asserted but kernel sent netlink cmd for remove and waiting on reply


bgly:
Aug 16 13:33:39 ip9-47-71-72 systemd[1]: Started Session 1645 of user gpfsadmin.
Aug 16 13:33:44 ip9-47-71-72 tcmu-runner[4131]: tcmu-runner: /tmp/jenkins/workspace/tcmu-runner/src/temp/tcmur_aio.c:179: cleanup_aio_tracking: Assertion `aio_track->tracked_aio_ops == 0' failed.
Aug 16 13:33:44 ip9-47-71-72 systemd[1]: tcmu-runner.service: Main process exited, code=killed, status=6/ABRT
Aug 16 13:33:44 ip9-47-71-72 systemd[1]: tcmu-runner.service: Unit entered failed state.
Aug 16 13:33:44 ip9-47-71-72 systemd[1]: tcmu-runner.service: Failed with result 'signal'.
Aug 16 13:33:55 ip9-47-71-72 kernel: [24632.521732] ABORT_TASK: Found referenced ibmvscsis task_tag: 7505882299573993664
Aug 16 13:33:55 ip9-47-71-72 rsyslogd-2007: action 'action 11' suspended, next retry is Wed Aug 16 13:35:25 2017 [v8.16.0 try http://www.rsyslog.com/e/2007 ]
Aug 16 13:33:58 ip9-47-71-72 systemd[1]: Started Session 1646 of user gpfsadmin.
Aug 16 13:34:17 ip9-47-71-72 systemd[1]: Started Session 1647 of user gpfsadmin.
Aug 16 13:34:32 ip9-47-71-72 systemd[1]: Started Session 1648 of user gpfsadmin.
Aug 16 13:34:49 ip9-47-71-72 kernel: [24686.845471] ABORT_TASK: Sending TMR_FUNCTION_COMPLETE for ref_tag: 7505882299573993664
Aug 16 13:34:51 ip9-47-71-72 systemd[1]: Started Session 1649 of user gpfsadmin.
Aug 16 13:35:01 ip9-47-71-72 CRON[124301]: (root) CMD (/usr/sbin/pvm-vios-health-check)
Aug 16 13:35:09 ip9-47-71-72 systemd[1]: Started Session 1651 of user gpfsadmin.
Aug 16 13:35:21 ip9-47-71-72 systemd[1]: Started Session 1652 of user gpfsadmin.
[...]

PR #276, opened 2017-08-30
Fix dev removed crash

zhuozh:


lxbsz:
@mikechristie 
And the kernel also needed to be fixed about the device removing.

This PR could resolve the tcmu-runner's crash, but the kernel's ring buffer broken case still exist.

tcmur_device.h
```diff
@@ -25,7 +25,7 @@
 
 #define TCMUR_DEV_FLAG_FORMATTING	(1 << 0)
 #define TCMUR_DEV_FLAG_IN_RECOVERY	(1 << 1)
-#define TCMUR_DEV_FLAG_SHUTTING_DOWN	(1 << 2)
+#define TCMUR_DEV_FLAG_SHUTTING_OR_REMOVING	(1 << 2)
 #define TCMUR_DEV_FLAG_IS_OPEN		(1 << 3)
```
mikechristie: How about just tcmur dev flag stopping

bgly:
@mikechristie Would this be the one that fixes our kernel crashing problem? Because all of ours occur on device remove. https://github.com/open-iscsi/tcmu-runner/issues/255

lxbsz:
@bgly Did your kernel crash log file has the ring broken logs?

The kernel has one bug: just before the remove nl sent to userspace, the ring has been released. But the userspace still doing IOs.

This PR couldn't fix/avoid the kernel's bug and could fix #255.

mikechristie:
I just sent @pkalever a kernel patch for a similar problem.

Is tcmu runner still doing IOs because the cmd_time_out fired and so the kernel freed the cmd and drop the reference to the device? The tcmu runner handler or runner itself still is executing it though?

Do you hit this problem with cmd_time_out = 0?

mikechristie:
I meant I sent him a patch to move the freeing of the ring to after the device is removed in user space.

We are seeing another problem with cmd time out disabled/0 and in that case IO should not be running. Are you by any chance also seeing kernel crashes in uio_poll or uio_read during dev removal?

lxbsz:
Kernel Crash case 2：

BUG: unable to handle kernel NULL pointer dereference at 0000000000000020
IP: tcmu_vma_fault+0xbe/0x1d0 [target_core_user]
PGD 202962f067 
P4D 202962f067 
PUD 202abc3067 
PMD 0 

Oops: 0000 [#1] SMP
CPU: 11 PID: 6639 Comm: tcmu-runner Not tainted 4.13.0-rc2+ #2Hardware name: Inspur NF5270M3/NF5270M3, BIOS CHEETAH_2.1.6 02/24/2014task: ffff9eafba7c1680 task.stack: ffffb0aa585dc000RIP: 0010:tcmu_vma_fault+0xbe/0x1d0 [target_core_user]
RSP: 0000:ffffb0aa585dfd68 EFLAGS: 00010246
RAX: 0000000000000000 RBX: ffffb0aa585dfde0 RCX: ffff9e9f80000108
RDX: 000000202c087000 RSI: 0000000000000000 RDI: 0000000000000000
RBP: ffffb0aa585dfda0 R08: 00003ffffffff000 R09: ffffc00000000fff
R10: 0000000000000000 R11: 0000000000000001 R12: ffff9eaf7dd00000
R13: 0000000000000000 R14: 0000000000000000 R15: ffff9ebfb4cfc5c0
FS:  00007f8606017800(0000) GS:ffff9ebfbf2c0000(0000) knlGS:0000000000000000
CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
CR2: 0000000000000020 CR3: 000000202a13c000 CR4: 00000000001406e0
Call Trace:
 __do_fault+0x1e/0x70
 __handle_mm_fault+0xab1/0x1120
 ? call_rwsem_wake+0x1b/0x30
 handle_mm_fault+0xd1/0x230
 __do_page_fault+0x22a/0x4a0
 ? __audit_syscall_exit+0x209/0x290
 do_page_fault+0x30/0x80
 ? do_syscall_64+0x147/0x150
 page_fault+0x28/0x30
RIP: 0033:0x7f8605c00ba2
RSP: 002b:00007ffca27e7780 EFLAGS: 00010213
RAX: 00007f8477800000 RBX: 0000000000000000 RCX: 00007f86045c1dca
RDX: 0000000001cd16d0 RSI: 0000000040800000 RDI: 0000000000000000
RBP: 00007ffca27e78f0 R08: 0000000000000032 R09: 0000000000000000
R10: 0000000000000001 R11: 0000000000000246 R12: 0000000000407b60
R13: 00007ffca27e7c90 R14: 0000000000000000 R15: 0000000000000000
Code: 24 40 93 00 00 44 89 e8 48 89 45 d0 48 89 f7 48 89 75 c8 48 89 c6 e8 b2 ba 21 c2 48 85 c0 49 89 c6 74 59 4c 89 ff e8 92 18 23 c2 <49> 8b 46 20 a8 01 75 43 4c 89 f0 f0 ff 40 1c 4c 89 73 48 48 83 
RIP: tcmu_vma_fault+0xbe/0x1d0 [target_core_user] RSP: ffffb0aa585dfd68
CR2: 0000000000000020

lxbsz:
@mikechristie 

The following kernel patch may fix the bugs above: https://www.spinics.net/lists/target-devel/msg15970.html

tcmur_device.h
zhuozh: stopping flag is good to me.

zhuozh:
@mikechristie This bug can be reproduced 100% by following steps below:
On initiator, do some fio test on your tcmu device,
On target, remove your device when device is IO busy. targetcli /backstores/user:rbd delete yourdevice
Runner will crash.
If you restart runner again after former crash, kernel will crash due to reference to NULL pointer, which one xiubo's kernel patch intended to fix.

lxbsz:
Tests okay and looks good to me.

bgly:
I am going to assume another case of this occurring is:

[23980.220457] tcmu daemon: command reply support 1.
[23980.387336] TARGET_CORE[ibmvscsis]: Detected NON_EXISTENT_LUN Access for 0x00000002
[23980.484651] Unable to handle kernel paging request for data at address 0x00000020
[23980.484662] Faulting instruction address: 0xd00000000e0b1e14
[23980.484666] Oops: Kernel access of bad area, sig: 11 [#1]
[23980.484668] SMP NR_CPUS=2048 NUMA pSeries
[23980.484674] Modules linked in: hvcs(OE) hvcserver rpadlpar_io rpaphp mmfs26(OE) mmfslinux(OE) tracedev(OE) dccp_diag dccp tcp_diag udp_diag inet_diag unix_diag af_packet_diag netlink_diag target_core_pscsi(OE) target_core_file(OE) target_core_iblock(OE) iscsi_target_mod(OE) vport_vxlan openvswitch target_core_user(OE) uio binfmt_misc pseries_rng ibmvmc(OE) vmx_crypto rtc_generic xt_conntrack nf_nat_ftp nf_conntrack_ftp nf_conntrack_netlink nfnetlink nf_conntrack_netbios_ns nf_conntrack_broadcast nf_conntrack_ipv6 nf_defrag_ipv6 nbd br_netfilter bridge ib_iser rdma_cm iw_cm ib_cm ib_sa ib_mad ib_core ib_addr vhost_net vhost macvtap macvlan iscsi_tcp libiscsi_tcp libiscsi scsi_transport_iscsi ip_vs libcrc32c iptable_nat nf_conntrack_ipv4 nf_defrag_ipv4 nf_nat_ipv4 iptable_mangle iptable_filter ipt_REJECT
[23980.484730]  nf_reject_ipv4 ipt_MASQUERADE nf_nat_masquerade_ipv4 nf_nat nf_conntrack ip_tables ip6table_filter ip6_tables ebtables x_tables dm_snapshot dm_bufio dm_multipath 8021q garp mrp stp llc autofs4 mlx4_en vxlan ip6_udp_tunnel udp_tunnel ses enclosure ibmvscsis(OE) target_core_mod(OE) configfs(OE) ibmveth(OE) lpfc mlx4_core ipr scsi_transport_fc
[23980.484757] CPU: 2 PID: 10360 Comm: tcmu-runner Tainted: G           OE   4.4.0-92-generic #115-Ubuntu
[23980.484760] task: c0000007ee2ae130 ti: c0000007ee35c000 task.ti: c0000007ee35c000
[23980.484763] NIP: d00000000e0b1e14 LR: d00000000e0b1e84 CTR: c0000000002854c0
[23980.484766] REGS: c0000007ee35f8f0 TRAP: 0300   Tainted: G           OE    (4.4.0-92-generic)
[23980.484769] MSR: 8000000000009033 <SF,EE,ME,IR,DR,RI,LE>  CR: 28004828  XER: 00000000
[23980.484777] CFAR: c000000000008468 DAR: 0000000000000020 DSISR: 40000000 SOFTE: 1
               GPR00: d00000000e0b1e84 c0000007ee35fb70 d00000000e0be360 0000000000000000
               GPR04: c0000007ee35fc30 0000000000000000 0000000000000054 0000000000000000
               GPR08: c0000007ee35fcd0 0000000000000000 c000000ff9180000 d00000000e0b5040
               GPR12: c0000000002854c0 c00000000fb41300 0000000000000000 0000000010010560
               GPR16: 0000000010010390 0000010035e538b0 0000000010030a58 c00000043ac9c400
               GPR20: c0000007f0661998 0000000000000000 c00000043ac9c000 0000000000000400
               GPR24: 0000000000000000 0000000000000000 0000000000001998 c0000007ee35fcd0
               GPR28: c0000007edff0000 c0000007ee35fc30 0000000000000054 0000000000000000
[23980.484816] NIP [d00000000e0b1e14] tcmu_vma_fault+0x104/0x290 [target_core_user]
[23980.484819] LR [d00000000e0b1e84] tcmu_vma_fault+0x174/0x290 [target_core_user]
[23980.484822] Call Trace:
[23980.484825] [c0000007ee35fb70] [d00000000e0b1e84] tcmu_vma_fault+0x174/0x290 [target_core_user] (unreliable)
[23980.484833] [c0000007ee35fc10] [c000000000270be4] __do_fault+0x84/0x160
[23980.484837] [c0000007ee35fcb0] [c000000000276818] handle_mm_fault+0xd78/0x1920
[23980.484842] [c0000007ee35fd80] [c000000000b0fe44] do_page_fault+0x354/0x7f0
[23980.484846] [c0000007ee35fe30] [c000000000008664] handle_page_fault+0x10/0x30
[23980.484848] Instruction dump:
[23980.484850] e8410018 3b7b91e8 7bd90020 7f24cb78 7f63db78 48003181 e8410018 7c7f1b79
[23980.484856] 418200ac 7f43d378 480027bd e8410018 <e93f0020> 792a07e1 40820074 393f001c
[23980.484865] ---[ end trace 59a40520dd1b4690 ]---

Or is this one addressed via the kernel patch from Xiubo? 
https://www.spinics.net/lists/target-devel/msg15970.html

Its hard for me to determine the cause of this crash since it was buried under automated tests that just do scale testing deploys and removes over and over. Something could've caused tcmu to crash while state was BUSY but I'm not 100% sure

mikechristie:
> @mikechristie This bug can be reproduced 100% by following steps below:
On initiator, do some fio test on your tcmu device,
On target, remove your device when device is IO busy. targetcli /backstores/user:rbd delete yourdevice
Runner will crash.

What is the cmd_time_out value being used?

The root question is why is IO still running in tcmu-runner when the targetcli delete is done, because every IO gets a reference to the device. You cannot remove a device while IO is running in runner unless commands have timed out. If commands have not timed out then we have another bug.

Or, is there a slight miscommunication? When you write that IO is busy, do you know if when the delete command is run IO is actually in the LIO and/or runner layer? Is there a bug where IO is not actually in those layers (IO is being sent by the initiator but IO running in the lio and runner layers had been completed so the refcount went to zero and we were able to delete the device), but the runner cmdproc thread is still accessing the ring buffer and doing uio read/write/poll calls when it shouldn't?

mikechristie:
Let's summarize the bugs:

1. If cmd_time_out != 0, and a command has timed out then we definitely need these patches, and the kernel patch that moves the ring buffer code to fix the page fault crashes here.

2. Why do we need to change where uio_unregister_device is being called in this kernel patch:
https://www.spinics.net/lists/target-devel/msg15970.html
Did you do it because uio_write was in the trace in kernel crash 1?

If cmd_time_out 0 or > 0, then we should not be making uio calls after tcmu_netlink_event has been called successfully. In tcmu runner we will do dev_removed -> tcmulib_cleanup_cmdproc_thread -> cancel_thread  and on successful completion that should have stopped the thread doing uio calls and ring accesses. Is this not correct? Is there a bug in this code?

If runner crashes then we could be in a inconsistent state, and I do not know what will happen. If you start runner again we need more code to detect what is left open.

3. If cmd_time_out = 0, or if it is > 0 but no commands timedout, then why do we need this patches? The kernel should not be calling tcmu_destroy_device until all outstanding commands have completed. Is that a correct assumption? If not, then we need more fixes. For example target_free_device needs to move the destroy_workqueue to after the destroy_device call. It would also seem that all the backends are buggy and will free memory for outstanding commands.

lxbsz:
@mikechristie 
Sorry for late.

    What is the cmd_time_out value being used?

It's the default value 30s.

    The root question is why is IO still running in tcmu-runner when the targetcli delete is done, because
    every IO gets a reference to the device. You cannot remove a device while IO is running in runner
    unless commands have timed out. If commands have not timed out then we have another bug.

    Or, is there a slight miscommunication? When you write that IO is busy, do you know if when the
    delete command is run IO is actually in the LIO and/or runner layer? Is there a bug where IO is not
    actually in those layers (IO is being sent by the initiator but IO running in the lio and runner layers had
    been completed so the refcount went to zero and we were able to delete the device), but the runner
    cmdproc thread is still accessing the ring buffer and doing uio read/write/poll calls when it shouldn't?

As we know, the udev's kref count will be equal to 2( open device + configure device).

Our crashes are caused by the following case:
1, IOs are running
2, try to remove the target device through the configfs
3, then the TCMU will release the cmd area and other memories in tcmu_destroy_device()
4, BUT NOW the page fault or write maybe running and will use the memories above.
5, then the TCMU will send a REMOVE netlink message to the runner in tcmu_destroy_device() and the kref count will be 1.
6, the runner receive the REMOVE nl message then try to remove the device and crash due to the tracker.
7, ....

If there hasn't any IOs when doing all these above, everything will works well and the step 7 will close the uio dev and then kref count will be zero.

And the following patch will release all the possible memories and resources only when the last user has kref_put the count to 0:
https://www.spinics.net/lists/target-devel/msg15970.html

For the kernel fix patch's I just copy it here:

See the following case if without :

-    uio_unregister_device(&udev->uio_info);
-

There should still be a bug:
1, IOs are running
2, try to remove the target device through the configfs
3, then the TCMU will send a REMOVE netlink message to the runner in tcmu_destroy_device().
4, since there maybe lots of IOs and the runner will wait for them being done, which will still use the UIO device.
5, the TCMU doesn't know what the runner's status and just called uio_unregister_device() to release the UIO device .
6, ....

And I am not very sure will the above case will cause the kernel's another crash ?

lxbsz:
If runner crashes then we could be in a inconsistent state, and I do not know what will happen. If you start runner again we need more code to detect what is left open.

Currently, if the runner crashes, in kernel the CMD area and other resources has been released and some IOs are still running.

Though the uio_unregister_device() has already called, but when the runner still using the UIO device, will it success immediately? or just returns and wait for the user release it asynchronously? but when runner crash the wait reply message lost. I'm not very sure for now, but it works like this.

When we start the runner again, the runner will access the CMD area and then the kernel crashes.

mikechristie:
For the page fault case:

Are you doing the sync or async tcmu removal calls? I am assuming sync.

I am not sure what IO running has to do with it.

In your cases you are saying that IO is running in runner when tcmu (target_core_user module) sends the remove nl event right? My question is how is that possible? Here is what I am saying is supposed to happen:

1. you are doing some IO test. IOs are in the iscsi layer, LIO core, tcmu, and tcmu-runner and rbd.c.
2. you do the targetcli removal command.
3. targetcli/rtslib first unlinks the tcmu device from the lun. This calls target_fabric_port_unlink() in the kernel which does:

core_dev_del_lun ->  core_tpg_remove_lun -> transport_clear_lun_ref

In that function we will wait for all outstanding IOs and will make it so new IOs cannot be sent to tcmu.

4. When IOs have completed, this will return and the LUN will be unlinked/unmapped.

5. targetcli/rtslib will then remove the tcmu device which will call tcmu_destroy_device and do the NL remove call. At this time, if no commands have timed out, then runner should be sitting in the poll call.


So I am sort of agreeing with you:

0. In general is a bug for tcmu to release the memory user space is using before we have removed the user space device.

1. if cmd_time_out has run and left commands running in runner, then the above code does not work since the refcount is dropped, so we can hit what you describe.

Did commands timeout and were freed by the kernel but left running in runner? If so, then the bug you hit makes sense.

2. If commands have not timed out then, how are we hitting your bug? Does this happen:

dev_removed -> tcmulib_cleanup_cmdproc_thread -> cancel_thread -> pthread_cancel

this then wakes tcmur_cmdproc_thread. Does poll then return that there are events and so we call  tcmulib_processing_start/tcmulib_get_next_command an extra time and we can hit your bug that way too? In this case, it would explain why we are seeing something similar with cmd_time_out=0.

In this case is poll supposed to return success and revents=POLLIN or is that a bug in the poll code? If that is how poll is supposed to work then I agree we need your patch for this case too.

mikechristie:
For the uio case:

For this again, I have been assuming that you are using the sync nl interface.

In this case, targetcli/rtslib and kernel should work the same. If cmd_time_out has fired and left commands running in runner, I do see how the patch in this PR where we wait for running to stop processing IO helps.

However it is not clear how the uio_unregister_device move helps in the normal case.

For this:

1, IOs are running
2, try to remove the target device through the configfs
3, then the TCMU will send a REMOVE netlink message to the runner in tcmu_destroy_device().
4, since there maybe lots of IOs and the runner will wait for them being done, which will still use the UIO device.
5, the TCMU doesn't know what the runner's status and just called uio_unregister_device() to release the UIO device .

we should be doing 


remove_device -> dev_removed -> tcmulib_cleanup_cmdproc_thread -> cancel_thread 

We then wait for the cmd thread to stop. When that is done we know it is not accessing the uio device.

remove_device then does the final close() on the uio device and then we return back to the kernel which is waiting for our nl reply. We can then do the uio_unregister_device where we do today because user space has is no longer touching the UIO device.

If you are hitting a bug in the uio code, then is it something else? Is tcmur_cmdproc_thread still running? Did cancel_thread not cleanup properly?

For the non normal case, if runner crashes then I think I can see how we need your patch.

mikechristie:
Oh yeah sync removal means using the newer tcmu runner and tcmu code where we send the NL event for removal/addition, then the kernel waits for a response.

lxbsz:
@mikechristie 

Yes, you are right.

It is doing the sync tcmu removal calls, the following is the kernel debug log:

Target_Core_ConfigFS: Releasing ALUA Target Port Group: alua/tg_pt_gps/tpg0, ID: 1
Allocated UNIT ATTENTION, mapped LUN: 0, ASC: 0x3f, ASCQ: 0x03
iSCSI_TPG[1]_LUN[0] - Removed ACL for InitiatorNode: iqn.2017-03.org.ceph:10.142.40.221 Mapped LUN: 0
iSCSI_TPG[1] - Freeing ACL for iSCSI InitiatorNode: iqn.2017-03.org.ceph:10.142.40.221 Mapped LUN: 0
iSCSI_TPG[1]_LUN[0] - Deactivating iSCSI Logical Unit from device object
sleeping for nl reply
close
genl cmd done got id 121 curr 2 done 2 rc 0
CORE_HBA[121] - Detached HBA from Generic Target Core
Started NOPIN Response Timer on CID: 0 to 5 seconds

For the kernel patch, the following code removal is no need:
\-    uio_unregister_device(&udev->uio_info);
\-

lxbsz:
In this case, targetcli/rtslib and kernel should work the same. If cmd_time_out has fired and left
commands running in runner, I do see how the patch in this PR where we wait for running to stop
processing IO helps.

In kernel for the cmd_time_out, in our environment it hasn't been fired and the timer will be deleted in tcmu_destroy_device(). Then the kernel just frees all the CMD area and some other memories.

And then just before the REMOVE nl message has been sent out, maybe the tcmu-runner is doing the uio write or triggering the tcmu page fault then kernel will crash.

Without this PR, the runner maybe crash and the nl reply will lost ? So the kernel will wait it without has any chance to call uio_unregister_device(), and for this case the kernel will crash when we try to run the runner again.

lxbsz:
@mikechristie 
Sorry for missing your old patch about this:
https://www.spinics.net/lists/target-devel/msg15947.html

zhuozh:
@mikechristie 
I'd like to add more additional explanation that this crash only happened when our ceph cluster is overwhelmed
After expend our ceph cluster global size from 200G to 32T, this bug became hard to reproduce. And I setup a smaller ceph cluster that only 30G for analysing purpose, and create several 30G RBD blocks make it filled with fio test, this time the bug appeared again.

mikechristie:
@zhuozh and @lxbsz are you guys hitting this code in tcmu_check_expired_cmd:
```

static int tcmu_check_expired_cmd(int id, void *p, void *data)
{
        struct tcmu_cmd *cmd = p;

        if (test_bit(TCMU_CMD_BIT_EXPIRED, &cmd->flags))
                return 0;

        if (!time_after(jiffies, cmd->deadline))
                return 0;

        set_bit(TCMU_CMD_BIT_EXPIRED, &cmd->flags);
        target_complete_cmd(cmd->se_cmd, SAM_STAT_CHECK_CONDITION);
        cmd->se_cmd = NULL;

        return 0;
```

If a commands has timedout meaning we passed the timer_after check and called target_complete_cmd then we can hit the bug you are describing. In that case the lun refcounts could be zero but tcmu-runner is still executing the command, and so we have to have the code in this PR.

We could hit this timeout more easily when the cluster is overworked.

Maybe just add a printk in that function to make sure that is the case and we are not hitting some other type of race.

lxbsz:
@mikechristie
Yes, we are tring to add one log msg there and other places, but didn't have chance to do the test yet and will test it today.

Yesterday we were tring to reproduce the bugs again and again because our test environment has changed then the.bug disappeared.

We can make sure that, the runner has several log about cmd time out from the ceph because having no space.

lxbsz:
@mikechristie 
We just reproduce this bug again:
Follows the following reproduce steps will 100% to trigger the bug:
1, setup one ceph cluster which has 30G hole size.
2, create 10 rbd blocks and each size is 30G.
3, use dd to write block0 about about 28G.
4, use fio to write block1 about 100G.
5, if the ceph cluster is full and the ceph cluster will be HEALTH_ERR(from 'ceph -s') and the runner will try to recovery the blockX
6, then try to delete the block1 using targetcli and  the targetcli will stuck.
7, remove the rbd block0 in ceph cluster to make the ceph cluster become okay.
8, sometimes we can see the runner crash and then the kernel crash too.
9, if the runner and kernel doesn't crash, just restart the tcmu-runner again then the kernel should crash.
10, then it's about the page fault crash logs.

We have add one log msg in timedout function, but didn't find any of the logs when kernel crash:
    diff --git a/drivers/target/target_core_user.c b/drivers/target/target_core_user.c
    index 942d0942..e358668 100644
    --- a/drivers/target/target_core_user.c
    +++ b/drivers/target/target_core_user.c
    @@ -1037,6 +1037,9 @@ static int tcmu_check_expired_cmd(int id, void *p, void *data)
            if (!time_after(jiffies, cmd->deadline))
                    return 0;
     
    +       pr_err("cmd %d is expired for about %u msec\n",
    +                id, jiffies_to_msecs(cmd->deadline - jiffies));
    +
            set_bit(TCMU_CMD_BIT_EXPIRED, &cmd->flags);
            target_complete_cmd(cmd->se_cmd, SAM_STAT_CHECK_CONDITION);
            cmd->se_cmd = NULL;
    @@ -1597,6 +1600,7 @@ static void tcmu_destroy_device(struct se_device *dev)
            idr_destroy(&udev->commands);
            spin_unlock_irq(&udev->commands_lock);
            WARN_ON(!all_expired);
    +       printk("all_expired = %d\n", all_expired);
     
            tcmu_blocks_release(udev);

zhuozh:
@mikechristie 
My test is slightly different with Xiubo's procedure, and the cmd_time_out in kernel fired this time.
1, setup one ceph cluster which has 30G whole size.
2, create 10 rbd blocks and each size is 30G.
3, use dd to write block0 about 28G.
**4, use fio randrw block1 about 100G.**
5, run "watch ceph df" and 'watch ceph -s', the moment ceph cluster become HEALTH_ERR and ceph pool is 100%USED, try to delete the block1 using targetcli, recovery thread is not triggered, but runner crash immediately.
`2017-09-06 14:35:12.288 3694 [DEBUG] handle_netlink:210 : cmd 2. Got header version 2. Supported 2.
2017-09-06 14:35:12.288 3694 [DEBUG] tcmu_cancel_recovery:135 block4: Waiting on recovery thread
2017-09-06 14:35:12.288 3694 [DEBUG] cmdproc_thread_cleanup:624 block4: cmdproc cleanup done
tcmu-runner: /data/tcmu-runner/tcmur_aio.c:226: cleanup_aio_tracking: Assertion `aio_track->tracked_aio_ops == 0' failed.
Aborted (core dumped)`
With Xiubo's printk msg added in timedout function, we get the cmd_time_out expired as expected:
`<3>[  914.910977] cmd 0 is expired for about 4294967052 msec
<3>[  914.913304] cmd 1 is expired for about 4294967052 msec
<3>[  914.915561] cmd 2 is expired for about 4294967052 msec
<3>[  914.917786] cmd 3 is expired for about 4294967052 msec
<3>[  914.919954] cmd 4 is expired for about 4294967052 msec
<3>[  914.922088] cmd 5 is expired for about 4294967042 msec
<3>[  914.924184] cmd 6 is expired for about 4294967042 msec
<3>[  914.926225] cmd 7 is expired for about 4294967042 msec
<3>[  914.928221] cmd 8 is expired for about 4294967036 msec
<3>[  914.930179] cmd 9 is expired for about 4294967036 msec
<4>[  914.933069] all_expired = 1`
7, then restart runner, we'll get the page fault crash logs.

zhuozh:
My test is slightly different with Xiubo's procedure, and the cmd_time_out in kernel fired this time.
1, setup one ceph cluster which has 30G whole size.
2, create 10 rbd blocks and each size is 30G.
3, use dd to write block0 about 28G.
**4, use fio randrw block1 about 100G.**
5, run "watch ceph df" and 'watch ceph -s', the moment ceph cluster become HEALTH_ERR and ceph pool is 100%USED, try to delete the block1 using targetcli, recovery thread is not triggered, but runner crash immediately.
```
2017-09-06 14:35:12.288 3694 [DEBUG] handle_netlink:210 : cmd 2. Got header version 2. Supported 2.
2017-09-06 14:35:12.288 3694 [DEBUG] tcmu_cancel_recovery:135 block4: Waiting on recovery thread
2017-09-06 14:35:12.288 3694 [DEBUG] cmdproc_thread_cleanup:624 block4: cmdproc cleanup done
tcmu-runner: /data/tcmu-runner/tcmur_aio.c:226: cleanup_aio_tracking: Assertion `aio_track->tracked_aio_ops == 0' failed.
Aborted (core dumped)
```
With Xiubo's printk msg added in timedout function, we get the cmd_time_out expired as expected:
```
<3>[  914.910977] cmd 0 is expired for about 4294967052 msec
<3>[  914.913304] cmd 1 is expired for about 4294967052 msec
<3>[  914.915561] cmd 2 is expired for about 4294967052 msec
<3>[  914.917786] cmd 3 is expired for about 4294967052 msec
<3>[  914.919954] cmd 4 is expired for about 4294967052 msec
<3>[  914.922088] cmd 5 is expired for about 4294967042 msec
<3>[  914.924184] cmd 6 is expired for about 4294967042 msec
<3>[  914.926225] cmd 7 is expired for about 4294967042 msec
<3>[  914.928221] cmd 8 is expired for about 4294967036 msec
<3>[  914.930179] cmd 9 is expired for about 4294967036 msec
<4>[  914.933069] all_expired = 1
```
7, then restart runner, we'll get the page fault crash logs.

mikechristie:
Ok.

1.
I think @zhuozh is hitting the bug I mentioned where commands are still running in runner due to kernel timing them out, and I agree these patches and the memory related parts of https://www.spinics.net/lists/target-devel/msg15970.html will fix that problem.

@bgly, if you are also seeing timeouts then this would fix your bug in the aio tracking code.

2.
There is another uio crash that we are debugging here (gluster and rbd teams at Red Hat), and in that case it does not require timeouts to have occurred. I wonder if you guys are also hitting that too.

I think your kernel patch will fix the problem, because it changes the refcounts, but I still think we should not need it.  We do not have an explanation for why when cancel_thread has returned successfully the last ref to the uio device has not been done. I am guessing we are just not understanding something about thread stoppage and if it is in the middle of sys calls at the time of the cancel.

3.
@lxsbz I do not know what you are hitting if you are not seeing time outs. I do not think the shutdown code is correct, because for #2, we are hitting uio creates with no IO running in runner. I am going to assume you are hitting this too and just merge these patches, and we can continue to debug if we see it still.

For the kernel patch please handle the review comments and resubmit.

main.c
```diff
@@ -810,6 +827,9 @@ static void dev_removed(struct tcmu_device *dev)
 	tcmulib_cleanup_cmdproc_thread(dev);
 
 	cleanup_io_work_queue(dev, false);
+
+	if (aio_wait_for_empty_queue(rdev))
```
mikechristie: I think we want to move this before tcmulib_cleanup_cmdproc_thread.

If runner is in somewhere in tcmur_generic_handle_cmd (after a track_aio_request_start call, but before the cmd has been sent to the handler), we might cancel the thread before the runner thread calls track_aio_request_finish or before the handler can setup its AIO which would eventually call track_aio_request_finish. We would then wait here forever.

main.c
```diff
@@ -685,6 +688,16 @@ static void *tcmur_cmdproc_thread(void *arg)
 			tcmu_err("ppoll received unexpected revent: 0x%x\n", pfd.revents);
 			break;
 		}
+
+		/*
+		 * If a device being removed is dealing with contiguous I/O requests,
+		 * we need to throttle cmdproc thread to make sure it have a chance
+		 * to become idle
```
mikechristie: Could you also change this comment to be:

LIO will wait for outstanding requests and prevent new ones from being sent to runner during device removal, but if the tcmu cmd_time_out has fired tcmu-runner may still be executing requests that LIO has completed. We only need to wait for replies for outstanding requests so exit from the cmdproc now.

main.c
```diff
@@ -685,6 +688,16 @@ static void *tcmur_cmdproc_thread(void *arg)
...
+		 */
+		pthread_mutex_lock(&rdev->state_lock);
+		if (rdev->flags & TCMUR_DEV_FLAG_STOPPING)
+			dev_stopping = true;
```
mikechristie: I think we can just kill dev_stopping and break from the loop here.

main.c
```diff
@@ -685,6 +688,16 @@ static void *tcmur_cmdproc_thread(void *arg)
...
+		pthread_mutex_unlock(&rdev->state_lock);
 	}
 
 	tcmu_err("thread terminating, should never happen\n");
```
mikechristie: We can now hit this path right, so we should remove this.

main.c
mikechristie: Ignore this.

main.c
mikechristie: It looks like we can now hit the following:

We go down this code path and call pthread_cleanup_pop -> cmdproc_thread_cleanup -> rhandler->close

while commands are still being executed by the handler. Maybe we want the call to track_aio_request_finish in cmdproc_thread_cleanup instead of below in dev_removed.

lxbsz:
@mikechristie 
The kernel patch has been updated, for my test case i will try it again if we can see the crash still in future.

Thanks.

main.c
zhuozh: Agreed.

main.c
zhuozh: More precise and detailed. I will update it in next upate.

main.c
zhuozh: In consideration of that we already have the cmd_time_out fired in kernel to make sure outgoing IOs be handled, there is no need to drain the cmd ring buffer, right? 
add track_aio_request_finish logic in cmdproc_thread_cleanup make things kind of complicated, we need to distinguish whether handler has really take over outgoing IO.

zhuozh:
@mikechristie 
For the cmd_time_out fired runner crash case, If these on-the-fly requests in runner lose it's connection with handler, it will trigger recovery thread, and TCMUR_DEV_FLAG_IN_RECOVERY flag will be set which prevents cmdproc thread exit and stuck it on tcmu_cancel_recovery waiting for handler recovery.

This is another bug, we can't catch it becasue it hides behind tracked_aio_ops assert bug.
```
void tcmu_cancel_recovery(struct tcmu_device *dev)
{
        struct tcmur_device *rdev = tcmu_get_daemon_dev_private(dev);

        /*
         * Only file and qcow can be canceled in their open/close calls, but
         * they do not support recovery, so wait here for rbd/glfs type of
         * handlers to fail/complete normally to avoid a segfault.
         */
        tcmu_dev_dbg(dev, "Waiting on recovery thread\n");
        pthread_mutex_lock(&rdev->state_lock);
        while (rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY) {
                sleep(1);     
                pthread_mutex_lock(&rdev->state_lock);
        }
}
```
after add the following code:
```
@@ -813,6 +832,9 @@ static void dev_removed(struct tcmu_device *dev)
         * are getting invoked when shutting down the handler.
         */
        cleanup_io_work_queue_threads(dev);
+
+       if (aio_wait_for_empty_queue(rdev))
+               tcmu_dev_err(dev, "could not flush queue.\n");
        tcmulib_cleanup_cmdproc_thread(dev);
 
        cleanup_io_work_queue(dev, false);
```
 I got the following log below, runner hang, targetcli delete stuck on waiting for nl reply.
2017-09-07 14:43:02.149 4722 [DEBUG] handle_netlink:210 : cmd 2. Got header version 2. Supported 2.
2017-09-07 14:43:02.149 4722 [DEBUG] aio_wait_for_empty_queue:102 block4: waiting for 10 commands
2017-09-07 14:43:02.489 4722 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block4: Timing out cmd.
2017-09-07 14:43:02.489 4722 [ERROR] tcmu_notify_conn_lost:163 block4: Handler connection lost (lock state 1)
(-----------stuck here--------)

when device is removed and it's lifecycle ends, recovery is not needed, we can fix it by checking TCMUR_DEV_FLAG_STOPPING in tcmu_cancel_recovery, so I update this PR for reviewing purpose.

main.c
mikechristie: > @mikechristie In consideration of that we already have the cmd_time_out fired in kernel to make sure outgoing IOs be handled, there is no need to drain the cmd ring buffer, right?

For device removal, there is no need.

Later, when we fix up support for stopping and restarting tcmu-runner while devices are still present, then it could be more complicated. Would we want to just let the kernel cmd_time_out fire or loop over the ring buffer and fail commands or execute the commands? We can handle that later, so we can fix this crash now.


> add track_aio_request_finish logic in cmdproc_thread_cleanup make things kind of complicated, we need to distinguish whether handler has really take over outgoing IO.

Ok. Sounds good. Do the simple option now.

mikechristie:
I can fix that issue. Working on it now.

tcmur_device.c
```diff
@@ -156,7 +156,7 @@ void tcmu_notify_conn_lost(struct tcmu_device *dev)
 	struct tcmur_device *rdev = tcmu_get_daemon_dev_private(dev);
 
 	pthread_mutex_lock(&rdev->state_lock);
-	if (rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY)
+	if ((rdev->flags & TCMUR_DEV_FLAG_STOPPING) || (rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY))
 		goto unlock;
```
mikechristie: There are 2 checks for STOPPING in __tcmu_reopen_dev which is called a little later by the thread started by this function. Why are you not hitting those checks?

Without this patch, do you get stuck in a handler->open/close call for extended periods (minutes) and it is looking like a hang?

mikechristie:
@lxbsz for the uio read/poll related crashes you saw above I think I found the problem here:

https://github.com/open-iscsi/tcmu-runner/pull/290

tcmur_device.c
zhuozh: @mikechristie 
Test again without this patch, and this log reveals more details:
1. cmd timeout arrives
2. nl msg arrives
3. tcmu_add_dev_to_recovery_list failed to start recovery thread due to  enabled_port = 0
and returned with ret=0, which will cause TCMUR_DEV_FLAG_IN_RECOVERY wrongly set. 
('enabled_port = 0' is wired because I created my device with alua_access_type=1, need your advise here.) whatever TCMUR_DEV_FLAG_IN_RECOVERY wrongly set is a bug when enabled_port = 0.

# tcmu_add_dev_to_recovery_list()
{
...
        if (!enabled_port) {
                /* User disabled port from under us? */
                tcmu_dev_err(dev, "Could not find enabled port.\n");
                goto done;
        }  
...
done:
        tcmu_release_alua_grps(&alua_list);
        pthread_mutex_unlock(&tpg_recovery_lock);
        return ret;
}

# tcmu_notify_conn_lost()
{
...
        if (!tcmu_add_dev_to_recovery_list(dev))
                rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
...
}

2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_notify_conn_lost:166 block0: Handler connection lost (lock state 1)
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_add_dev_to_recovery_list:327 block0: Could not find enabled port.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.334 5466 [ERROR] tcmu_rbd_handle_timedout_cmd:655 block0: Timing out cmd.
2017-09-08 10:01:06.335 5466 [DEBUG] handle_netlink:210 : cmd 2. Got header version 2. Supported 2.
2017-09-08 10:01:06.335 5466 [DEBUG] tcmu_cancel_recovery:135 block0: Waiting on recovery thread

zhuozh: STOPPING checking at the beginning of tcmu_notify_conn_lost is needed, because code path
tcmu_rbd_handle_timedout_cmd -> tcmu_notify_conn_lost -> tcmu_add_dev_to_recovery_list -> tcmu_get_alua_grps will access the configfs dir resource which is holded by kernel in configfs_remove.
If tcmu_rbd_handle_timedout_cmd is blocked, in tcmulib_cmd->done() will never get a chance to clear tracked_aio_ops, which will cause deadlock in the following code path:

        cleanup_io_work_queue_threads(dev);

        if (aio_wait_for_empty_queue(rdev))
                tcmu_dev_err(dev, "could not flush queue.\n");
        tcmulib_cleanup_cmdproc_thread(dev);

        cleanup_io_work_queue(dev, false);
        cleanup_aio_tracking(rdev);

lxbsz: @mikechristie @zhuozh 

With this new fix I couldn't reproduce the stuck any more on my environment.

mikechristie: @zhuozh Nice debugging!

mikechristie: For the final patch could you break up the check so it is in the 80 col limit we have been using?
if ((rdev->flags & TCMUR_DEV_FLAG_STOPPING) ||
    (rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY))

and add a comment about the issue, so new devs that do not do kernel work will know that we cannot change the behavior due to the kernel and not anything runner is doing.

main.c
mikechristie: For this code path were you going to drop the code where we can exit from the loop so we have a single place that can call cmdproc_thread_cleanup?

So we do not call cmdproc_thread_cleanup from 

tcmur_cmdproc_thread->pthread_cleanup_pop->cmdproc_thread_cleanup

and we were only going to call cmdproc_thread_cleanup from 

dev_removed->tcmulib_cleanup_cmdproc_thread->cancel_thread->pthread_cancel->cmdproc_thread_cleanup 

right?

tcmur_device.c
zhuozh: @mikechristie Thanks mike, PR updated by following your comments.

main.c
mikechristie: I saw you repushed your patches. Were you going to handle this too? Right now wen can do tcmur_cmdproc_thread->pthread_cleanup_pop->cmdproc_thread_cleanup->rhandler->close which could free memory that the handler is using to execute io. I think we can just do the attached patch.

[only-execute-cmdproc-cleanup-once.txt](https://github.com/open-iscsi/tcmu-runner/files/1296877/only-execute-cmdproc-cleanup-once.txt)

zhuozh: @mikechristie Sorry for missing your comments here. @lxbsz Thanks for your reminding:-) Yes, you are right about cmdproc_thread_cleanup, repush this PR again with your patch merged.

mikechristie:
Thanks for all the work put into this @zhuozh and @lxbsz!

main.c
```diff
@@ -690,12 +693,21 @@ static void *tcmur_cmdproc_thread(void *arg)
...
+		 * requests that LIO has completed. We only need to wait for replies
+		 * for outstanding requests so throttle the cmdproc thread now.
+		 */
+		pthread_mutex_lock(&rdev->state_lock);
+		if (rdev->flags & TCMUR_DEV_FLAG_STOPPING)
+			dev_stopping = true;
+		pthread_mutex_unlock(&rdev->state_lock);
+	}
 
+	pthread_cleanup_pop(0);
```
mikechristie: Shoot. This was wrong.

I think we can now hit the race where this completes first, we just do pop(0) and tcmur_cmdproc_thread never gets run leaving the device opened. Will make a new PR for this.

PR #274, opened 2017-08-29
Logdir misc fix

lxbsz:
1, add tcmu_logdir_check helper support 
2, fix and clean up the file output callout 
3, make tcmu_log_dir as static
4, move tcmu_logdir_create to log file and make it globally 
5,  fix error when making recursion directories 
6,  fix tcmu_conf_set_options 
7, make log dir configurable in tcmu.conf

PR #281, opened 2017-09-01
tcmur: fixed possible uninitialized variable warning

dillaman:


tcmur_cmd_handler.c
```diff
@@ -277,7 +277,8 @@ static int align_and_split_unmap(struct tcmu_device *dev,
 	uint8_t *sense = origcmd->sense_buf;
 	uint64_t opt_unmap_gran;
 	uint64_t unmap_gran_align, mask;
-	int ret, j = 0;
+	int ret = TCMU_NOT_HANDLED;
+	int j = 0;
 	struct unmap_descriptor *desc;
```
mikechristie: @lxbsz  I think we need the attached patch along with this. Basically the above case should never happen with the attached patch and the above code is just to handle the compile error.


[unmap-handle-zero.txt](https://github.com/open-iscsi/tcmu-runner/files/1271209/unmap-handle-zero.txt)

dillaman: @mikechristie assuming that was a comment to me, applied your patch. Thanks.

file_optical.c
```diff
@@ -233,6 +233,7 @@ static int fbo_emulate_request_sense(struct tcmu_device *dev, uint8_t *cdb,
 {
```
mikechristie: For this patch we will do the linux kernel'ish style of get/put be* in this PR:

https://github.com/open-iscsi/tcmu-runner/pull/282

libtcmu_log.c
```diff
@@ -506,16 +506,16 @@ int tcmu_setup_log(void)
 	if (ret < 0)
 		tcmu_err("create stdout output error \n");
```
mikechristie: This will be fixed in:

https://github.com/open-iscsi/tcmu-runner/pull/274/commits/10686538cbc7bfd557c2fd53f395bca596b164f4

tcmur_cmd_handler.c
lxbsz: Yes, looks good to me.

dillaman:
@mikechristie dropped the two commits handled by other PRs

mikechristie:
Thanks for the patches dillaman and review lxbsz.

PR #283, opened 2017-09-02
rbd: try to remain cluster's osd op timeout default value

lxbsz:
If the cluster's default osd op timeout already larger than grace + interval, will remain the old.


mikechristie:
Thanks. I hand merged this

https://github.com/open-iscsi/tcmu-runner/commit/cfd8a35c2c7bbdfe414ee4445613a48b41a6e2c2

and fixed some of the language in the comments when I merged it.

PR #284, opened 2017-09-02
libtcmu: handle error when netlink family group is not found

nak3:
genl_ctrl_resolve_grp returns a negative error code[1] when failed to
look up the family object. This patch handles the negative error code.

[1] http://www.infradead.org/~tgr/libnl/doc/api/group__genl__ctrl.html
>    Numeric group identifier or a negative error code.


mikechristie:
Huh. That is a weird. I wonder what happened. Did we just forget to handle the ret value or did we do it on purpose due to some issue in the netlink code.

Going to assume we just forgot to handle ret.

Thanks.

PR #285, opened 2017-09-04
open_handlers: fix dlsym return

pkalever:
from 'man dlsym':
[...]
the correct way to test for an error is to call dlerror(3) to clear any
old error conditions, then call dlsym(), and then call dlerror(3) again,
saving its return value into a variable, and check whether this saved
value is not NULL.


mikechristie:
Thanks.

PR #286, opened 2017-09-04
libtcmu: move genl_unregister_family to teardown_netlink

pkalever:


mikechristie:
Weird. I wonder how it got stuck there in the first place. Thanks.

PR #288, opened 2017-09-06
rbd: use semicolon to separate options from the pool/image spec

dillaman:


dillaman:
Related tooling change: https://github.com/ceph/ceph-iscsi-config/pull/17

dillaman:
... guess colon was a bad choice: https://github.com/open-iscsi/rtslib-fb/blob/master/rtslib/tcm.py#L826

mikechristie:
Adding myself a note:

I think ":" is not allowed because the LIO configfs info file (/sys/kernel/config/target/core/user_N/dev_name/info) file exports info with the format:

setting: value

A colon in the config string value might throw off parsing somewhere like in rtslib _parse_info.

Will do some more research and bug Andy, because it looks like we can just fix up rtslib as the kernel code will be ok.

dillaman:
@mikechristie Turns out that comma is not allowed either -- everything after the comma is dropped:

> $ echo "dev_config=rbd/rbd/lun0,test=1" > control
> $ cat info
> Status: ACTIVATED  Max Queue Depth: 0  SectorSize: 512  HwMaxSectors: 128
>         Config: rbd/rbd/lun0 Size: 1073741824
> $ echo "dev_config=rbd/rbd/lun0:test=1" > control 
> $ cat info
> Status: ACTIVATED  Max Queue Depth: 0  SectorSize: 512  HwMaxSectors: 128
>         Config: rbd/rbd/lun0:test=1 Size: 1073741824
>

lxbsz:
And the '=' will be dropped by configshell_fb too, and there is one fix PR for it: https://github.com/open-iscsi/configshell-fb/pull/36, after this the targercli could work for rbd/bock/osd_op_timeout=30 now.

dillaman:
Switching to semicolon since it isn't parsed out by the various tooling

PR #289, opened 2017-09-06
tcmu: implement TEST_UNIT_READY 

pkalever:
* The first patch was just some code doc updates, typo fix and indentation fixes in the code doc.
* The later one for implementing TEST UNIT READY command.

### DIFF
--- a/api.c
+++ b/api.c
@@ -30,6 +30,7 @@
 #include <endian.h>
 #include <errno.h>
 
+#include "libtcmu.h"
 #include "libtcmu_log.h"
 #include "libtcmu_common.h"
 #include "libtcmu_priv.h"
@@ -757,11 +758,22 @@ int tcmu_emulate_inquiry(
 }
 
 int tcmu_emulate_test_unit_ready(
+	struct tcmu_device *dev,
 	uint8_t *cdb,
 	struct iovec *iovec,
 	size_t iov_cnt,
 	uint8_t *sense)
 {
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+
+	if (!rhandler->check_stat)
+		return SAM_STAT_GOOD;
+
+	if (!rhandler->check_stat(dev)) {
+		tcmu_err("check_config failed for %s\n", dev->dev_name);
+		return SAM_STAT_CHECK_CONDITION;
+	}
+
 	return SAM_STAT_GOOD;
 }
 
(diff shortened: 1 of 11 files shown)


api.c
```diff
@@ -757,11 +758,22 @@ int tcmu_emulate_inquiry(
...
 	size_t iov_cnt,
 	uint8_t *sense)
 {
+	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+
+	if (!rhandler->check_stat)
+		return SAM_STAT_GOOD;
+
+	if (!rhandler->check_stat(dev)) {
+		tcmu_err("check_config failed for %s\n", dev->dev_name);
```
mikechristie: Just do a tcmu_dev_err(dev, "blah");

api.c
```diff
@@ -757,11 +758,22 @@ int tcmu_emulate_inquiry(
...
+		return SAM_STAT_CHECK_CONDITION;
```
mikechristie: What do you need TUR support for?

Is this for linux multiapth-tools path testing? For linux multipath support, we can return anything, and it eventually retries, so you would just need to change this to return sense properly:

return tcmu_set_sense_data(sense, .... ASC_LOGICAL_UNIT_COMMUNICATION_FAILURE);

For some other OSs I think you might hit issues though.

If this runner instance cannot reach the gluster network and we returned that error then it might indicate to the OS the device is gone when only that instance's network connection is bad. The OS might then mark the entire device as gone/dead when we only wanted it to mark the path to this instance as down and to switch paths.

We could:

1. For issues where the handler cannot connect to its backend storage we have the tcmu_notify_conn_lost function. This basically, disables the iscsi port, and then does a reopen of the device. When it can reopen it, we assume the connection is ok and enable the iscsi port.

We could have check_stat return error codes indicating if the problem is a device error or transport error. For device we return LUNN COMM error and for transport we do the tcmu_notify_conn_lost call for the handler.

I think the problem is that conn_lost is tied to ALUA support right now and I do not think you guys are setting that up right?

2. For now, we can just return LUNN COMM error when check_stat files and just leave it to the handler to call conn_lost if it can.

mikechristie: I am ok with either one for now, so do either one and repush the patch.

mikechristie: @pranithk @pkalever @lxbsz 

Hey,

Above is basically the same thing I was talking about in how to handle READ CAP dynamically and not forcing initiator side device level failures when just one target node cannot reach the gluster cluster.

PR #290, opened 2017-09-08
libtcmu: fix name check for nl commands

mikechristie:
If dev->dev_name is uio1 but dev_name is uio10 we would
match it and return the dev for uio1.

This would cause problems for device removal because we would
leave the cmdproc running and accessing the uio device for
the device we wanted to shutdown (we shutdown the cmdproc
for uio1 instead of uio10). The kernel would then do
uio_unregister_device and free the uio device, but runner
could still be doing uio read/poll calls and cause crashes.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -136,12 +136,9 @@ lookup_dev_by_name(struct tcmulib_context *ctx, char *dev_name, int *index)
 	*index = 0;
 
 	darray_foreach(dev_ptr, ctx->devices) {
-		size_t len;
-
 		dev = *dev_ptr;
-		len = strnlen(dev->dev_name, sizeof(dev->dev_name));
 
-		if (!strncmp(dev->dev_name, dev_name, len)) {
+		if (!strcmp(dev->dev_name, dev_name)) {
 			*index = i;
 			return dev;
 		}


mikechristie:
One other note.

As far as I can tell, we do not need to be so defensive in lookup_dev_by_name because the callers all make sure there is a terminating null in the names being passed in.

lxbsz:
@mikechristie 
Yeah, good catch. I also think this is the root cause of the uio write crash problem.

And the snprintf will add one '\0' at the end to make sure the dev_name to be null terminated.

This look nice for me.

pkalever:
Awesome catch @mikechristie :-)

I read this code like at least 5-10 times, but every time I missed the point, rather thought about other approaches [1] & [2].

Let me test this if it solves uio_read crash that we have seen, will update you once I'm done with the it. 

[1] https://github.com/open-iscsi/tcmu-runner/pull/213/commits/20dda83880891666e241039fbdd2bf218daaae85
[2] https://github.com/open-iscsi/tcmu-runner/pull/218

pkalever:
Have tested this by making '200 block creates and then 200 block deletes' in a loop. Executed this for 5 times. And I do not see any uio_read crashes in my steup.

This patch is all set to get merged IMO.

Thanks Mike.

mikechristie:
Thanks for the review lxbsz and pkalever.

humblec:
Nice catch @mikechristie ! I believe this will fix many crashes in uio_{read,write}

PR #291, opened 2017-09-09
make sure only negative error code is returned

zhuozh:
just a small code style fix.


lxbsz:
This will also fix bugs like : 

        ret = tcmu_rbd_service_register(dev);                                                                                                                                                  
        if (ret < 0)
                goto rados_shutdown;

lxbsz:
Looks good to me.

mikechristie:
Nice. Thanks.

PR #295, opened 2017-09-12
log_internal: append device subtype if available

pkalever:
example:
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1061 glfs/abc: 35 0 0 0 0 0 0 0 0 0
[...]
[DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1061 rbd/xyz: 12 0 0 0 24 0
[...]

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -161,6 +161,7 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	unsigned int head;
 	char *msg;
 	int n;
+	struct tcmur_handler *rhandler;
 
 	if (pri > tcmu_log_level)
 		return;
@@ -179,8 +180,16 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	head = logbuf->head;
 	rb_set_pri(logbuf, head, pri);
 	msg = rb_get_msg(logbuf, head);
-	n = sprintf(msg, "%s:%d %s: ", funcname, linenr,
-		    dev ? dev->tcm_dev_name: "");
+
+	if (dev) {
+		rhandler = tcmu_get_runner_handler(dev);
+		n = sprintf(msg, "%s:%d %s/%s: ", funcname, linenr,
+		            rhandler ? rhandler->subtype: "",
+		            dev ? dev->tcm_dev_name: "");
+	} else {
+		n = sprintf(msg, "%s:%d: ", funcname, linenr);
+	}
+
 	vsnprintf(msg + n, LOG_MSG_LEN - n, fmt, args);
 
 	rb_update_head(logbuf);


lxbsz:
Yeah, this looks very nice and tests okay.

lxbsz:
2017-09-12 17:27:54.628 11269 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1199 rbd/block0: a3 a 0 0 0 0 0 0 0 cc 0 0 
    2017-09-12 17:28:04.633 11269 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1199 rbd/block0: 0 0 0 0 0 0 
    2017-09-12 17:28:04.634 11269 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1199 rbd/block0: a3 a 0 0 0 0 0 0 0 cc 0 0 
    2017-09-12 17:28:24.643 11269 [DEBUG_SCSI_CMD] tcmu_cdb_debug_info:1199 rbd/block0: 0 0 0 0 0 0

pkalever:
@lxbsz thanks!

libtcmu.c
```diff
@@ -704,7 +704,7 @@ int tcmulib_master_fd_ready(struct tcmulib_context *ctx)
```
mikechristie: Is this fixing a bug you are hitting?

It seems in the cases you are returning points to structs you only moved the problem. For example tcmu_get_runner_handler will not crash now, but it returns a NULL pointer and I think every call point will just reference that.

For the places you return a numerical value like tcmu_get_dev_num_lbas we just have another weird bug where incorrect values are being used.

Is there any place below where we can possibly have a null device or handler, etc or in every case would it be a bug?

mikechristie: > Is there any place below where we can possibly have a null device or handler, etc or in every case would it be a bug?

Nevermind. I see you added one in the next patch :)

libtcmu_log.c
```diff
@@ -179,8 +180,13 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	head = logbuf->head;
 	rb_set_pri(logbuf, head, pri);
 	msg = rb_get_msg(logbuf, head);
-	n = sprintf(msg, "%s:%d %s: ", funcname, linenr,
-		    dev ? dev->tcm_dev_name: "");
+	if (dev || rhandler) {
```
mikechristie: I do not like the other patch, because it will just hide real bugs. I think here you just need to code the checks.

```

If (dev) {
    handler = tcmu_get_dev_handler();

    .....

```

Note you should use tcmu_get_dev_handler because some of the other callers are from comsumer and synth which are not runner based.

mikechristie: Nevermind on the dev_handler comment. I am guessing the ltcmulib_handler->name is too verbose.

I guess since we are saying we do not care about the libtcmu abstraction anymore, do:
```

If (dev) {
    rhandler = tcmu_get_runner_handler();

```

pkalever:
@mikechristie updated the PR as per the review comments. Thanks.

mikechristie:
Thanks.

PR #296, opened 2017-09-15
systemd: define glusterd precedence

pkalever:
Added "After=glusterd.service",
This will not have any effect on the way tcmu-runner is started today, this can
be leveraged by any layered services (example gluster-blockd) who starts
glusterd/target/tcmu-runner, to help define the order in which they should get
started.

In case if glusterd is not started, this change:
* will not start glusterd
* will still start tcmu-runner

### DIFF
--- a/tcmu-runner.service
+++ b/tcmu-runner.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=LIO Userspace-passthrough daemon
+After=glusterd.service
 
 [Service]
 LimitNOFILE=1000000


tcmu-runner.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=LIO Userspace-passthrough daemon
+After=glusterd.service
```
dillaman: Does this affect deployments that don't have a glusterd.service? Perhaps the glusterd.service should just have a ```Before=tcmu-runner```?

mikechristie: @pkalever I saw you comment on the other systemd PR. Just wanted to make sure you saw Jason's comment. I am just waiting for a reply.

We added Wants/Afters in the ceph services that use tcmu-runner and it works for us. Our design might be different though, so let us know.

pkalever: @mikechristie @dillaman sorry, looks like I have missed this one.

@dillaman nope, IMO this will not affect deployments which doesn't have gluster. Because "After=" will not start any unit. This just helps in sorting the dependencies chain if a parent unit starts both tcmu-runner and glusterd. 

For me it is more meaning full to add storage units as dependencies to tcmu-runner in the tcmu-runner unit because ceph and gluster are reasonably a layer below tcmu-runner. Because tcmu-runner needs storage before itself gets up, it is important that we defend on the precedence in the runner unit. More context at https://github.com/open-iscsi/tcmu-runner/pull/301

lxbsz: @mikechristie @dillaman 

I have test the "After=glusterd.service" at #301 case, and without gluster.service locally, it works fine as @pkalever mentioned above.

pkalever: @lxbsz Thanks!

mikechristie: @pkalever what do you guys use for the target initialization? Do you use the target.service to start lio? To setup targets and map luns to gluster block devs, do you rely on target.service and it reading the /etc/target info or is that done by something in the glusterd.service?

pkalever: @mikechristie We rely on target.servic.

Here is the sequence that we follow.
It is a Prerequisite that glusterd.service should be running.

On a start of gluster-blockd.service (distributed block management framework) gluster-blockd will trigger gluster-block-target.service (wrapper over target.service) which intern trigger tcmu-runner and start itself after tcmu-runner comes up. Once tcmu-runner gluster-block-target comes up gluster-blockd will get started.

We have Requisite=glusterd.service (intentional) in gluster-block-target.service and gluster-blockd.service, which means just check if glusterd is up, in case it is up we go further else we will fail the entire transaction. Unlike Requires, Requisite will not trigger the listed service (glusterd).

mikechristie: Just to make sure I am on the same page you do not actually need this change, because the gluster unit files have the necessary After/Before settings. You just want this patch to be defensive.

https://github.com/gluster/gluster-block/blob/master/systemd/gluster-blockd.service.in

has

After=gluster-block-target.service

so it will start after

https://github.com/gluster/gluster-block/blob/master/systemd/gluster-block-target.service.in

which has

After=glusterd.service tcmu-runner.service

so when gluster-block-target and gluster-blockd are up, glusterd and tcmu-runner are going to be up.

nixpanic:
On Tue, Sep 26, 2017 at 07:03:28PM +0000, mikechristie wrote:
... snip!
> so when gluster-block-target and gluster-blockd are up, glusterd and tcmu-runner are going to be up.

But glusterd also needs to be up before tcmu-runner starts (and tries to
connect to glusterd for its glfs backend).

I think this change is not defensive, but rather crucial for forcing the
correct order of starting the services.

mikechristie:
> But glusterd also needs to be up before tcmu-runner starts (and tries to
connect to glusterd for its glfs backend).

What I am not understanding is that I thought it is for the cases where you can have gluster devs setup in runner.

If you use targetcli directly to setup targets and map gluster block devices to luns then I get glusterd has to be up and something needs to make sure it is up. You guys do not support that though, so let's ignore it unless this is what you are talking about.

You use your gluster-block-target service over all this, and that starts glusterd and tcmu-runner before it starts up.

So what I am not understanding is when will tcmu-runner be up but glusterd will not be and there will be tcmu-runner/target luns mapped to gluster block devices. Is there a race type case where something can run the gluster block/iscsi setup tool before the gluster-block-target and gluster-block service is up. Is this for a tcmu-runner restart case (all gluster services stopped but the target one leaves running targets/luns), and then runner is restarted and it needs to reconnect to gluster? Also, why do you only need glusterd up and not the gluster-block service up?

mikechristie:
> You use your gluster-block-target service over all this, and that starts glusterd and tcmu-runner before it starts up.

Just to be clear, I am probably not understanding the problem because it my understanding that there will be no tcmu-runner calls that lead to gluster calls until after gluster-block-target service is up. If that service makes those calls then because it has the After= glusterd and tcmu-runner then both those will be up before gluster-block-target starts making targetcli/tcmu-runner calls.

Also to speed things up, because we are in different time zones, I am wrong then rebase your patch, so I can merge it when I get your reply and we do not have to wait a day.

mikechristie:
> Is there a race type case where something can run the gluster block/iscsi setup tool before the gluster-block-target and gluster-block service is up

Oh yeah, for this, I just have no idea how your tools work :) I was thinking the gluster block/iscsi tools talk to some "gluster daemon" running on each node that then makes targetcli/rtslib calls which makes tcmu-runner calls. Is that "gluser daemon" also part of the glusterd service or another daemon started by the gluster-block-target or gluster-blockd service?

seroyer:
I could be missing something, but starting tcmu-runner does not cause tcmu-runner to start accessing backstores unless those backstores are already in configfs.  The way we handle this (admittedly not using gluster), is that we don't specify ordering in tcmu-runner.service.  Instead we have another service that Wants=tcmu-runner.service but does the targetcli restore to repopulate configfs.  We put any ordering we need into that service (i.e., After=remote-fs.target)

PR #297, opened 2017-09-15
[RFC]rbd: fix the rados osd op timeout setting with cmd_time_out

lxbsz:
The rados osd op timeout must be longer than the timeouts to detect
unreachable OSDs (osd heartbeat grace + osd heartbeat interval) or
we will end up failing the transport connection when we just needed
to try a different OSD.

And also the cmd time out in ring buffer in kernel must be longer
than the rados osd op timeout too, then we can make sure that before
the cmd in the osd timed out and get the reply from the osd, the cmd
in ring buffer is still alive.

Or for the multipath case, the client may resend the same cmd just
before the osd really timedout.


rbd.c
```diff
@@ -215,39 +252,39 @@ static int timer_check_and_set_def(struct tcmu_device *dev)
 	tcmu_dev_dbg(dev, "The cluster's default osd op timeout(%f), osd heartbeat grace(%d) interval(%d)\n",
 		     timeout, grace, interval);
 
+	cmd_time_out = tcmu_get_dev_cmd_time_out(dev);
+
 	/* Frist: Try to use new osd op timeout value */
-	if (state->osd_op_timeout && atof(state->osd_op_timeout) > grace + interval)
+	if (state->osd_op_timeout && atof(state->osd_op_timeout) > grace + interval) {
+		/* Make sure osd_op_timeout is larger than cmd_time_out */
+		if (atoi(state->osd_op_timeout) <= cmd_time_out) {
```
mikechristie: If cmd_time_out is turned off (0) then this will increase the osd timeout when it shouldn't, so you need an extra "cmd_time_out > 0 && other test".

Are you guys setting cmd_time_out to catch tcmu-runner hangs/crashes? What is the initiator behavior when this happens? It seems like with multipathing you are going to get odd behavior where runner is down, tcmu will timeout the command and send the initiator a response. That will allow the initiator and multipath layer to complete its error handling if it fired and end up retrying this same path. Since runner is down, more IO will be sent and it will timeout and possibly kick off the eh again and we repeat over and over. You then have to restart runner and possibly the box because runner's restart code is buggy.

If runner is down due to crash or bug with cmd_time_out=0 then IO is stuck, it will time out on the initiator side and it will end up marking the path as bad. We will not have the ping ponging. You will still have to reboot, becuase the tcmu-runner/tcmu code is buggy for restart in this case.

When we fix the runner restart code, then with cmd_time_out=0 you will not have the ping ponging between paths.

lxbsz: If cmd_time_out is turned off (0) then this will increase the osd timeout when it shouldn't, so
    you need an extra "cmd_time_out > 0 && other test".
Yes, this should be fixed.

lxbsz: @mikechristie
I think I have made a mistake, the osd_op_timeout should be < cmd_time_out for this case.

lxbsz: Our test case and environment are:
1, Ceph cluster with the size 30GB totailly and with creating 2 rbd devices(16G for each)
2, Only one tcmu node with two net interface: 10.XX.XX.222 and 126.XX.XX.222, will map two LUNs for each target devices.
3, The cmd_time_out is 3s, the osd op timeout is 30s.

Firstly write one rbd device with random data almost full, then try writing random data to another rbd device using multipath. The ceph cluster will be stuck when just the second target will be almost full, then remove the first rbd device from the cluster and the cluster will recovery.

There still has enough empty area in the 1G Data area of the ring buffer for the rest of IOs, so even the cluster is stuck the initiator could send all the rest of IOs cmds successfully.

if cmd_time_out < osd op timeout, when the old cmd has timed out in ring buffer and the initiator may retry the other path when it must. While at the same time the cluster recovers and the new one cmd comes, does there has problems in ceph cluster with two same IO operations at the same time ?

I think there also has the same problems with two different tcmu nodes use the same produce steps here.

mikechristie:
lxbsz,

I think the patches look ok. I am traveling and cannot test with the ceph-iscsi-* tools right now. When I get home Wed, I will test and merge.

Thanks.

main.c
```diff
@@ -705,6 +705,16 @@ static int dev_added(struct tcmu_device *dev)
 	tcmu_dev_dbg(dev, "Got block_size %ld, size in bytes %lld\n",
 		     block_size, dev_size);
 
+	cmd_time_out = tcmu_get_attribute(dev, "cmd_time_out");
+	if (cmd_time_out < 0) {
```
mikechristie: Older kernels maybe not have cmd_time_out. If we want to support them, then you need to add a check for -EOPNOTSUPP.

If we do not want to support them, then we should log a more descriptive error message about the device addition failing because the kernel does not support cmd_time_out and also probably advise what kernel they need to upgrade to.

rbd.c
```diff
@@ -181,32 +181,88 @@ static void tcmu_rbd_image_close(struct tcmu_device *dev)
 	state->image = NULL;
 }
 
+static int update_cmd_time_out(struct tcmu_device *dev,
+					uint32_t timeout)
+{
+	int ret;
+
+	ret = tcmu_set_attribute(dev, "cmd_time_out", timeout);
```
mikechristie: Same here as above about handling old kernels. Add support or add a descriptive message so we do not have to handle user questions.

main.c
mikechristie: Oh yeah, I think for this, we want to add the -EOPNOTSUPP check because glfs just turns it off in their tools if it exists and does not care about changing it.

lxbsz: Yes, right.
But I think we should check -ENOENT, because if cmd_time_out is not exist, the open will fail with the errno = ENOENT.

lxbsz: @mikechristie @pranithk 
BYW, the glfs tools won't just write zero to cmd_time_out to disable it if it exists ?

mikechristie:
How are you guys setting up tcmu-runner? Do you use targetcli and are you saving the target config in /etc/target/saveconfig.json or doing your own tools?

This patch does not seem to work with the restore commands because cmd_time_out gets stored in the attributes field in saveconfig.json and that gets set after the initial device setup.

When you run targetcli restoreconfig or start the target.service, targetcli/rtslib will do the device creation command which passes the osd_op_timeout value stored in the saveconfig.json config field  to tcmu-runner, and so your patch works as expected and will update the cmd_time_out value. The problem is that the saveconfig.json's cmd_time_out value has not yet been set, so rbd.c only tested against and updated against the kernel default of 30 seconds. Once the device is created rtslib will set the attributes in that dictionary. This will then overwrite the value your rbd.c code just set.


Here is an example from saveconfig.json:

>       ],
>       "attributes": {
>         "cmd_time_out": 3,
>         "dev_size": 11811160064,
>       },
>       "config": "rbd/rbd/gwcli0;osd_op_timeout=10",
>       "hw_max_sectors": 128,
>       "name": "gwcli0",
>       "plugin": "user",
>       "size": 11811160064,
>       "wwn": "163c422a-4fbf-426e-9757-123456789861"
>     }
>

mikechristie:
Maybe we need to:

1. In rtslib tcm.py UserBackedStorageObject _configure() make it so we set the attributes first?
2. Do a kernel patch to make cmd_time_out a setting settable with tcmu_set_configfs_dev_params, modify _configure to pass this in before doing self.enable(), and then I think we would have to then make the configfs cmd_time_out attr read-only to avoid that code resettting it.
3. Drop the code to try and adjust the settings and just document how to set them. Just keep checks in rbd.c and log a message if the settings are going to be bad and/or just return failure at device creation time.
4. ?

lxbsz:
@mikechristie 
As you mentioned above.

How about the following changes:

Firstly, do a kernel patch to make  cmd_time_out a setting settable with tcmu_set_configfs_dev_params as the default value, which could be overlapped by the tcmu_cmd_time_out_store configfs later if this is called.

Then in rtslib/root.py restore(), if exist try to append the  

    "attributes": {
            "cmd_time_out": 3,
      },
to

    "config": "rbd/rbd/gwcli0;osd_op_timeout=10",

like:

    "config": "rbd/rbd/gwcli0;osd_op_timeout=10;cmd_time_out=3",

mikechristie:
Sorry @lxbsz I missed your reply.

With the deferred_enable proposal here:

https://github.com/open-iscsi/targetcli-fb/pull/95

we can make rtslib/targetcli/ceph-iscsi-* set the control and attributes before doing the enable, so tcmu-runner has all the settings set at device setup time.

lxbsz:
@mikechristie 

Yes, I will add this.

Thanks,

lxbsz:
Do a kernel patch to make cmd_time_out a setting settable with tcmu_set_configfs_dev_params,
    modify _configure to pass this in before doing self.enable(), and then I think we would have to then
    make the configfs cmd_time_out attr read-only to avoid that code resettting it.

@mikechristie About the read-only problem.
How about just let the cmd_time_out attr to be the following format:
 
    "cmd_time_out=30[L]"   // the 'L' means Lock

If It's locked, the next time when reconfiguring it, the value must be larger than the current, or will be fail and keep the old.

This will one patch in kernel to support the 'L'.

PR #301, opened 2017-09-20
systemd: setup the tcmu-runner in multi-user.target

lxbsz:
This will enable the tcmu-runner service booted when the system
is up.

More detail please see:
https://www.freedesktop.org/software/systemd/man/bootup.html#

### DIFF
--- a/tcmu-runner.service
+++ b/tcmu-runner.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=LIO Userspace-passthrough daemon
+After=network.target
 
 [Service]
 LimitNOFILE=1000000
@@ -7,3 +8,6 @@ Type=dbus
 BusName=org.kernel.TCMUService1
 KillMode=process
 ExecStart=/usr/bin/tcmu-runner
+
+[Install]
+WantedBy=multi-user.target


tcmu-runner.service
```diff
@@ -7,3 +7,6 @@ Type=dbus
 BusName=org.kernel.TCMUService1
 KillMode=process
 ExecStart=/usr/bin/tcmu-runner
+
+[Install]
+WantedBy=multi-user.target
```
pkalever: Why do we wanted to start tcmu-runner before even storage comes up ?
IMHO, stores like rbd, glfs need to come up first, so that on targetctl restore, tcmu-runner will get storage ready.

lxbsz: @pkalever 

For our case, we'd like the tcmu node could recover from system crash or powered down/up atomically. And this [Install] section will only works if you 'systemctl enable tcmu-runner'.

pkalever:
@lxbsz I got the intention behind you adding 'multi-user.target'. But what I'm trying to highlight is, the precedence of missing pieces here or the chain.

Say, we had some blocks on the node, and for some reason the node crashed followed by a reboot. If 'multi-user.target' starts tcmu-runner at its run level, before even network comes up, and even storage comes up, it is definitely a problem.

Refer: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sect-Managing_Services_with_systemd-Targets.html
> Systemd targets are represented by target units. Target units end with the .target file extension and their only purpose is to group together other systemd units through a chain of dependencies. For example, the graphical.target unit, which is used to start a graphical session, starts system services such as the GNOME Display Manager (gdm.service) or Accounts Service (accounts-daemon.service) and also activates the multi-user.target unit. Similarly, the multi-user.target unit starts other essential system services such as NetworkManager (NetworkManager.service) or D-Bus (dbus.service) and activates another target unit named basic.target.

So I fell, if we are taking this patch in, we should also add
After=network.target glusterd.service $(ceph_service)

For me it is more meaning full to add storage units as dependencies to tcmu-runner in the tcmu-runner unit because ceph and gluster are reasonably a layer below tcmu-runner. Because tcmu-runner needs storage before itself gets up, it is important that we defend on the precedence in the runner unit.

lxbsz:
@pkalever 
Yes, you are right. I have fix and test this and your #296 PR.

pkalever:
@lxbsz Thanks!

mikechristie:
@bgly 

For file_optical and qcow do you need 

After=sys-kernel-config.mount local-fs.target

in the systemd runner unit file or is that guaranteed to be done before multiusr.arget? I thought in one doc it said it would be, but the lio target.service has them.

PR #302, opened 2017-09-21
zbc: Cleanup zone related definitions

damien-lemoal:
Remove the local copy of blkzoned.h and integrate needed definitions
directly in file_zbc.c. While at it, cleanup access to the zone
descriptor structure fields using helper macros.

Signed-off-by: Damien Le Moal <damien.lemoal@wdc.com>


mikechristie:
Thanks for fixing that.

PR #307, opened 2017-09-25
runner: add a spec file support

lxbsz:
Just add a spec file support to make it easy to build with rpmbuild
or with SUSE's open build service.

### DIFF
--- a/tcmu-runner.spec
+++ b/tcmu-runner.spec
@@ -1,0 +1,75 @@
+Name:          tcmu-runner
+Summary:       A daemon that handles the userspace side of the LIO TCM-User backstore
+Group:         System Environment/Kernel
+License:       Apache 2.0
+Version:       1.3.0
+URL:           https://github.com/open-iscsi/tcmu-runner
+
+%define _RC rc4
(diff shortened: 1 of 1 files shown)


mikechristie:
Could you incorporate the changes in the suse and fedora specs:

https://build.opensuse.org/package/view_file/openSUSE:Factory/tcmu-runner/tcmu-runner.spec?expand=1

http://pkgs.fedoraproject.org/cgit/rpms/tcmu-runner.git/tree/tcmu-runner.spec

For example

1. Don't hard code the install dirs. Use those rpm macors.
2. Build gluster and ceph since we included their deps in the build requires.

Do not copy the distro libtcmu-devel parts.
1. We want to move away from supporting a stable libtcmu for now, so do not do a -devel package, so it is clear that you cannot build against the lib. Maybe we just want to statically build against libtcmu and not install it at all?

lxbsz:
Yes, will update later.
Thanks,

lxbsz:
@mikechristie 
Has updated, please review.

For now didn't contain the libtcmu and -devel parts, this will be added in the future.

Thanks,

mikechristie:
Thanks.

PR #308, opened 2017-09-26
fix compiling error for gcc 5.3.1.

pacepi:
error: format not a string literal and no format arguments [-Werror=format-security]

Signed-off-by: zhenwei.pi <zhenwei.pi@youruncloud.com>


mikechristie:
CCing zbc maintainer @damien-lemoal 

This patch looks ok to me.

damien-lemoal:
It looks good.

mikechristie:
Thanks.

PR #309, opened 2017-09-26
Add unit attention support for device resizes

mikechristie:
This patchset adds support for the capacity data changed UA. It is sent on successful device resize operations.

Notes:
- This patchset does not coordinate with the kernel to clear the UA queue when there have been resets in the kernel.
- I only added resize support. I did not do the other ones in PR 189 yet.


bgly:
Hi Mike, a quick question... Did you ever test whether or not the client gets a UA after the reconfig? I was adding some code to add media change for FBO under these changes and I do not see any UA's being handled via the handle_pending_ua code. I do see that it is realized in configfs and that targetcli has all the new values, but if a UA isn't sent then the client doesn't realize the change.

bgly:
@mikechristie

mikechristie:
Yeah, it works for me.

Does handle_pending_ua have any pending UAs setup (is rdev->pending_uas 0)?

In your function to handle the reconfigure event did you call tcmur_set_pending_ua() like how dev_resize does but with a new UA value you are adding?

bgly:
@mikechristie It was an error on my part, i forgot that fbo is a passthrough only device so it doesn't even go through the generic handle command, thus it never does the handling of any of this stuff.

PR #311, opened 2017-09-29
Fix ceph install deps to support [librbd1 librbd]-devel

lxbsz:
https://download.ceph.com/rpm-jewel/rhel7/x86_64/
Before and Jewel Version, should be librbd1-devel.

https://download.ceph.com/rpm-kraken/el7/x86_64/
After and Kraken Version, should be librbd-devel.


mikechristie:
We just can't get this right :) Thanks. I forgot it changes in the different versions.

PR #314, opened 2017-10-18
rbd:member variable bounce_buffer of struct rbd_aio_cb can not distinguish r/w io

tangwenji:
The bounce_buffer member variable also malloc buffer in tcmu_rbd_write。
So if ceph return write io error， but the target returned to the initiator of the ascq is ASC_READ_ERROR，and for wirte io, the processing is completed, do not need to continue copy buffer to iovec。

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -74,7 +74,7 @@ struct rbd_aio_cb {
 	struct tcmu_device *dev;
 	struct tcmulib_cmd *tcmulib_cmd;
 
-	/* Only for reading cmds */
+	bool read;
 	int64_t length;
 	char *bounce_buffer;
 };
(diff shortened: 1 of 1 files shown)


rbd.c
```diff
@@ -885,6 +889,7 @@ static int tcmu_rbd_flush(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 
 	aio_cb->dev = dev;
 	aio_cb->tcmulib_cmd = cmd;
+	aio_cb->read = false;
 	aio_cb->bounce_buffer = NULL;
```
mikechristie: tcmu_rbd_aio_writesame also needs to get modified right?

tangwenji: I modified it。Thanks！

mikechristie:
Thanks.

PR #316, opened 2017-10-18
rbd: include image id within service daemon registration

dillaman:
This is needed to allow an in-memory lookup of librbd performance
metrics on the Ceph dashboard without the need to query the Ceph
cluster or iterate through all available metrics.


mikechristie:
LGTM.

Did the moving of the tcmu_rbd_service_register(dev) call fix other bugs or just a code movement preference?

dillaman:
@mikechristie I had to move it so that the image was opened prior to the registration of the service (so that I could get its image id).

PR #317, opened 2017-10-18
rbd: new optional for specifying location of Ceph config file

dillaman:
A custom Ceph config file can now be specified via a
"conf=/path/to/ceph.conf" optional within the cfgstring.


mikechristie:
Thanks.

PR #318, opened 2017-10-19
tcmur: when tcmu_notify_lock_lost just warning it

lxbsz:
Warning it instead of output one error message, the error message
may confuse the users.


mikechristie:
Thanks.

PR #322, opened 2017-10-26
Struct iovec fix v2

gonzoleeman:
This pull request fixes an error where compilation of tcmu-runner on newer (4.13 on) kernels produces an error message complaining about "duplicate struct iovec". This is caused by more than one definition of "struct iovec" in linux/uio.h and struct_iovec.h in newer kernels, and may be addressed eventually. In the mean time, this fix eliminates the issue for tcmu-runner.

I tested on 4.13.8 kernel.


mikechristie:
This still does not work for me:

> [  3%] Building C object CMakeFiles/tcmu.dir/libtcmu.c.o
> In file included from /root/targetcli/tcmu-runner/libtcmu.c:36:0:
> /root/targetcli/tcmu-runner/target_core_user_local.h:113:17: error: array type has incomplete element type
>     struct iovec iov[0];
>   

....

I tried different combos of includes and could not get it to work with older setups and ones where glib has the struct_iovec.h. Let's just say forget niceness for now and break out the hacks. How would we do it? Something like check if struct_iovec.h exists and if not then include a local struct_iovec.h?

Also for the PR, there is the actual patch but then those 2 other things:

Merge pull request #1 from open-iscsi/master  …
f7bc516
Merge remote-tracking branch 'upstream/master'
53651cc
 
What happened with that? It seems harmless because there is nothing in there. If I had pulled it, would there just be entered for those 2 things but nothing in the actual pulls?

ls-zhu:
This issue still can be reproduced in both openSUSE and Fedora, and Lee's patch works for me, shall we merge it?

mikechristie:
ls-zhu, see the previous comment. We cannot merge it as is, because it breaks setups without the new iovec definition.

ls-zhu:
[build_tcmu_runner.log](https://github.com/open-iscsi/tcmu-runner/files/1676158/build_tcmu_runner.log)

ls-zhu:
Hello Mike,

I tried again, I can build tcmu-runner with Lee's patch. I run cmake and make  without any parameters. Do you have any special configurations? I uploaded the build log in my previous comment.
Would you please provide your "setup"? Then I can try again, maybe can have another solution.

gonzoleeman:
On Jan 29, 2018, at 6:30 PM, Zhu Lingshan <notifications@github.com> wrote:
> 
> Hello Mike,
> 
> I tried again, I can build tcmu-runner with Lee's patch. I run cmake and make without any parameters. Do you have any special configurations? I uploaded the build log in my previous comment.
> Would you please provide your "setup"? Then I can try again, maybe can have another solution.
> 
> 

Lingshan:

If you take the proposed fix and try to compile it on an older kernel and libc package (such as SLES 12 SP3), the compile fails with the message Mike showed earlier.

I worked for a couple of days on a fix that worked with both old and new libc include files to no avail. So I punted and figured I’d wait until everybody was on the new version.

The real fix, IMHO, is in the libc compatibility fix proposed upstream, but that may never be accepted.
— 
Lee

ls-zhu:
@gonzoleeman @mikechristie 
Oh, I never tried old kernels. In my own environment, I can apply the patch manually. Still trying to figure out how to fix it in this upstream code base. 
(I think it is easily to control the code in build farm for SLE / RHEL)

mikechristie:
@gonzoleeman 

Does the attached patch work for you? It works in older distros like RHEL 7 and newer ones with the uio header change like in fedora 27+ which I think fails similar to SUSE.

[fix-comp2.txt](https://github.com/open-iscsi/tcmu-runner/files/2133913/fix-comp2.txt)

gonzoleeman:
@mikechristie I will check. I have to set up my own build environment to do so, since our build service has my compatability patch in place, negating any testing value.

mikechristie:
I merged the fix-comp2.txt patch:

https://github.com/open-iscsi/tcmu-runner/commit/6b9dc381e231c93c61efe736698dab69e37d8adf

Open a new PR if there are issues.

PR #325, opened 2017-10-27
runner: zero default unmap granularity alignment

ddiss:
The unmap granularity alignment is advertised to initiators via the
Block Limits INQUIRY VPD, which is specified as follows (sbc3r25):

  The UNMAP GRANULARITY ALIGNMENT field indicates the LBA of the first
  logical block to which the OPTIMAL UNMAP GRANULARITY field applies.
  The unmap granularity alignment is used to calculate an optimal unmap
  request starting LBA as follows:
  optimal unmap request starting LBA = (n × optimal unmap granularity)
                                       + unmap granularity alignment

Using a default unmap granularity alignment of max_xfer_length, which
is currently the case, means that the optimal unmap request starting LBA
is incorrectly staggered.

### DIFF
--- a/main.c
+++ b/main.c
@@ -797,12 +797,12 @@ static int dev_added(struct tcmu_device *dev)
 	rdev->flags |= TCMUR_DEV_FLAG_IS_OPEN;
 
 	/*
-	 * Set the optimal unmap granularity and the alignment to
-	 * max xfer len
+	 * Set the optimal unmap granularity to max xfer len. Optimal unmap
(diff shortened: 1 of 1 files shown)


mikechristie:
It looks ok to me.

@lxbsz do you remember why we did it the original way?

lxbsz:
Sorry for missing this.

It's long time ago, I couldn't get any useful messages from the history PRs about this code.

And this looks good to me too.

Thanks,

mikechristie:
Thanks.

PR #332, opened 2017-11-06
tcmu-runner service: increase default nproc limit

lxbsz:
It may hitting the nproc limit issue with hundreds of devices
running, likes: "Could not start implicit transition thread:Cannot
allocate memory". This patch just increases the default limit of
nproc.

@mikechristie 
We have produced this bug may times in our different environments:

    Nov 03 21:41:43 sds-3-94 tcmu-runner[2155]: alua_implicit_transition:461 rbd/rbd.block23: Could not start implicit transition thread:Cannot allocate memory
    Nov 03 21:41:43 sds-3-94 tcmu-runner[2155]: alua_implicit_transition:461 rbd/rbd.block151: Could not start implicit transition thread:Cannot allocate memory
    Nov 03 21:41:43 sds-3-94 tcmu-runner[2155]: check_lbas:106 rbd/rbd.block55: cmd exceeds last lba 20971520 (lba 30077, xfer len 31427203)

This patch just set the limitation to 1000000.

### DIFF
--- a/tcmu-runner.service
+++ b/tcmu-runner.service
@@ -4,6 +4,7 @@ After=network.target
 
 [Service]
 LimitNOFILE=1000000
+LimitNPROC=1000000
 Type=dbus
 BusName=org.kernel.TCMUService1
 KillMode=process


pkalever:
From https://www.freedesktop.org/software/systemd/man/systemd.exec.html, it looks like "LimitNPROC" is equivalent to 'ulimit -u'

From docs of ulimit
-u  ==>   The maximum number of processes available to a single user

But as I understand, we need to increase thread count per process, we might not be interested in tuning processes per user ?

$ sysctl -n kernel.threads-max
$ cat /proc/sys/kernel/threads-max

mikechristie:
@lxbsz are you hitting the nproc limit because the pthread_create does a clone which in the kernel does a fork which ends up calling copy_process then copy_creds and that increments the nproc count so later we hit that check?

lxbsz:
@pkalever As I know, the thread count per process is limited not only by  /proc/sys/kernel/threads-max, this doesn't work for me.

There has another limitation, which we could increase this by increasing vm.max_map_count:

    mmap(NULL, 8392704, PROT_READ|PROT_WRITE,  MAP_PRIVATE|MAP_ANONYMOUS|MAP_STACK, -1, 0) = 0x7fbbe4bd0000
    mprotect(0x7fbbe4bd0000, 4096, PROT_NONE) = 0
    clone(child_stack=0x7fbbe53cffb0, flags=CLONE_VM|CLONE_FS|CLONE_FILES|CLONE_SIGHAND|CLONE_THREAD|CLONE_SYSVSEM|CLONE_SETTLS|CLONE_PARENT_SETTID|CLONE_CHILD_CLEARTID, parent_tidptr=0x7fbbe53d09d0, tls=0x7fbbe53d0700, child_tidptr=0x7fbbe53d09d0) = -1 ENOMEM (Cannot allocate memory)
    rt_sigprocmask(SIG_BLOCK, [CHLD], [], 8) = 0
    rt_sigaction(SIGCHLD, NULL, {SIG_DFL, [], 0}, 8) = 0
    rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
 
And also the virtual memories of the process could totally have is on another limitation of this.

While there still has others to limit this, on my one environment the max threads of the system is always around 32400.

@mikechristie 

Yes. I think it is. But not very sure for now and still tunning this.

mikechristie:
Just wanted to make sure you were not waiting on me on this one. Last comment sounded like you were trying some other settings.

lxbsz:
@mikechristie  Sorry for missed you last comment.
Please refer to https://github.com/open-iscsi/tcmu-runner/pull/351.

mikechristie:
Closing assuming it was just the leak.

PR #334, opened 2017-11-20
0x3a support

zjcmszh:


lxbsz:
The tgt code is under GPL, but the tcmu-runner is under Apache. So we cannot integrate the tgt code.

Please see: https://github.com/open-iscsi/tcmu-runner/pull/73

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2308,361 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	case WRITE_SAME_16:
+		goto sup;
+	default:
+		return 0;
+	}
+sup:
+	return 1;
+}
+
+int report_opcode_one(struct tcmulib_cmd *cmd, int rctd, uint8_t opcode, uint16_t service_action, int is_service_action, uint32_t alloc_len)
```
mikechristie: Try to follow 80 char columns like in the rest of the code.

If in doubt about coding style go with the kernel coding style:
https://www.kernel.org/doc/html/v4.10/process/coding-style.html

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2308,361 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+{
+	struct iovec *iovec = cmd->iovec;
+	size_t iov_cnt = cmd->iov_cnt;
+	int cdb_len = get_cdb_length(opcode);
+        uint32_t avail_len;
```
mikechristie: Extra space.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2308,361 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+
+	if (buf) {
+		buf[0] = op;
+		return buf;
+	}
+
+	return usage;
+}
+
+int is_support_opcode( int opcode )
```
mikechristie: No spaces between the "(" ")" and "int opcode".

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2308,361 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+{
+	switch( opcode ){
+	case INQUIRY:
+  		goto sup;
+	case TEST_UNIT_READY:
+		goto sup;
+	case WRITE_SAME:
+		goto sup;
+	case WRITE_SAME_16:
+		goto sup;
```
mikechristie: Just do the fall through way:

```
case INQUIRY:
case TEST_UNIT_READY:
case WRITE_SAME:
....
    return true.
```

Also use bools.


What is this function for though? It is not clear to me why it only checks these specific commands?

Also, I think we will need a way for handler to report specific commands they support. For example, file_optical and file_zbc support some uncommon commands.

mikechristie:
Thanks for the patch @zjcmszh  and thanks for catching that @lxbsz.

I gave you some generic review comments for things like coding style if you decide to redo the patch without the licensing issues.

zjcmszh:
Thanks a lot for your suggestion which is very practical,and I will take it. @lxbsz  @mikechristie .

mikechristie:
Sorry I missed you re-pushing this PR! Will look at it this week.

In the future, just add a comment to the PR if you do a "push -f" to update the patches, because I do not get notifications when you do a forced push.

zjcmszh:
Updated patches, please refer to it. 
Thanks @mikechristie

tcmur_cmd_handler.c
```diff
@@ -35,6 +34,15 @@
 #include "tcmu-runner.h"
 #include "alua.h"
 
+#define CONTROL_BYTE 0x07  
+
+typedef struct{
```
lxbsz: There should be one space between "struct" and '{'

tcmur_cmd_handler.c
```diff
@@ -35,6 +34,15 @@
...
+		uint8_t opcode;
+		uint8_t service_action;
+		int cdb_length;
+		uint8_t cdb_usage[16];
+	}cdb_info;
```
lxbsz: Please remove the spaces here.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2316,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+			break;
+	}
+	if(buf)
+		return buf;
+	return usage;
+}
+
+int is_support_opcode(int opcode) 
+{
+	switch(opcode){
```
lxbsz: Space here.
switch (opcode) {

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2316,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	case RECEIVE_COPY_RESULTS:
+	case MAINTENANCE_IN:
+	case UNMAP:
+	case EXTENDED_COPY:
+	case READ_16:
+	case WRITE_16:
+	case WRITE_SAME_16:
+	case WRITE_VERIFY_16:
+	case FORMAT_UNIT:
+		goto sup;
```
lxbsz: You could just "return 1" here instead of "goto sup".

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2316,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	default:
+		return 0;
+	}
+sup:
+	return 1;
```
lxbsz: Then these could be removed.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2316,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+
+	memcpy(ptr, get_cdb_usage_data(opcode, service_action), cdb_len);
+	ptr += cdb_len;
+
+	if (rctd) {
+		ptr[1] = 0x0a;
+		ptr += 12;
+	}
+	avail_len = ptr - &buf[0];
+	tcmu_memcpy_into_iovec( iovec, iov_cnt, buf, avail_len );
```
lxbsz: Please remove spaces just after '(' and before ')'. 
And the same with the other places.

tcmur_cmd_handler.c
```diff
@@ -2458,3 +2785,4 @@ int tcmur_generic_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		ret = tcmur_cmd_handler(dev, cmd);
 	return ret;
 }
+
```
lxbsz: Please remove the useless space line.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2316,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	};
+
+	cdb_info opcode_format_unit = {
+		.opcode = FORMAT_UNIT,
+		.cdb_length = 6,
+		.cdb_usage = {FORMAT_UNIT, 0x00, 0xff, 0xff, 0xff, CONTROL_BYTE},
+	};
+
+	switch(op){
+		case INQUIRY:
```
lxbsz: Should keep the code style the same as the current ones.
like:

    switch () {
    case XX:
            .....
            break;
    case YY:
    ....
    }

tcmur_cmd_handler.c
```diff
@@ -2354,9 +2676,14 @@ static int handle_sync_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
-			return handle_rtpg(dev, cmd);
-		return TCMU_NOT_HANDLED;
+		switch(cdb[1] & 0x1f) {
+			case MI_REPORT_TARGET_PGS:
+				return handle_rtpg(dev, cmd);
+	        	case MI_REPORT_SUPPORTED_OPERATION_CODES:
+                                return report_supported_opcodes(cmd);
+			default:
+				return TCMU_NOT_HANDLED;
+                }
```
lxbsz: Same here.

lxbsz:
Please remove the noise file:
 
    root@10.254.4.23

And the noise in commit comment:

    Conflicts:
	tcmur_cmd_handler.c

zjcmszh:
Thank you very much for your detailed suggestions@lxbsz.  Updated patches.@mikechristie

mikechristie:
Did you test these patches? Do they work with sg utils's sg_opcodes command?

tcmu-runner.spec
```diff
@@ -0,0 +1,75 @@
+Name:          tcmu-runner
```
mikechristie: I think you added this by mistake.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return ret;
```
mikechristie: In this file below, if the function you are adding is only used in this file make the
function static.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
 }
 
+int get_cdb_length(uint8_t opcode)
+{
+	int cdb_size[8] = {6,10,10,12,16,12,10,10};
+	return cdb_size[((opcode)>>5)&7];
+}
```
mikechristie: Why can't you just use tcmu_get_cdb_length?

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		buf = opcode_format_unit.cdb_usage;
+		break;
+	default:
+		break;
+	}
+
+	return buf;
+}
+
+int is_support_opcode(int opcode)
```
mikechristie: Drop the extra space after the ending ")".

Also, I think it should be renamed to:

is_supported_opcode()

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+
+int report_opcode_one(struct tcmulib_cmd *cmd, int rctd, uint8_t opcode, 
+                      uint16_t service_action, int is_service_action)
+{
+	struct iovec *iovec = cmd->iovec;
+	size_t iov_cnt = cmd->iov_cnt;
+	int cdb_len = get_cdb_length(opcode);
+	uint8_t buf[256], *ptr, *usage_data;
+	
+	if ((!is_support_opcode(opcode)) || (is_service_action && !service_action) ||
```
mikechristie: You do not need to add "(" ")" around (!is_support_opcode(opcode)).

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+           (!is_service_action && service_action))
```
mikechristie: Same.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+                      uint16_t service_action, int is_service_action)
+{
+	struct iovec *iovec = cmd->iovec;
+	size_t iov_cnt = cmd->iov_cnt;
+	int cdb_len = get_cdb_length(opcode);
+	uint8_t buf[256], *ptr, *usage_data;
+	
+	if ((!is_support_opcode(opcode)) || (is_service_action && !service_action) ||
+           (!is_service_action && service_action)) 
+		return SAM_STAT_CHECK_CONDITION;
```
mikechristie: You can't just return SAM_STAT_CHECK_CONDITION because there is nothing
in the sense buffer. You need to set the sense and ASC/ASCQ by calling
tcmu_set_sense_data(). Here I think you wanted to return ILLEGAL_REQUEST
and ASC_INVALID_FIELD_IN_CDB for the asc/ascq value.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+
+	memset(buf, 0, sizeof(buf));
+	ptr = &buf[0];
+	ptr++;
+	*ptr++ = rctd ? 0x83 : 0x03;
```
mikechristie: Could you add a comment about what this bit is here and below. Just maybe the same of
the bit in plain english like it is in the spec. No one remembers what
these are off the top of their head so it makes reading the code a little
easier. Also if there is a mistake we can at least figure out what we were
trying to do :)  Do that below too.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	uint8_t reporting_options;
+	uint8_t opcode;
+	uint16_t requested_service_action;
+	uint32_t alloc_len;
+	int rctd;
+	int ret = SAM_STAT_CHECK_CONDITION;
+
+	reporting_options = cmd->cdb[2] & 0x07;
+	opcode = cmd->cdb[3];
+	requested_service_action = htobe16((uint16_t)cmd->cdb[4]);
```
mikechristie: I think this is reverse? The SCSI struct is in BE and you want to
convert to the host.

Use get_unaligned_be16.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+     
+	alloc_len = htobe32((uint32_t)cmd->cdb[6]);
```
mikechristie: Use get_unaligned_be32

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	if (tcmu_get_cdb_length(cmd->cdb) < alloc_len)
```
mikechristie: To get the alloc_len you want to use tcmu_get_xfer_length which returns
the amount of data we are going to transfer.

The get cdb length function just returns the cdb length (6, 10, 12, etc)
so you do not want to compare that to alloc_len as it will almost always
fail.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2318,312 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	rctd = cmd->cdb[2] & 0x80;
+
+ 	switch (reporting_options) {
+	case 0x01: 
+		return report_opcode_one(cmd, rctd, opcode,
+                             requested_service_action, 0);
+	case 0x02: 
+		return report_opcode_one(cmd, rctd, opcode,
+                             requested_service_action, 1);
+	default:
```
mikechristie: You can't just return SAM_STAT_CC. You need to fill out the sense buffer. In this case you can just return TCMU_NOT_HANDLED and libtcmu will handle everything for you.

tcmu-runner.spec
mikechristie: Just to be clear, I mean I think you added that patch or file by mistake, because we already added a spec.

zjcmszh:
Patches updated @mikechristie   
test result：
[root@zj-client ~]# sg_opcodes /dev/sdc --opcode=0x42

  Opcode=0x42
  Command_name: Unmap
  Command supported [conforming to SCSI standard]
  Usage data: 42 01 00 00 00 00 00 ff ff 07 
[root@zj-client ~]# sg_opcodes /dev/sdc --opcode=0x83

  Opcode=0x83
  Command_name: Extended copy(LID1)
  Command supported [conforming to SCSI standard]
  Usage data: 83 00 00 00 00 00 00 00 00 00 ff ff ff ff 00 07

tcmur_cmd_handler.c
zjcmszh: I intend to obtain the cdb_length of REQUESTED OPERATION CODE(cmd->cdb[3]), parameter type of tcmu_get_cdb_length is uint8_t*, meanwhile, data type of REQUESTED OPERATION CODE is uint8_t, therefore, I need to rewrite one function that used to calculate cdb_length.

mikechristie: Take the code in the switch in tcmu_get_cdb_length, make a new function out of it and name it something like tcmu_get_cdb_length_by_opcode. Then have tcmu_get_cdb_length call that new function and your code.

zjcmszh:
Patches updated, please refer to it, thanks.@mikechristie

zjcmszh:
@mikechristie  Patches updated, please refer to it. Thanks.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2319,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return ret;
 }
 
+static uint8_t *get_cdb_usage_data(uint8_t op, uint8_t sa)
+{
+	static uint8_t usage[16];
+	uint8_t *buf = usage;
+
```
mikechristie: Hey,

I think the patch looks ok.

Just FYI, it is taking me a while to check the cbd_usage bits. It is a not so fun task :)

I think some commands the kernel handles for us like report luns and the PGR ones are not handled here.

tcmur_cmd_handler.c
```diff
@@ -2308,6 +2319,320 @@ static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	return ret;
```
seroyer: It would be safer to return const uint8_t *, to prevent any callers from accidentally modifying the static data.  It appears to be used safely for now, though.

tcmur_cmd_handler.c
zjcmszh: @mikechristie I apologized that It was not until now that I read the mail. 
Thanks a lot.

tcmur_cmd_handler.c
zjcmszh: Thanks for your suggestion.
Updated patches.

PR #339, opened 2017-12-06
fix restarts from unclean shutdowns

mikechristie:
The following patches allow us to restart runner after it has crashed. They require the tcmu restart patches posted to target-devel (no link yet as spinics has not updated).

### DIFF
--- a/configfs.c
+++ b/configfs.c
@@ -228,3 +228,13 @@ int tcmu_set_cfgfs_ul(const char *path, unsigned long val)
 	sprintf(buf, "%lu", val);
 	return tcmu_set_cfgfs_str(path, buf, strlen(buf) + 1);
 }
+
+int tcmu_exec_cfgfs_dev_action(struct tcmu_device *dev, const char *name,
+			       unsigned long val)
+{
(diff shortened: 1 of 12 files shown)


bgly:
Nice! I'd like to test this but I am pinned to a super old tcmu-runner where my fbo reconfig patch still works.... I'm trying to update that and move all our code to the latest tcmu-runner.

mikechristie:
@bgly, just FYI, there is one more issue.

These patches will restart runner cleanly if runner crashes or if you do

systemctl stop tcmu-runner

while LUNs are mapped to tcmu devices.

For the systemctl shutdown case, I am working on a smoother shutdown where commands do not timeout in the ring if runner stops while IO is in there still. The restart code in this patchset will be the same for that case and the unclean shutdown case.

mikechristie:
Pushed a update to these patches. Both the clean and unclean shutdown cases and restart from both should be working in this patchset.

zhuozh:
@mikechristie @bgly We tested pkill and restart runner with IO in flight, both clean and unclean shutdown works well with your patchset.

PR #342, opened 2017-12-22
tcmu:fix error using the return value of function  tcmu_get_dev_max_xfer_length

tangwenji:
The return value of fuction tcmu_get_dev_max_xfer_len is number of sector, but some functions used it as data size, for example handle_writesam and handle_format_unit functions.


mikechristie:
Thanks!

PR #345, opened 2018-01-03
libtcmu: fix traverse darray for remove_device

pengliang233:
Otherwise some devices will not be removed when the devices more than 1.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -517,7 +517,7 @@ static void close_devices(struct tcmulib_context *ctx)
 	struct tcmu_device *dev;
 	char *cfgstring = "";
 
-	darray_foreach(dev_ptr, ctx->devices) {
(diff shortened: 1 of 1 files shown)


mikechristie:
Nice.

This patch breaks my patchset in PR 339:

https://github.com/open-iscsi/tcmu-runner/pull/339

which I thought fixed that code path for once, but I guess I missed this bug :)

I will merge your patch after my set, because there is a conflict and it will be easier to merge your 1 liner after my set.

Thanks!

mikechristie:
Hey, I ported this patch for you and merged it in this PR

https://github.com/open-iscsi/tcmu-runner/pull/339/commits/0580fd2aa30417f47c2ce9d56c695610316281f4

with the other shutdown related patches.

Thanks.

pengliang233:
ok, that's good. @mikechristie

PR #346, opened 2018-01-03
rbd: do not recheck size on resize event

mikechristie:
There is not need to recheck the size on a resize event
because the calling app, gwcli/ansible, has already set
the ceph side size and then called us. This works around
the issue where if we are blacklisted the size check call
will fail and we end up with a partially updated device.


zhuozh:
@mikechristie Thanks for your remind. Tested yours and Dillaman's patch, Pefectly solved all the resize problem we came across.
Please ignore this PR.

zhuozh:
Test pass, and looks good to me.

PR #347, opened 2018-01-04
tcmu: Reconfig to handle multiple attributes

bgly:
This patch depends on the kernel patch:
tcmu: Allow reconfig to handle multiple attributes

It allows for userspace to process a reconfig netlink
that contains more than one attribute to change.

Example command:
set attribute dev_reconfig=dev_config=fbo//home/path:dev_size=2147483648


mikechristie:
Just FYI,  I will get to this, this week.

PR #351, opened 2018-01-18
alua: fix memory leakage when doing the alua implicit transition

lxbsz:
The error log as:
"Could not start implicit transition thread:Cannot allocate memory"

When a joinable thread terminates, its memory resources (thread
descriptor and stack) are not deallocated until another thread
performs pthread_join on it. Therefore, pthread_join must be
called once for each joinable thread created to avoid memory leaks.

Here we will make the lock pthread as detached to avoid this.

### DIFF
--- a/main.c
+++ b/main.c
@@ -805,6 +805,10 @@ static int dev_added(struct tcmu_device *dev)
 
 	rdev->flags |= TCMUR_DEV_FLAG_IS_OPEN;
 
+	ret = pthread_cond_init(&rdev->lock_cond, NULL);
+	if (ret < 0)
+		goto cleanup_lock_cond;
+
 	/*
 	 * Set the optimal unmap granularity to max xfer len. Optimal unmap
 	 * alignment starts at the begining of the device.
@@ -821,6 +825,8 @@ static int dev_added(struct tcmu_device *dev)
 
 close_dev:
 	rhandler->close(dev);
+cleanup_lock_cond:
+	pthread_cond_destroy(&rdev->lock_cond);
 cleanup_aio_tracking:
 	cleanup_aio_tracking(rdev);
 cleanup_io_work_queue:
@@ -863,6 +869,10 @@ static void dev_removed(struct tcmu_device *dev)
 
 	cleanup_io_work_queue(dev, false);
 	cleanup_aio_tracking(rdev);
+
+	ret = pthread_cond_destroy(&rdev->lock_cond);
+	if (ret != 0)
+		tcmu_err("could not cleanup lock cond %d\n", ret);
 
 	ret = pthread_mutex_destroy(&rdev->state_lock);
 	if (ret != 0)
--- a/tcmur_device.c
+++ b/tcmur_device.c
@@ -43,7 +43,7 @@ bool tcmu_dev_in_recovery(struct tcmu_device *dev)
 /*
  * TCMUR_DEV_FLAG_IN_RECOVERY must be set before calling
  */
-int __tcmu_reopen_dev(struct tcmu_device *dev)
+int __tcmu_reopen_dev(struct tcmu_device *dev, bool in_lock_thread)
 {
 	struct tcmur_device *rdev = tcmu_get_daemon_dev_private(dev);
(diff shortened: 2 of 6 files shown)


lxbsz:
This is similar as https://github.com/open-iscsi/tcmu-runner/pull/332

main.c
```diff
@@ -847,6 +853,8 @@ static void dev_removed(struct tcmu_device *dev)
 	rdev->flags |= TCMUR_DEV_FLAG_STOPPING;
 	pthread_mutex_unlock(&rdev->state_lock);
 
+	pthread_cond_destroy(&rdev->lock_cond);
+
```
mikechristie: Nice find.

This should be moved to the end of dev_removed with the mutext destroy calls. The lock_cond can still be used a couple lines down in tcmur_stop_device -> tcmu_cancel_lock_thread

tcmur_device.c
```diff
@@ -207,26 +207,26 @@ void tcmu_notify_lock_lost(struct tcmu_device *dev)
...
+
 	/*
 	 * We may be called from the lock thread to reopen the device.
 	 */
-	if (ret != EDEADLK)
-		tcmu_dev_err(dev, "pthread_join failed with value %d\n", ret);
+	tcmu_dev_dbg(rdev->dev, "waiting for lock thread to exit\n");
+	ret = pthread_cond_wait(&rdev->lock_cond, &rdev->state_lock);
+	pthread_mutex_unlock(&rdev->state_lock);
+
```
mikechristie: I think we can hit a deadlock here. See the comment above about if we are called from tcmu_acquire_dev_lock -> tcmu_reopen_dev -> tcmu_cancel_lock_thread. In this case we would get stuck in the wait call above.  I can send a patch over this though. Will update this with the PR.

lxbsz: Yes, just see your this reply.
I have pushed a new one. but still has some problems.

main.c
lxbsz: This has been fixed.

tcmur_device.c
```diff
@@ -207,26 +208,26 @@ void tcmu_notify_lock_lost(struct tcmu_device *dev)
...
 	/*
 	 * It looks like lock calls are not cancelable, so
 	 * we wait here to avoid crashes.
 	 */
 	tcmu_dev_dbg(dev, "Waiting on lock thread\n");
-	ret = pthread_join(rdev->lock_thread, &join_retval);
+
 	/*
 	 * We may be called from the lock thread to reopen the device.
 	 */
```
mikechristie: Remove the comment above since that is not true any more.

tcmur_device.c
mikechristie: >  I have pushed a new one. but still has some problems.

What issues are you still hitting? Is it a bug added in this patch or did you mean something related to the issue https://github.com/open-iscsi/tcmu-runner/pull/332
?

I tested the patch real quick here and it seems ok.

tcmur_device.c
lxbsz: Has removed.

tcmur_device.c
lxbsz: Sorry, I think I have just copied the wrong rpm packages to do the test.
And it works fine on my RHEL 7.2(this is very easy to reproduce this bug) and Centos environments.

lxbsz:
@mikechristie Have updated all mentioned above, please have a review.
Thanks,

tcmur_device.c
```diff
@@ -276,7 +274,7 @@ int tcmu_acquire_dev_lock(struct tcmu_device *dev)
 	switch (ret) {
 	case TCMUR_LOCK_BUSY:
 		new_state = TCMUR_DEV_LOCK_LOCKING;
-		break;
+		return ret;
```
mikechristie: Let's just remove all the TCMUR_LOCK_BUSY code since it is not used. I think without your patch it is buggy and with your patch we would hit a new bug where we leave the device blocked.

lxbsz: Actually this code wouldn't be executed by any case.
Just removed it.

mikechristie:
Thanks.

PR #352, opened 2018-01-19
libtcmu:fixed memory leaks of log_dir_path

tangwenji:
If the user has configured log_dir_path in tcmu.conf file,there will be a memory leak.


mikechristie:
Thanks.

PR #353, opened 2018-01-19
libtcmu:fixed conf file parse error

tangwenji:
If there is no space betwwen the option and the character '=',
     after the following while loop, the charactor '=' may be replace by '\0'
     and therefore can not be parsed correctly.


mikechristie:
Thanks.

Issue #350, opened 2018-01-16
The multipath device mounted as read-only


lxbsz:
Using the ceph iscsi gateway tools and just after the multipath device is made filesystem(XFS or EXT4), sometimes the mount will return fail as read-only, and then just wait for a moment, it will be okay.

@mikechristie Any ideas about this ? As the rbd handler is async type for now. Could this be as the later read is a little faster than the mkfs.xfs/ext4's format data writing ? 

Thanks

mikechristie:
Normally, I do not think we could hit a issue where a read sent after a mkfs could be executed before any of the mkfs related IO. Once mkfs is done, the IOs for it are on disk. All later reads will see that data.

If you were doing direct IO reads while while another process was  it was doing a mkfs then you could hit this problem.

If the RO state is only temp, maybe an initiator side timer is off.  You could have a temporary all paths down (APD), when the initiator recovers the paths, the multipath layer is able to execute IOs again.

Are you using the initiator side settings from here:

http://docs.ceph.com/docs/master/rbd/iscsi-initiator-linux/

? With "no_path_retry = queue", IO will be queued when we hit this transient APDs so the FS layer would never see it.

mikechristie:
Send me the kernel logs (/var/log/messages or wherever journctl puts them) from the initiator side.

lxbsz:
Yes, all the settings are the same as http://docs.ceph.com/docs/master/rbd/iscsi-initiator-linux/

We are reproducing the readonly issue but hasn't reproduced it yet, and will send the log then.

Thanks,

lxbsz:
https://github.com/open-iscsi/tcmu-runner/pull/355

lxbsz:
@mikechristie Since after https://github.com/open-iscsi/tcmu-runner/pull/355, the readonly won't be reproduced and this could be closed.

PR #355, opened 2018-01-22
handler: do not perform implicit transition if command not supported

zhuozh:
If unsupported command was handled by tcmur_cmd_handler, alua implicit
transition have a chance to cause image exclusive-lock threshing between
runners, so do not perform implicit transition if command not supported


zhuozh:
This patch is for issue #350

lxbsz:
https://github.com/open-iscsi/tcmu-runner/issues/350

mikechristie:
Thanks.

What OS and version are you hitting this with btw? Is it ESX?

lxbsz:
Mostly both of the CentOS7.X and RHEL7.X with multipath.

PR #360, opened 2018-01-23
runner/lib: use uint64_t for dev_size

mikechristie:
Use uint64_t for dev_size to prevent issues with 32 bit setups.

Attached is the kernel patch that will also be pushed.

[dev-size64-target-pending.txt](https://github.com/open-iscsi/tcmu-runner/files/1654736/dev-size64-target-pending.txt)


mikechristie:
Attached is the correct patch for the kernel.

[dev-size64-target-pending-v2.txt](https://github.com/open-iscsi/tcmu-runner/files/1654780/dev-size64-target-pending-v2.txt)

JuniorJPDJ:
bump

JuniorJPDJ:
> Attached is the correct patch for the kernel.

Did you maybe try to send this patch to linux mainling lists?

mikechristie:
> > Attached is the correct patch for the kernel.
> 
> Did you maybe try to send this patch to linux mainling lists?

The kernel patch is merged upstream already. It was only attached in this PR to make it easier for people to test.

The userspace parts in this PR are not merged so it's still open. I don't work on this project anymore, so feel free to take up the issue/PR/code.

JuniorJPDJ:
Well, would be cool if someone would maintain this software, but it seems dead :/

JuniorJPDJ:
Your commit does loads of other things, not only uses uint64_t for dev size.
Could you please split it and name it correctly?

JuniorJPDJ:
@lxbsz what do you think about this PR? It seems to fix some issues I've stepped into on 32-bit arches, but would need rebase and also probably spliting into more commits.

lxbsz:
> @lxbsz what do you think about this PR? It seems to fix some issues I've stepped into on 32-bit arches, but would need rebase and also probably spliting into more commits.

I am okay with it. If you can test on the 32-bit arches, please go ahead to take it.
I will test it after that.

PR #365, opened 2018-01-29
tcmu-runner: fix bug in EXTENDED_COPY and COMPARE_AND_WRITE command

umuzhaohui:
When EXTENDED_COPY command is processed,  data is
readed from rbd image of the source dev,  the data
should be stored in xcopy’s iovec, not in tcmulib_cmd’s
iovec , because the data will be written to rbd
image of the destination dev from xcopy’s iovec,
not from tcmulib_cmd’s iovec

When COMPARE_AND_WRITE command is processed,
the total iove length should be equal to double
NUMBER OF LOGICAL BLOCKS of the CDB，but
checking logical block address，the last logical
block address should be LOGICAL BLOCK ADDRESS
of the CDB plus NUMBER OF LOGICAL BLOCKS of the CDB.

When COMPARE_AND_WRITE command is processed,
after the data which was readed from rbd image
was compared, tcmu runner should seek the data
which will be wtitten to  rbd image  in
tcmulib_cmd’s iovec, tcmu runner not only upate
the tcmulib_cmd’s iovec but also update the
tcmulib_cmd’s iovec iov_cnt.

Signed-off-by: Zhaohui Yang <yangzhaohui@umcloud.com>


mikechristie:
Thanks for the patch.

Could you split the patch into multiple patches with a fix/change per patch?

tcmur_cmd_handler.c
```diff
@@ -142,6 +142,10 @@ static int check_lba_and_length(struct tcmu_device *dev,
 	ret = check_iovec_length(dev, cmd, sectors);
 	if (ret)
 		return ret;
+    
+    if (cdb[0] == COMPARE_AND_WRITE) {
+        sectors = sectors/2;
+    }
```
mikechristie: I think what you wanted to fix here was fixed already.

We want to be checking this from:

tcmur_handle_caw -> handle_caw_check (here sectors is whatever is in the cdb) -> check_lbas

so check_lba_and_length is never called for a CAW and we should not have to do the sectors divided by 2 trick.

rbd.c
```diff
@@ -943,8 +943,18 @@ static void rbd_finish_aio_generic(rbd_completion_t completion,
 {
 	struct tcmu_device *dev = aio_cb->dev;
 	struct tcmulib_cmd *tcmulib_cmd = aio_cb->tcmulib_cmd;
-	struct iovec *iovec = tcmulib_cmd->iovec;
-	size_t iov_cnt = tcmulib_cmd->iov_cnt;
+
+    struct iovec *iovec;
```
mikechristie: We use linux kernel coding style when we can, so use tabs instead of spaces and the tabs must be 8 cols wide.

We also prefer to declare variables at the beginning of the function, so this chunk of code would go after the "int tmu_r" declaration in this function. It is ok to do declare variables later if they are inside a if/loop.

tcmur_cmd_handler.c
```diff
@@ -1802,7 +1806,7 @@ static void handle_caw_read_cbk(struct tcmu_device *dev,
 	}
 
 	/* perform write */
-	tcmu_seek_in_iovec(origcmd->iovec, state->requested);
+	tcmu_seek_in_cmd_iovec(origcmd, state->requested)
```
mikechristie: Missing ";".

rbd.c
```diff
@@ -943,8 +943,18 @@ static void rbd_finish_aio_generic(rbd_completion_t completion,
...
+    size_t iov_cnt;
+    if ((tcmulib_cmd->cdb[0] & 0x83) == 0x83){
```
mikechristie: Each handler should should not have to know what command is being executed and know this low level of details about info like cmdstate.

We have wanted to rewrite the iovec code, because some handlers modify the iovec and some do not. The tcmur_cmd_handler code was supposed to temporarily hack around this and handle it for all handlers. Let me get back to you on this in a day or so when I have more time. I am on PTO right now, so do not have access to everything I would like to be able to give you a good review comment on how to handle this properly.

lxbsz:
Good catch about the XCOPY bug.

tcmur_cmd_handler.c
umuzhaohui: I have fixed it. Thanks for your review.

rbd.c
umuzhaohui: I have fixed it. Thanks for your review.

tcmur_cmd_handler.c
umuzhaohui: I have fixed it. Thanks for your review.

rbd.c
```diff
@@ -943,12 +943,21 @@ static void rbd_finish_aio_generic(rbd_completion_t completion,
...
-	struct iovec *iovec = tcmulib_cmd->iovec;
-	size_t iov_cnt = tcmulib_cmd->iov_cnt;
 	uint32_t cmp_offset;
 	uint16_t asc_ascq;
 	int64_t ret;
 	int tcmu_r;
+	struct iovec *iovec;
+	size_t iov_cnt;
+
+	if ((tcmulib_cmd->cdb[0] & 0x83) == 0x83) {
```
mikechristie: As I said before the handlers should not be digging into the low level details. I had also said we should start to rewrite this code. However, let's do a quick fix here first. It looks like a rbd.c specific bug that does the read/write callouts but also digs into the tcmulib_cmd. It should just be using the iovec passed into the read/write callouts.

So for now just add new fields onto rbd_aio_cb  for the iovec and iov_cnt. Have tcmu_rbd_write/read set the new fields to the values passed in. In the completion function then just access that, so rbd.c is not digging into the tcmulib_cmd fields.

api.c
```diff
@@ -160,6 +160,27 @@ void tcmu_seek_in_iovec(struct iovec *iovec, size_t count)
 	}
 }
 
+/*
+ * Consume an iovec. Count must not exceed the total iovec[] size.
+ * iove count should be updated.
+ */
+void tcmu_seek_in_cmd_iovec(struct tcmulib_cmd *origcmd, size_t count)
```
mikechristie: origcmd should be renamed to just cmd, because it is a general function and my be not be used in the tcmur_cmd_handler specific way.

api.c
```diff
@@ -160,6 +160,27 @@ void tcmu_seek_in_iovec(struct iovec *iovec, size_t count)
...
+{
+	struct iovec *iovec = origcmd->iovec;
```
mikechristie: Add a newline here.

rbd.c
mikechristie: Oh yeah, when you re-upload this next time break the patches in to 2 patches. One for the xcopy issue and one for the CAW seek issue. Thanks.

mikechristie: Hey, @umuzhaohui  I saw you uploaded new versions of these patches. Were you also going to handle this review comment?

Here is a untested patch that does what I was requesting. Does that fix your issue?

[rbd-dont-access-cmd-iov.txt](https://github.com/open-iscsi/tcmu-runner/files/1790624/rbd-dont-access-cmd-iov.txt)

api.c
```diff
@@ -160,6 +160,27 @@ void tcmu_seek_in_iovec(struct iovec *iovec, size_t count)
...
+			iovec->iov_len = 0;
+			iovec++;
+			cmd->iov_cnt--;
+		} else {
+			iovec->iov_base += count;
+			iovec->iov_len -= count;
+			count = 0;
+		}
+	}
+}
```
mikechristie: For this patch could we just do the attached so we do not duplicate that loop.

[api-seek.txt](https://github.com/open-iscsi/tcmu-runner/files/1790648/api-seek.txt)

tcmur_cmd_handler.c
umuzhaohui: OK

mikechristie: > OK

Hey, I saw this reply, but was not sure what it is for.

Did you see my other comments with the patches I attached:

https://github.com/open-iscsi/tcmu-runner/pull/365#discussion_r172960667
https://github.com/open-iscsi/tcmu-runner/pull/365#discussion_r172963556

?

In those comments, I was asking that you drop some of your code, incorporate my code into your patchset, confirm it works for your tests and then repush. This way you still get the GitHub pull request credit.

If you do not have time, we can do the reverse where I take your patches, merge them with mine, add your signed-off, and then make a new pull request. I thought you might not want to do this because you would just get a signed off and it might not show up in your GitHub pull request stats.

mikechristie:
Thanks.

rbd.c
```diff
@@ -943,12 +945,14 @@ static void rbd_finish_aio_generic(rbd_completion_t completion,
...
-	struct iovec *iovec = tcmulib_cmd->iovec;
-	size_t iov_cnt = tcmulib_cmd->iov_cnt;
+	struct iovec *iov = aio_cb->iov;
+	size_t iov_cnt = aio_cb->iov_cnt;
 	uint32_t cmp_offset;
 	uint16_t asc_ascq;
 	int64_t ret;
 	int tcmu_r;
+	struct iovec *iovec;
+	size_t iov_cnt;
```
mikechristie: Note: I fixed this in commit:

https://github.com/open-iscsi/tcmu-runner/commit/d1422948d0285c3453fff5f43abf4ebe6459ff82

PR #367, opened 2018-02-01
extra: Add make_runnerrpms.sh support

lxbsz:
This script will simplify the RPMs package building.

USAGE:

Will build the RPMs in current dir by using the HEAD commit ID as default.


lxbsz:
The package will be like:
tcmu-runner-v1.3.0.5.g182d05e-0.el7.centos.x86_64.rpm
tcmu-runner-debuginfo-v1.3.0.5.g182d05e-0.el7.centos.x86_64.rpm

mikechristie:
Thanks for the patch.

Would it be difficult to drop the "v" prefix. It does not seem standard and I am worried it might throw off some scripts version parsing.

lxbsz:
Yes, I will update by dropping it.

lxbsz:
tcmu-runner-1.3.0.5.g9d057bd-0.el7.centos.x86_64.rpm
tcmu-runner-debuginfo-1.3.0.5.g9d057bd-0.el7.centos.x86_64.rpm

Updated and the packages as above.

mikechristie:
Thanks!

PR #368, opened 2018-02-02
config: add the logdir option dynamic reloading support

lxbsz:
NOTE: for now this couldn't support the glfs's logging setting, and
the glfs will use the old path .

### DIFF
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -130,6 +130,8 @@ do { \
 	char buf[1024]; \
 	option = tcmu_get_option(#key); \
 	if (option) { \
+		if (cfg->key) \
+			free(cfg->key); \
 		cfg->key = strdup(option->opt_str); \
 		if (option->opt_str) \
 			free(option->opt_str); \
@@ -179,6 +181,7 @@ static void tcmu_conf_free_str_keys(struct tcmu_config *cfg)
 	 * For example:
 	 * TCMU_FREE_CFG_STR_KEY(cfg, 'STR KEY');
 	 */
+	 TCMU_FREE_CFG_STR_KEY(cfg, log_dir_path);
 }
 
 #define TCMU_MAX_CFG_FILE_SIZE (2 * 1024 * 1024)
@@ -545,8 +548,6 @@ void tcmu_destroy_config(struct tcmu_config *cfg)
 	}
 
 	tcmu_conf_free_str_keys(cfg);
-	if (cfg->log_dir_path)
-		free(cfg->log_dir_path);
 	free(cfg->path);
 	free(cfg);
 }
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -114,12 +114,13 @@ do { \
 	} \
 } while (0)
 
-#define TCMU_PARSE_CFG_BOOL(cfg, key) \
+#define TCMU_PARSE_CFG_BOOL(cfg, key, def) \
 do { \
 	struct tcmu_conf_option *option; \
 	option = tcmu_get_option(#key); \
 	if (option) { \
 		cfg->key = option->opt_bool; \
+		option->opt_bool = def; \
 	} \
 } while (0)
 
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -123,12 +123,18 @@ do { \
 	} \
 } while (0)
 
-#define TCMU_PARSE_CFG_STR(cfg, key) \
+#define TCMU_PARSE_CFG_STR(cfg, key, def) \
 do { \
 	struct tcmu_conf_option *option; \
+	char buf[1024]; \
 	option = tcmu_get_option(#key); \
 	if (option) { \
-		cfg->key = strdup(option->opt_str); } \
+		cfg->key = strdup(option->opt_str); \
+		if (option->opt_str) \
+			free(option->opt_str); \
+		sprintf(buf, "%s", def); \
+		option->opt_str = strdup(buf); \
+	} \
 } while (0);
 
 #define TCMU_FREE_CFG_STR_KEY(cfg, key) \
@@ -146,7 +152,7 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg, bool reloading)
 
 	if (!reloading) {
 		/* set log_dir path option */
-		TCMU_PARSE_CFG_STR(cfg, log_dir_path);
+		TCMU_PARSE_CFG_STR(cfg, log_dir_path, TCMU_LOG_DIR_DEFAULT);
 		/*
 		 * The priority of the logdir setting is:
 		 * 1, --tcmu_log_dir/-l LOG_DIR_PATH
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -153,9 +153,9 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg, bool reloading)
 		tcmu_set_log_level(cfg->log_level);
 	}
 
+	/* set log_dir path option */
+	TCMU_PARSE_CFG_STR(cfg, log_dir_path, TCMU_LOG_DIR_DEFAULT);
 	if (!reloading) {
-		/* set log_dir path option */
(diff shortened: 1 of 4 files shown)


libtcmu_config.c
```diff
@@ -149,11 +151,19 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg, bool reloading)
...
 		if (!tcmu_get_logdir())
-			tcmu_logdir_create(cfg->log_dir_path);
+			tcmu_logdir_create(cfg->log_dir_path, false);
 		else
 			tcmu_warn("The logdir option from the tcmu.conf will be ignored\n");
 	} else {
-		tcmu_warn("The logdir option is not supported by dynamic reloading for now!\n");
+		/*
+		 * Here we asume that users want to change the
+		 * log_dir_path without considering tha priority
```
mikechristie: Change "tha" to "the".

libtcmu_config.c
```diff
@@ -149,11 +151,19 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg, bool reloading)
...
+		 * mentioned above.
+		 */
+		tcmu_logdir_create(cfg->log_dir_path, true);
+		ret = tcmu_create_file_output(tcmu_get_log_level(), TCMU_LOG_FILENAME, true);
+		if (ret < 0)
+			tcmu_err("create file output error \n");
```
mikechristie: Let's change this to:

Could not change log path to %s.\n, new log dir path

I would say also report the error code, but create_file_output is a mess of -1 and -errno.

libtcmu_config.c
```diff
@@ -149,11 +151,19 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg, bool reloading)
...
 			tcmu_warn("The logdir option from the tcmu.conf will be ignored\n");
 	} else {
-		tcmu_warn("The logdir option is not supported by dynamic reloading for now!\n");
+		/*
+		 * Here we asume that users want to change the
+		 * log_dir_path without considering tha priority
+		 * mentioned above.
+		 */
+		tcmu_logdir_create(cfg->log_dir_path, true);
+		ret = tcmu_create_file_output(tcmu_get_log_level(), TCMU_LOG_FILENAME, true);
```
mikechristie: 80 char cols.

libtcmu_log.c
```diff
@@ -416,15 +435,30 @@ static void log_output(int pri, const char *msg)
...
+				output->output_fn(pri, timestamp, msg, output->data);
+		} else {
+			/*
+			 * We just close and free the resource here to make
+			 * sure no outputing operation is in process.
+			 */
+			if (output->close_fn != NULL)
+				output->close_fn(output->data);
+			if (output->name != NULL)
+				free(output->name);
```
mikechristie: Since this is the same as what's in log_cleanup, let's put it in a function.

libtcmu_log.c
```diff
@@ -275,12 +276,26 @@ static int append_output(log_output_fn_t output_fn, log_close_fn_t close_fn, voi
...
 	return 0;
 }
 
+static void log_output_disable(const tcmu_log_destination dest)
+{
+	struct log_output *output;
+
+	/* This will just find the first one and disable it. */
+	darray_foreach(output, logbuf->outputs) {
+		if (output->dest == dest) {
```
mikechristie: Do we need to also check for enabled == true. This would work:

1. update 1.
2. something does tcmu_log()
3. update 2.

If there were 2 updates back to back

1. update 1.
2. update 2.
3. something does tcmu_log()

then this would only disable the first one twice and the later delayed removal in log_output would only remove the original.

libtcmu_config.c
mikechristie: I think a general issue here is that we do

tcmu_destroy_log()
then
tcmu_destroy_config()

so if during the log resources are freed and destroyed then someone does a update, this would leak and access freed resources.

I think we need to change the ordering or add some new log function that handles checking if the log has been torn down before trying to update it.

lxbsz:
Updated all. Thanks.

libtcmu_config.c
mikechristie: Hey, did you handle this last comment? We need to change the shutdown ordering, or do something so if logging is shutdown while the config update is going on, we do not race?

lxbsz:
@mikechristie, 
I think I have missed one old comment about the  shutdown ordering of destory_log and destory_config and updated it again.

Thanks.

mikechristie:
Sorry. I missed your update. I will fix up this patch and merge for you.

lxbsz:
Hey Mike,
Thanks, I resolved the conflicts, test and update it again :-)

mikechristie:
How about the attached patch?

[log-reset-dir.txt](https://github.com/open-iscsi/tcmu-runner/files/1977086/log-reset-dir.txt)

lxbsz:
> How about the attached patch?
> 
> log-reset-dir.txt
> 
> I have done zero testing.

Yeah, looks good.

lxbsz:
@mikechristie 
Update with you patch and with 3 fixes.
1, reset to str default values after the options being set
2, reset to boolean default values after the options being set
3, fix about:
```
diff --git a/libtcmu_log.c b/libtcmu_log.c
index b232d42..70fa28a 100644
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -727,14 +727,14 @@ int tcmu_logdir_resetup(char *log_dir_path)
        if (!tcmu_logbuf) {
                ret = -ESHUTDOWN;
                goto unlock;
-       }               
+       }
 
        if (!tcmu_logdir_create(log_dir_path, true)) {
                ret = -ENOENT;
                goto unlock;
        }
 
-       ret = tcmu_create_file_output(tcmu_get_log_level(), TCMU_LOG_FILENAME,
+       ret = tcmu_create_file_output(TCMU_LOG_DEBUG, TCMU_LOG_FILENAME,
                                      true);
        if (ret < 0)
                tcmu_err("Could not change log path to %s, ret:%d.\n",
```

libtcmu_config.c
```diff
@@ -123,12 +123,18 @@ do { \
...
-#define TCMU_PARSE_CFG_STR(cfg, key) \
+#define TCMU_PARSE_CFG_STR(cfg, key, def) \
 do { \
 	struct tcmu_conf_option *option; \
+	char buf[1024]; \
 	option = tcmu_get_option(#key); \
 	if (option) { \
-		cfg->key = strdup(option->opt_str); } \
+		cfg->key = strdup(option->opt_str); \
+		if (option->opt_str) \
```
mikechristie: Do you remember what TCMU_FREE_CFG_STR_KEY was for?

lxbsz: Yeah, there is a memory leak bug introduced ahead of this PR. I will fix it together with this.
Thanks,

lxbsz:
@mikechristie Please review. Thanks.

libtcmu_config.c
```diff
@@ -180,6 +180,7 @@ static void tcmu_conf_free_str_keys(struct tcmu_config *cfg)
 	 * For example:
 	 * TCMU_FREE_CFG_STR_KEY(cfg, 'STR KEY');
 	 */
+	 TCMU_FREE_CFG_STR_KEY(cfg, log_dir_path);
```
mikechristie: I think if you do the above then you need to fix tcmu_destroy_config:

```
        tcmu_conf_free_str_keys(cfg);
        if (cfg->log_dir_path)
                free(cfg->log_dir_path);
```
because tcmu_conf_free_str_keys call will delete cfg->log-dir_path and then the call above will try to delete the same pointer.

I actually thought the leak was if tcmu_conf_set_options is called twice. The first call would do:

`                cfg->key = strdup(option->opt_str); \`

but then the second call would see the cfg->key is set and we would end up allocating memory again without freeing and leak the old memory at cfg->key. I thought we needed a 

```
if (!cfg->key)
     cfg->key = strdup(option->opt_str); 
```

Issue #361, opened 2018-01-24
glfs: crash when filesize is wrong


KirbyDE:
OS: Debian Stretch
Kernel: 4.14.13-1~bpo9+1
TCMU-Runner: 1.3.0
GlusterFS: 3.13.2-1
rtslib-fb: 2.1.fb64

>tcmu-runner[527]: tcmu_glfs_open:521 glfs/test1.img: device size and backing size disagree: device 20971518 backing 20971520
tcmu-runner[527]: 2018-01-25 00:06:42.834 527 [ERROR] tcmu_glfs_open:521 glfs/test1.img: device size and backing size disagree: device 20971518 backing 20971520
kernel: [  661.290565] show_signal_msg: 1 callbacks suppressed
kernel: [  661.290568] tcmu-runner[527]: segfault at 7f4a53dc7b88 ip 00007f4ae071b2ea sp 00007ffcb66c8e80 error 6 in libglusterfs.so.0.0.1[7f4ae06b0000+f1000]
systemd[1]: tcmu-runner.service: Main process exited, code=killed, status=11/SEGV
systemd[1]: tcmu-runner.service: Unit entered failed state.
systemd[1]: tcmu-runner.service: Failed with result 'signal'.

mikechristie:
Adding the gluster developers.

@pkalever @pranithk

mikechristie:
@lxbsz  I guess this is your issue now :)

lxbsz:
Yes, sure it's mine now, and I will have a look at this later bacause I look at another gluster-block vs tcmu-runner's poor performance problem.

The bug is :https://bugzilla.redhat.com/show_bug.cgi?id=1535780, and BTW any idea about this bug ?

lxbsz:
I will debug this from today.

lxbsz:
This has been fixed, please see: https://github.com/open-iscsi/tcmu-runner/pull/371

PR #371, opened 2018-02-21
glfs: fix crash for the error path

lxbsz:
From the glfs.h we can know that:

"glfs_fini() must be called after all operations on glfs_t is
finished."

Or it will crash, like here.

Fix issue #361.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -492,25 +492,25 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
 
 	config = tcmu_get_path(dev);
 	if (!config) {
-		goto fail;
+		goto free_glsp;
 	}
 
 	gfsp->fs = tcmu_create_glfs_object(dev, config, &gfsp->hosts);
 	if (!gfsp->fs) {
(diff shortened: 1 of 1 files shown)


lxbsz:
https://github.com/open-iscsi/tcmu-runner/issues/361

mikechristie:
Looks ok to me.

@pkalever, @pranithk  and @lxbsz how do you guys want to handle maintainership type of duties for glfs going forward? In the past, I wanted pkalever or pranithk's review before merging glfs changes. Should I assume that if lxbsz is posting a patch it is already internally reviewed and ackd or that lxbsz is maintainer and ok to merge?

pranithk:
@mikechristie We are doing the review on github, so if one of us posts the patch, you can probably hear from one more person before merging it.

lxbsz:
@pkalever 

Yes you are quit right.

@mikechristie 
Since @pkalever's https://github.com/open-iscsi/tcmu-runner/pull/372 has already fix this at the same time, then this patch will be no needed.

PR #372, opened 2018-02-27
remove size checks & improve logging details

pkalever:
1. Let the size checks be handled by glusterfs tools
2. Improve the logging details

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -479,9 +479,7 @@ static char* tcmu_get_path( struct tcmu_device *dev)
 static int tcmu_glfs_open(struct tcmu_device *dev)
 {
 	struct glfs_state *gfsp;
-	int ret = 0;
 	char *config;
-	struct stat st;
 
 	gfsp = calloc(1, sizeof(*gfsp));
 	if (!gfsp)
@@ -507,30 +505,12 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
 		goto unref;
 	}
 
-	ret = glfs_lstat(gfsp->fs, gfsp->hosts->path, &st);
-	if (ret) {
-		tcmu_dev_err(dev, "glfs_lstat failed: %m\n");
-		goto unref;
-	}
-
-	if (st.st_size != tcmu_get_device_size(dev)) {
-		tcmu_dev_err(dev,
-		             "device size and backing size disagree: "
-		             "device %lld backing %lld\n",
-		             tcmu_get_device_size(dev),
-		             (long long) st.st_size);
-		goto unref;
-	}
-
 	return 0;
 
 unref:
 	gluster_cache_refresh(gfsp->fs, tcmu_get_path(dev));
-
-fail:
-	if (gfsp->gfd)
-		glfs_close(gfsp->gfd);
 	gluster_free_server(&gfsp->hosts);
+fail:
 	free(gfsp);
 
 	return -EIO;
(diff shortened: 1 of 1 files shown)


pkalever:
review-request: @mikechristie @lxbsz @pranithk 

Thanks!

glfs.c
```diff
@@ -636,34 +616,15 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
 static int tcmu_glfs_reconfig(struct tcmu_device *dev,
                               struct tcmulib_cfg_info *cfg)
 {
 	switch (cfg->type) {
 	case TCMULIB_CFG_DEV_SIZE:
-		return tcmu_glfs_get_image_size(dev, cfg->data.dev_size);
+		/*
+		 * Let glusterfs tools handle size checks
+		 */
+		return 0;
```
bgly: What would be the benefit in having the size check done by GLFS tools and removing the reconfig of the size here? To me it just looks like you are losing a feature?

pkalever: Thanks for asking this @bgly :-)

1. When the glusterfs tools are already doing a size check before calling targetcli create/reconf then why do same stuff in tcmu-runner, feel it is just a overhead.

2. It will be difficult for the tools to switch between growing and shrinking the sizes while resize failures.

-> say if gluster tool has to perform resize (increase size) operation for a block configured on 3 nodes (HA 3). What we have to do in general is grow the size of the file and  issue targetcli /backstores/user:glfs/block set attribute dev_size=$NEWSIZE

-> But if the targetcli  command fails in one of the nodes, we will have to rollback, which involves shrinking of the size back to old size (dangerous)

But if there is no size check in tcmu-runner, first we run targetcli commands, only if all the HA number of nodes  return success then increase the size with confidence.

This is the flexibility that we get by having the size check at a layer above :-)

bgly: Thanks for the explanation! It looks good to me then.

glfs.c
```diff
@@ -495,13 +499,14 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
...
-		tcmu_dev_err(dev, "tcmu_create_glfs_object failed\n");
+		tcmu_dev_err(dev, "tcmu_create_glfs_object(config=%s) failed\n", config);
 		goto fail;
 	}
 
 	gfsp->gfd = glfs_open(gfsp->fs, gfsp->hosts->path, ALLOWED_BSOFLAGS);
 	if (!gfsp->gfd) {
-		tcmu_dev_err(dev, "glfs_open failed: %m\n");
+		tcmu_dev_err(dev, "glfs_open(vol=%s, file=%s) failed: %m\n",
+		             gfsp->hosts->volname, gfsp->hosts->path);
```
mikechristie: One FYI and one review comment.

The runner code normally follows the kernel coding style which is like above where the next line lines up with the "(". In your patch you seem to be changing that to where it lines up with the first quote, but here you used the runner style.

I do not really care about coding style in your own code as long as its not too far from the style used in the core code, but you should use one style.

glfs.c
```diff
@@ -429,26 +432,27 @@ static glfs_t* tcmu_create_glfs_object(struct tcmu_device *dev,
...
 		tcmu_dev_err(dev, "tcmu_make_absolute_logfile failed: %d\n",
-			     ret);
+		                  ret);
 		goto unref;
 	}
 
 	ret = glfs_set_logging(fs, logfilepath, TCMU_GLFS_DEBUG_LEVEL);
 	if (ret < 0) {
-		tcmu_dev_err(dev, "glfs_set_logging failed: %m\n");
+		tcmu_dev_err(dev, "glfs_set_logging(vol=%s, path=%s) failed: %m\n");
```
mikechristie: Do you need to do something->volname and something->path to match the added %s's.

Also

glfs.c
```diff
@@ -479,9 +479,7 @@ static char* tcmu_get_path( struct tcmu_device *dev)
 static int tcmu_glfs_open(struct tcmu_device *dev)
 {
```
mikechristie: Patch looks ok to me.

Just a question. For glfs, if a request is sent that is past the end of the device what error code do you get in glfs_async_cbk?

glfs.c
pkalever: > I do not really care about coding style in your own code as long as its not too far from the style used in the core code, but you should use one style.

tcmu_dev_log(dev, "%s %s",
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;x, y);
I read it like x & y are passed to fill "%s %s"   or rather x & y are input to fill "%s %s"

But that may be my own perception, I'm making changes to align with runner style :-)

glfs.c
pkalever: I call this as a golden catch :-)

glfs.c
pkalever: I think EIO, but I will have to read the code to confirm. Will reply here after verifying tomorrow ?

pkalever:
@mikechristie updated the PR addressing your review comments.

Thanks!

lxbsz:
There are some lines are exceeding 80 chars, but most of them are unavoidable, because spliting the  strings into two lines is not a good idea, which will make the log string search becoming more difficult.

The code look good to me.
And just have a quick test which also works well.

Thanks,

glfs.c
pranithk: @mikechristie There won't be any error. Write request on the block device is converted to write request on a file on glusterfs, so write will be successful. May I know what is the behaviour you would like to see?

mikechristie: I was not looking for any specific behavior to be implemented. I was just wondering what will happen for if/when we start to add thin provisioning features on the SCSI side.

PR #373, opened 2018-02-28
log: bypass the log ring buffer for file and std outputs

lxbsz:
Make sure the whole logs are recorded to somewhere before main
loop termination (crashed, SIGINT, ...)

Fix bug : https://bugzilla.redhat.com/show_bug.cgi?id=1475288

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -59,6 +59,7 @@ struct log_output {
 	char *name;
 	void *data;
 	tcmu_log_destination dest;
+	bool bypass;
 };
 
 static int tcmu_log_level = TCMU_LOG_INFO;
@@ -155,12 +156,29 @@ static inline void rb_update_head(struct log_buf *logbuf)
 	logbuf->head = (logbuf->head + 1) % LOG_ENTRYS;
 }
 
+static void log_output(int pri, const char *msg, bool bypass)
+{
+	struct log_output *output;
+	char timestamp[TCMU_TIME_STRING_BUFLEN] = {0, };
+	int ret;
+
+	ret = time_string_now(timestamp);
+	if (ret < 0)
+		return;
+
+	darray_foreach (output, logbuf->outputs) {
+		if (output->bypass == bypass && pri <= output->priority) {
+			output->output_fn(pri, timestamp, msg, output->data);
+		}
+	}
+}
+
 static void
 log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	     int linenr, const char *fmt, va_list args)
 {
 	unsigned int head;
-	char *msg;
+	char *msg, buf[LOG_MSG_LEN];
 	int n;
 	struct tcmur_handler *rhandler;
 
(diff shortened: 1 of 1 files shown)


lxbsz:
@mikechristie @pranithk @pkalever 
Please review thanks very much.

pkalever:
Patch looks good to me.

I have seen this issue on one of our QE setups which are running tcmu-runner on a open-shift container platform, I will test this in that Environment and update the results.

Thanks!

libtcmu_log.c
```diff
@@ -193,6 +211,9 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 
 	vsnprintf(msg + n, LOG_MSG_LEN - n, fmt, args);
 
+	memcpy(buf, msg, LOG_MSG_LEN);
+	log_output(pri, buf, true);
```
mikechristie: I think you need to give me more info o this, because I'm not sure I got the bug.

It looks like we could log the message here, but then below we queue it to be logged again later for the same output type.

Also why do we want to do this for file and stdout outputs and also why do we need to always log both of them from the calling context?

lxbsz: Here we just set the file and stdout as bypassing the log RingBuffer, and each log outputing will output the messages to the log file and stdout directly only. 

But only for the syslog, we need the RingBuffer to cache the log messages and avoid the tcmu-runner daemon hangs. As we discussed  before when the syslogd daemon is died, then all the syslog callout will be hanged forever.

The bug we encountered is that when we cache all the log messages in RingBuffer first, and then flush it to all the output, when crash or other cases the very important log messages in RingBuffer may be lost.

After this patch we can make sure that at least we can find some important log messages in the log file.

mikechristie: I am going to merge it.

I am just a little worried about adding something that we cannot control in the main IO/error paths. I was trying to actually use things like mlock in these paths so they are more predictable. For example, to log a message the kernel might have to now write out pages to get memory for the write operation. That could hit other issues like slow disks we have to then block on just to write out a message.  This could happen for hundreds of IOs. I think we might hit issues with failover/failure timers on the initiator side if it has to then wait for extra issues like these, but I am just not sure that it is a issue and I think I am just being too paranoid.

lxbsz: Yes, while mostly this should be rare to happen(?) And the same time in product cases, the INFO(as default) log level should be used.

IMHO, currently most machines the slow disks shouldn't be a problem, if it happens this could be that the system memories are almost exhausted and it is swapping memories to disk. If so, not only the tcmu-runner daemon will be slowed down but also all the other tasks in the machine.

Did I understand your thoughts correctly ?

pkalever:
Have got a chance to test this today in a setup where we have notice problems before. Previously I used to not see any logs on load_module failure while attempting to run tcmu-runner in a debug mode. With this patch I can see the logs been flush as expected.

This works as expected.

Thanks @lxbsz

mikechristie:
Ok. Thanks.

PR #375, opened 2018-03-03
zbc: fix write emulation, change returned sense codes, add install step and more

dmitry-fomichev:
Hello,

I'd like to publish a few commits to ZBC file handler. Most of these are fixes that came about as a result of our internal testing.

Besides the changes listed in the title, this PR contains some minor fixes to tcmu-runner core.
One commit is a fix for a compile error (see the comment attached to the commit). Not sure if anyone else has seen this, but both Damien and me get this error on a fresh install of Fedora 27 unless a specific include is commented out.

Cheers,
Dmitry

### DIFF
--- a/file_zbc.c
+++ b/file_zbc.c
@@ -627,6 +627,7 @@ static int zbc_format_meta(struct zbc_dev *zdev)
 	struct zbc_dev_config *cfg = &zdev->cfg;
 	struct zbc_meta *meta;
 	struct zbc_zone *zone;
+	unsigned int nr_seq_zones;
 	__u64 lba = 0;
 	unsigned int i;
 	int ret;
@@ -653,8 +654,12 @@ static int zbc_format_meta(struct zbc_dev *zdev)
 	}
 
 	zdev->nr_open_zones = cfg->open_num;
-	if (zdev->nr_open_zones >= zdev->nr_zones)
-		zdev->nr_open_zones = zdev->nr_zones;
+	nr_seq_zones = zdev->nr_zones - zdev->nr_conv_zones;
+	if (zdev->nr_open_zones >= nr_seq_zones / 2) {
+		zdev->nr_open_zones = nr_seq_zones / 2;
+		if (!zdev->nr_open_zones)
+			zdev->nr_open_zones = 1;
+	}
 
 	tcmu_dev_dbg(zdev->dev, "Formatting...\n");
 	tcmu_dev_dbg(zdev->dev, "  Model: %s\n",
--- a/file_zbc.c
+++ b/file_zbc.c
@@ -12,7 +12,7 @@
  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
  * License for the specific language governing permissions and limitations
  * under the License.
-*/
+ */
 
 /*
  * ZBC device emulation with a file backstore.
@@ -258,7 +258,7 @@ struct zbc_dev_config {
  */
 struct zbc_dev {
 
-	struct tcmu_device 	*dev;
+	struct tcmu_device	*dev;
 
 	struct zbc_dev_config	cfg;
 
@@ -352,7 +352,7 @@ static char *zbc_parse_open(char *val, struct zbc_dev_config *cfg, char **msg)
 #define ZBC_PARAMS	5
 
 struct zbc_dev_config_param {
-	char 	*name;
+	char	*name;
 	char	*(*parse)(char *, struct zbc_dev_config *, char **);
 } zbc_params[ZBC_PARAMS] = {
 	{ "model-",	zbc_parse_model	},
@@ -1332,7 +1332,7 @@ static int zbc_report_zones(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	len = tcmu_iovec_length(iovec, iov_cnt);
 	lba = tcmu_get_lba(cdb);
 	zone = zbc_get_zone(zdev, lba, false);
-	while (lba < zdev->capacity && len >=64) {
+	while (lba < zdev->capacity && len >= 64) {
 
 		if (zbc_should_report_zone(zone, ro)) {
 
@@ -1351,7 +1351,7 @@ static int zbc_report_zones(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 			memcpy(&data[24], &val64, 8);
 
 			tcmu_memcpy_into_iovec(iovec, iov_cnt, data, 64);
-			len -=64;
+			len -= 64;
 		}
 
 		lba = zone->start + zone->len;
@@ -1398,8 +1398,6 @@ static void __zbc_close_imp_open_zone(struct zbc_dev *zdev)
 			return;
 		}
 	}
-
-	return;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


mikechristie:
@dmitry-fomichev @damien-lemoal 

How should maintainership work for file_zbc? Most patches have @damien-lemoal  signed off, but @dmitry-fomichev is doing the PR and dimitry does not have a signed off on the damien patches. Should I assume if damien's signed off is on a patch it is ok to be merged? Is it a co-maintainer type of situation where if dimitry does a PR then it does not need a signed off from damien?

libtcmu.c
```diff
@@ -969,17 +969,21 @@ struct tcmulib_cmd *tcmulib_get_next_command(struct tcmu_device *dev)
...
-				 * This should never happen so just drop cmd
+				 * This can only happen if a vendor-specific
+				 * command is received, so just drop cmd
 				 * for now instead of adding a lock in the
 				 * main IO path.
 				 */
-				break;
+				TCMU_UPDATE_DEV_TAIL(dev, mb, ent);
+				return NULL;
 			}
```
mikechristie: I need a little extra info on how this helps.

It looks like without the patch we hit the break, do TCMU_UPDATE_DEV_TAIL at the bottom of the loop and either return NULL like above or try to handle another command. Either way, the vendor command will be dropped. The initaitor side will timeout, send TMFs which LIO basically just sits on, and then the initiator will offline the device.

With the patch it looks like we will do the update, and return NULL right away. The command will still be dropped and the same thing will happen on the initaitor side.

Since I was completely wrong about this never happening we should probably fix this properly. How about the attached patch which is not tested. It will handle the vendor command case. For the memory allocation failure there is nothing we can do and its going to hang. We could add some locking around the parts of tcmulib_command_complete that interact with the ring and then put that in a new function and call it, but I'm not sure that is worth it.
[runner-force-vendor-cdb-fail.txt](https://github.com/open-iscsi/tcmu-runner/files/1782406/runner-force-vendor-cdb-fail.txt)

mikechristie: One more question.

How did you hit the vendor command case? Is it a normal case for some distro/OS or did you use something like sg utils to inject them?

mikechristie: Maybe it's just better to move the tcmur device mbox lock to a common place. Attached is a patch which should handle all cases for all handler types. It is not tested.

[runner-make-mbox-lock-common.txt](https://github.com/open-iscsi/tcmu-runner/files/1782485/runner-make-mbox-lock-common.txt)

target_core_user_local.h
```diff
@@ -4,7 +4,7 @@
 /* This header will be used by application too */
 
 #include <linux/types.h>
-#include <linux/uio.h>
+//#include <linux/uio.h>
```
mikechristie: Drop this patch from the set. We will work on it so it works with old and new setups in the other PR.

dmitry-fomichev: Sure, will do

libtcmu.c
dmitry-fomichev: Hi Mike,

As Damien mentioned, I was experimenting with a handler explicitly designed to handle some vendor-specific commands. I observed that with my patch TMF will not get stuck and the device is not offlined. That saved me a reboot, but, the best solution would be to just fail the command.

It should be possible to add the capability to handle vendor-specific commands to tcmu-runner by adding a new "get_cdb_length()" function to handler template allowing a particular backend handler to override the default tcmu_get_cdb_lba() function. I am not sure if such a feature is useful enough to implement.

Anyway, I tried your patches and runner-force-vendor-cdb-fail.txt works ok. The vendor command is failed right away. There is a slight problem - two different "command not supported" messages are logged, one for the actual opcode and the other for 0xc0, which might be a bit confusing for the user.

The second patch, runner-make-mbox-lock-common.txt, doesn't work right away in my environment. SCSI sd device does not appear in the system and the processing thread sleeps in ppoll() in the main loop. This happens after the first incoming command is completed. I can't see any obvious problem with the patch or any lockup happening. I am going to investigate further what is going on and will let you know. I think this patch is the best solution to this whole thing and also a nice cleanup.

dmitry-fomichev:
Hi Mike,
Sorry about not adding sign-offs to Damien's commits. Since it is better to have everyone to sign, I am going to redo the PR without the include file commit and with the signoffs.

libtcmu.c
mikechristie: > Agree on the second patch being best. I will debug it real quick here too.

> It should be possible to add the capability to handle vendor-specific commands to tcmu-runner by adding a new "get_cdb_length()" function to handler template allowing a particular backend handler to override the default tcmu_get_cdb_lba() function. I am not sure if such a feature is useful enough to implement.

I am ok with adding vendor command support too. The override idea is fine with me.

mikechristie: Ok.  I found the bug. It works for me because I somehow get multiple commands. If you were to only get one when this happens then we never called tcmulib_processing_complete and so the kernel is never notified of the completed failed command and it is never completed in the kernel.

Here is a updated patch.
[runner-make-mbox-lock-common-v2.txt](https://github.com/open-iscsi/tcmu-runner/files/1786438/runner-make-mbox-lock-common-v2.txt)

mikechristie: Hey,

I made one other tiny change to the v2 patch and made a PR

https://github.com/open-iscsi/tcmu-runner/pull/377

We can discuss this specific issue there.

dmitry-fomichev:
Hi Mike,

I have removed the uio.h include commit as well as the one about handling vendor-specific opcodes since these are being fixed elsewhere. I might give a shot to that idea of adding a template function entry for vendor-specific command support once your fix is in place...  I signed off all the remaining commits.

Regards,
Dmitry

file_zbc.c
```diff
@@ -2027,6 +2032,86 @@ static int zbc_read(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		/* Check conv -> seq and seq -> seq zone boundary crossing */
+		if (zone_type == 0)
+			zone_type = zone->type;
+		if (zone->type != zone_type ||
+		    (zbc_zone_seq_req(zone) &&
+		     lba + nr_lbas > zone->start + zone->len)) {
+			tcmu_dev_err(dev,
+				     "Write boundary violation lba %llu, "
+				     "xfer len %lu\n",
+				     lba, nr_lbas);
```
mikechristie: Just a minor nit. I am not strict on coding style as long as it does not get too different form the core code where we try to use the kernel coding style because we have to go back and forth between the target_core_user module and user space.

However, it seems in this patch we are adding a new style. I think normally in file_zbc we are not breaking up the string even if it goes over 80 chars, so we would do

tcmu_dev_err(dev, "Write boundary violation lba %llu, "xfer len %lu\n",

but here and some places below you add the new style that strictly follows 80 chars and breaks it up.

dmitry-fomichev: Sure, let's keep it consistent with the rest of the codebase. I have amended the code not to break up lines. Also ran checkpatch and fixed a few things it complained about.

mikechristie:
Thanks and sorry for the late comment that caused the extra repush.

PR #377, opened 2018-03-06
runner: fail commands with unsupported CDBs.

mikechristie:
This patch just properly handles unsupported CDBs by failing
them instead of hanging the entire device.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -440,6 +440,9 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 		tcmu_err("calloc failed in add_device\n");
 		return -ENOMEM;
 	}
+	ret = pthread_spin_init(&dev->lock, 0);
+	if (ret != 0)
+		goto err_free;
 
 	snprintf(dev->dev_name, sizeof(dev->dev_name), "%s", dev_name);
 
@@ -447,12 +450,12 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 	ptr = strchr(oldptr, '/');
 	if (!ptr) {
 		tcmu_err("invalid cfgstring\n");
-		goto err_free;
+		goto err_cleanup_lock;
 	}
 
 	if (strncmp(cfgstring, "tcm-user", ptr-oldptr)) {
 		tcmu_err("invalid cfgstring\n");
-		goto err_free;
+		goto err_cleanup_lock;
 	}
 
 	/* Get HBA name */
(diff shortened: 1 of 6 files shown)


libtcmu.c
```diff
@@ -966,20 +1022,23 @@ struct tcmulib_cmd *tcmulib_get_next_command(struct tcmu_device *dev)
...
 			cmd = malloc(sizeof(*cmd) + sizeof(*cmd->iovec) * ent->req.iov_cnt + cdb_len);
-			if (!cmd)
-				return NULL;
+			if (!cmd) {
+				tcmulib_entry_complete(dev, cmd_id, NULL,
+						       TCMU_NOT_HANDLED);
+				completed = true;
+				break;
+			}
 			cmd->cmd_id = ent->hdr.cmd_id;
```
dmitry-fomichev: Could just do `cmd->cmd_id = cmd_id` here

libtcmu.c
```diff
@@ -928,6 +937,52 @@ struct tcmur_handler *tcmu_get_runner_handler(struct tcmu_device *dev)
...
+		char *buf = ent->rsp.sense_buffer;
+
+		ent->rsp.scsi_status = SAM_STAT_CHECK_CONDITION;
+
+		buf[0] = 0x70;	/* fixed, current */
+		buf[2] = 0x5;	/* illegal request */
+		buf[7] = 0xa;
+		buf[12] = 0x20; /* ASC: invalid command operation code */
+		buf[13] = 0x0;	/* ASCQ: (none) */
+	} else if (result == SAM_STAT_CHECK_CONDITION) {
```
dmitry-fomichev: Here, the SCSI status is only set if it is CHECK CONDITION. This was the reason why the patch initially didn't work in my environment. With the following change, SCSI scan succeeds and sd device properly appears in the system.
> } else {
                if (result == SAM_STAT_CHECK_CONDITION)
                        memcpy(ent->rsp.sense_buffer, sense_buf,
                               TCMU_SENSE_BUFFERSIZE);
                ent->rsp.scsi_status = result;
        }

mikechristie: Nice catch. Thanks.

libtcmu.c
mikechristie: Will fix.

libtcmu.c
```diff
@@ -966,20 +1022,23 @@ struct tcmulib_cmd *tcmulib_get_next_command(struct tcmu_device *dev)
...
 				break;
 			}
 
 			/* Alloc memory for cmd itself, iovec and cdb */
 			cmd = malloc(sizeof(*cmd) + sizeof(*cmd->iovec) * ent->req.iov_cnt + cdb_len);
-			if (!cmd)
-				return NULL;
+			if (!cmd) {
+				tcmulib_entry_complete(dev, cmd_id, NULL,
+						       TCMU_NOT_HANDLED);
```
lxbsz: Shouldn't we return to the initiator the  key=HARDWARE_ERROR  and ascq=ASC_INTERNAL_TARGET_FAILURE? or something like that ?
The default will be : 
key ==  illegal request
ascq == invalid command operation code

mikechristie: You are right. Will fix. Thanks for reviewing.

PR #378, opened 2018-03-07
libtcmu: add tcmu_is_ring_reset_support support

lxbsz:
This will compatible with the old kernel version which could
not support the ring reset operations.

### DIFF
--- a/configfs.c
+++ b/configfs.c
@@ -247,6 +247,19 @@ int tcmu_set_cfgfs_ul(const char *path, unsigned long val)
 	return tcmu_set_cfgfs_str(path, buf, strlen(buf) + 1);
 }
 
+bool tcmu_cfgfs_file_is_supported(struct tcmu_device *dev, const char *name)
+{
+	char path[PATH_MAX];
+
+	snprintf(path, sizeof(path), CFGFS_CORE"/%s/%s/%s",
+		 dev->tcm_hba_name, dev->tcm_dev_name, name);
+
+	if (access(path, F_OK) == -1)
+		return false;
+
+	return true;
+}
+
 int tcmu_exec_cfgfs_dev_action(struct tcmu_device *dev, const char *name,
 			       unsigned long val)
 {
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -390,6 +390,11 @@ void tcmu_block_device(struct tcmu_device *dev)
 {
 	int rc;
 
+	if (!tcmu_cfgfs_file_is_supported(dev, "action")) {
+		tcmu_dev_warn(dev, "Kernel does not support the block_dev action.\n");
+		return;
+	}
+
 	/*
 	 * Afer we set block_dev=1 the kernel will have freed
 	 * cmds in the qfull queue and no new commands will be sent to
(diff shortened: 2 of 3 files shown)


libtcmu.c
```diff
@@ -390,6 +390,11 @@ void tcmu_block_device(struct tcmu_device *dev)
 {
 	int rc;
 
+	if (!tcmu_is_ring_reset_support(dev)) {
+		tcmu_dev_warn(dev, "reset ring is not supported!\n");
+		return;
+	}
+
```
mikechristie: Without this patch, don't we just log a message already?

lxbsz: Yes, it was. But it would log a message as an error which will be confusing.

libtcmu.c
```diff
@@ -506,9 +516,11 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
...
 		 * Force a retry of the outstanding commands.
 		 */
-		ret = tcmu_exec_cfgfs_dev_action(dev, "reset_ring", 1);
-		if (ret)
-			tcmu_dev_err(dev, "Could not reset ring %d.\n", ret);
+		if (tcmu_is_ring_reset_support(dev)) {
+			ret = tcmu_exec_cfgfs_dev_action(dev, "reset_ring", 1);
+			if (ret)
+				tcmu_dev_err(dev, "Could not reset ring %d.\n", ret);
+		}
```
lxbsz: But here would just fail the path.

mikechristie: I did not understand this comment.

libtcmu.c
mikechristie: Do you mean the "Could not open configfs to write attribute" message from tcmu_set_cfgfs_str was a logged as an error?

What is confusing that one message is a error level but the messages like "Kernel does not support the block_dev action" were warning level? If so, ok.

configfs.c
```diff
@@ -247,6 +247,19 @@ int tcmu_set_cfgfs_ul(const char *path, unsigned long val)
 	return tcmu_set_cfgfs_str(path, buf, strlen(buf) + 1);
 }
 
+bool tcmu_is_ring_reset_support(struct tcmu_device *dev)
```
mikechristie: Let's make this generic something like

tcmu_cfgfs_file_exists or tcmu_cfgfs_file_is_supported where one of the args is the file or dir or path to check.

libtcmu.c
```diff
@@ -390,6 +390,11 @@ void tcmu_block_device(struct tcmu_device *dev)
 {
 	int rc;
 
+	if (!tcmu_is_ring_reset_support(dev)) {
+		tcmu_dev_warn(dev, "reset ring is not supported!\n");
```
mikechristie: Let's make this more descriptive like below:

 "Kernel does not support the block_dev action.\n"

You can then also kill the explicit check for ENOENT since we already check it above now.

libtcmu.c
lxbsz: Yes.

libtcmu.c
lxbsz: I meant without this changes here, if the reset ring is not support, here will return an error which will fail the current operation. I think we should also just output one WARN message and skip this setting.

configfs.c
lxbsz: Yes, agree and will fix it

libtcmu.c
lxbsz: sure

lxbsz:
Updated

mikechristie:
Thanks.

PR #379, opened 2018-03-07
tcmu-runner: fix bug in memcpy_from_iovec and tcmu_memcpy_into_iovec

umuzhaohui:
iov_base and iov_len should not be upated.

### DIFF
--- a/api.c
+++ b/api.c
@@ -241,15 +241,18 @@ size_t tcmu_memcpy_into_iovec(
 	size_t copied = 0;
 
 	while (len && iov_cnt) {
-		size_t to_copy = min(iovec->iov_len, len);
+		void *iov_base = iovec->iov_base;
+		size_t iov_len = iovec->iov_len;
+		size_t to_copy = min(iov_len, len);
 
-		if (to_copy) {
-			memcpy(iovec->iov_base, src + copied, to_copy);
+		while (to_copy) {
+			memcpy(iov_base, src + copied, to_copy);
 
 			len -= to_copy;
 			copied += to_copy;
-			iovec->iov_base += to_copy;
-			iovec->iov_len -= to_copy;
+			iov_base += to_copy;
+			iov_len -= to_copy;
+			to_copy = min(iov_len, len);
 		}
 
 		iovec++;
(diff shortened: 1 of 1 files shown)


mikechristie:
What was the specific bug here? Was a iovec being updated then used in a place where it was assumed the iov base/len were not updated? What code path?

I ask because in the past, to handle partial completions callers would do something like:

```
while (total > 0) {
    ret = some_read_fn(buf)
    tcmu_memcpy_into_iovec(iov, iovcnt, buf, len)
   total -= ret;
}
```

I think we have a couple similar cases like in file_zbc.c:zbc_report_zones. That might be the last one though so we could just change that.

umuzhaohui:
I want to print the content of iovec for testing the command,  i find if the iov base/len were updated, i can not print the content  of Data-Out area.
The test code between "// begin yangzhaohui added for test" and " // end   yangzhaohui added for test" like this:
static int handle_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
{
        struct list_head group_list;
        struct tgt_port *port;
        int ret;

        list_head_init(&group_list);

        ret = tcmu_get_alua_grps(dev, &group_list);
        if (ret)
                return tcmu_set_sense_data(cmd->sense_buf, HARDWARE_ERROR,
                                           ASC_INTERNAL_TARGET_FAILURE, NULL);

        port = tcmu_get_enabled_port(&group_list);
        if (!port)
                tcmu_dev_dbg(dev, "no enabled ports found. Skipping ALUA support\n");

        ret = tcmu_emulate_inquiry(dev, port, cmd->cdb, cmd->iovec,
                                   cmd->iov_cnt, cmd->sense_buf);

        // begin yangzhaohui added for test
        tcmu_iovec_debug_info(dev, cmd);
        // end   yangzhaohui added for test
        return ret;
}

// begin yangzhaohui added for test
void tcmu_iovec_debug_info(struct tcmu_device *dev, const struct tcmulib_cmd *cmd)
{
        size_t length = tcmu_iovec_length(cmd->iovec, cmd->iov_cnt);
        uint8_t *bounce_buffer = malloc(length);
        if (!bounce_buffer) {
                tcmu_dev_err(dev, "out of memory.\n");
                return ;
        }

        tcmu_memcpy_from_iovec(bounce_buffer, length, cmd->iovec, cmd->iov_cnt);
        char *buf = malloc(length * 3 + 1);
        if (!buf) {
                tcmu_dev_err(dev, "out of memory.\n");
                free(bounce_buffer);
                return;
        }

        int i, n;
        for (i = 0, n = 0; i < length; i++) {
                n += sprintf(buf + n, "%x ", bounce_buffer[i]);
        }

        sprintf(buf + n, "\n");

        tcmu_dev_dbg(dev, "length: %u, terminal: %d\n", length, n);

        tcmu_dev_dbg(dev, buf);

        free(bounce_buffer);
        free(buf);
}
// end   yangzhaohui added for test

I review the code, i think  iov base/len should not be updated, so we can read content of iov correctly.

mikechristie:
> I review the code, i think iov base/len should not be updated, so we can read content of iov correctly.

Ok with me. Let's find all the places where the caller is relying on it to be updated like the one mentioned in file_zbc.c. We will just update that code.

PR #382, opened 2018-03-09
glfs resize fail recovery support

lxbsz:


glfs.c
```diff
@@ -626,17 +627,58 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
+		}
+
+		gfsp->fs = tcmu_create_glfs_object(dev, config, &gfsp->hosts);
+		if (!gfsp->fs) {
+			tcmu_dev_err(dev,
+			             "tcmu_create_glfs_object(config=%s) failed\n", config);
+			goto fail;
+		}
+
+		gfsp->gfd = glfs_open(gfsp->fs, gfsp->hosts->path, ALLOWED_BSOFLAGS);
```
pkalever: glfs_close() is missing. Will fix this part.

Rest I will test and update.

@lxbsz Thanks!

alua.c
```diff
@@ -295,8 +295,10 @@ int tcmu_get_alua_grps(struct tcmu_device *dev,
 	}
 
 	for (i = 0; i < n; i++) {
-		if (!strcmp(namelist[i]->d_name, "default_tg_pt_gp"))
-			continue;
+		if (!strcmp(namelist[i]->d_name, "default_tg_pt_gp")) {
+			if (failover_is_supported(dev))
+				continue;
+		}
```
mikechristie: Did you see the comment at the top of tcmu_get_alua_grps? Older kernels are pretty nasty.

The problem is older kernels barf at various places when trying to use ALUA, so if we blindly scan we can end up hanging the entire kernel at the worst case. We assume some tool has done some check kernel check and when it is able to setup a group we are able to access it. But we recently added a kernel file so it now tells us if it is supported, but there are some kernels that did not have the file, but ALU A is supported, so I guess take your poison :)

But we could safely do your patch if we check for the alua_supported bits. Using the helper function to check if a file exists from that other PR check for the file:

/sys/kernel/config/target/core/user_N/devname/attrib/alua_support

if it exists and the value is 1 then the above chunk of code will be ok.

Note: for RHGS RHEL 7.4 does not have this file but I think all the needed alua fix ups are in it. RHEL 7.5 has everything so you would be ok for that.

glfs.c
```diff
@@ -626,17 +627,58 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
 	default:
 		return -EOPNOTSUPP;
 	}
+
+out:
+	gluster_cache_refresh(gfsp->fs, tcmu_get_path(dev));
+	gluster_free_server(&gfsp->hosts);
+fail:
+	free(gfsp);
+
```
mikechristie: If IO is running then you could free this while the handler is accessing it.

glfs.c
```diff
@@ -626,17 +627,58 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
+notify:
+	if (ret) {
+		tcmu_notify_conn_lost(dev);
```
mikechristie: This function is supposed to do the right thing for the handler. It will wait for running IO, then call handler->close(), and then call handler->open until that succeeds. Above if we have done free(gfsp) already then I think that could do a double free when tcmu_glfs_close is called.

In that email thread when I made the comment about doing it from this context instead of adding the READ CAP callout I thought you just needed to do a glfs_lstat, so I thought you could just do
'''

if (glfs_lstat() == some net error like ENOTCOONN)
       tcmu_notify_conn_lost()
else
       if (st.st_size == cfg->data.dev_size)
                return ok
      else
               return mismatch

'''

If you need to reopen the device to resize it then you need to do block, then wait for aio and then flush (basically take the top of tcmu_acquire_dev_lock and stick in in a new function an call it).

glfs.c
lxbsz: Yes，right

alua.c
lxbsz: Yes, I saw that comments. So just one RFC here and asking for you advice here.

So I am thinking if we won't use the default alua, just create a new one, such as "glfstpg", will this be okay ?

mikechristie: > So I am thinking if we won't use the default alua, just create a new one, such as "glfstpg", will this be okay ?

Yes.

pkalever:
Hi @lxbsz,

I see error while starting tcmu-runner with this patch, could you fix this ?

2018-03-12 23:19:42.965 6263 [ERROR] open_handlers:158: Could not open handler at /usr/lib64/tcmu-runner/handler_glfs.so: /usr/lib64/tcmu-runner/handler_glfs.so: **undefined symbol: tcmur_set_pending_ua**

Thanks!

lxbsz:
@pkalever 
Has updated, you could just have try and test and the resize should be okay.  But  when adding tcmu-runner is restarting there will still have a problem when the backend size is not equal to the saveconfig.json's, the tcmu-runner will stuck

lxbsz:
Update again.

glfs.c
```diff
@@ -626,17 +627,35 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
+				ret = -1;
+			}
+		}
+		break;
 	case TCMULIB_CFG_DEV_CFGSTR:
 	case TCMULIB_CFG_WRITE_CACHE:
 	default:
 		return -EOPNOTSUPP;
 	}
+
```
mikechristie: The ret value should be a -Exyz value like -ENOTCONNN,-EINVAL, etc. The value will get propagated to the caller so it can have a better idea of what happened.

lxbsz: Will update.

lxbsz:
Updated.

pkalever:
@lxbsz The basic functionality is working as per expected i.e. disabling of tpg after reconfig failure and enabling tpg after reopen is successful.

Here are the related logs:

2018-03-14 20:33:34.086 1103 [DEBUG] tgt_port_grp_recovery_thread_fn:232: Disabling iscsi/iqn.2016-12.org.gluster-block:c328bd9e-1f2f-4050-ad52-1d6ac101e025/tpgt_3.
2018-03-14 20:33:34.086 1103 [INFO] tgt_port_grp_recovery_thread_fn:254: Disabled iscsi/iqn.2016-12.org.gluster-block:c328bd9e-1f2f-4050-ad52-1d6ac101e025/tpgt_3.
2018-03-14 20:33:34.086 1103 [DEBUG] __tcmu_reopen_dev:53 glfs/block: Waiting for outstanding commands to complete
2018-03-14 20:33:34.086 1103 [DEBUG] __tcmu_reopen_dev:84 glfs/block: Closing device.
2018-03-14 20:33:35.906 1103 [DEBUG] __tcmu_reopen_dev:97 glfs/block: Opening device.
2018-03-14 20:33:35.928 1103 [ERROR] tcmu_create_glfs_object:457 glfs/block: glfs_init(vol=sample) failed: Input/output error
2018-03-14 20:33:36.913 1103 [ERROR] tcmu_glfs_open:507 glfs/block: tcmu_create_glfs_object(config=sample@192.168.124.136/block-store/c328bd9e-1f2f-4050-ad52-1d6ac101e025) failed
2018-03-14 20:33:37.913 1103 [DEBUG] __tcmu_reopen_dev:97 glfs/block: Opening device.
2018-03-14 20:33:37.931 1103 [ERROR] tcmu_create_glfs_object:457 glfs/block: glfs_init(vol=sample) failed: Input/output error
2018-03-14 20:33:38.914 1103 [ERROR] tcmu_glfs_open:507 glfs/block: tcmu_create_glfs_object(config=sample@192.168.124.136/block-store/c328bd9e-1f2f-4050-ad52-1d6ac101e025) failed
2018-03-14 20:33:39.914 1103 [DEBUG] __tcmu_reopen_dev:97 glfs/block: Opening device.
2018-03-14 20:33:39.953 1103 [ERROR] tcmu_create_glfs_object:457 glfs/block: glfs_init(vol=sample) failed: Input/output error

...

2018-03-14 20:34:09.980 1103 [DEBUG] __tcmu_reopen_dev:97 glfs/block: Opening device.                                 
2018-03-14 20:34:10.009 1103 [ERROR] tcmu_create_glfs_object:457 glfs/block: glfs_init(vol=sample) failed: Input/output error
2018-03-14 20:34:11.982 1103 [ERROR] tcmu_glfs_open:507 glfs/block: tcmu_create_glfs_object(config=sample@192.168.124.136/block-store/c328bd9e-1f2f-4050-ad52-1d6ac101e025) failed
2018-03-14 20:34:12.982 1103 [DEBUG] __tcmu_reopen_dev:97 glfs/block: Opening device.
2018-03-14 20:34:14.496 1103 [INFO] tgt_port_grp_recovery_thread_fn:279: Enabled iscsi/iqn.2016-12.org.gluster-block:c328bd9e-1f2f-4050-ad52-1d6ac101e025/tpgt_3.


I will try to test this with more from the multipathing side sometime tomorrow and update.

configfs.c
```diff
@@ -115,6 +136,18 @@ char *tcmu_get_wwn(struct tcmu_device *dev)
 	return ret_buf;
 }
 
+int tcmu_set_device_size(struct tcmu_device *dev, bool with_nl)
```
mikechristie: Rename to tcmu_set_dev_size to match the other naming.

I think we can also drop the with_nl argument, because we probably never want to call into ourself and outside users doing the other model that does not use rtslib also do not want nl notifications into themself.

glfs.c
```diff
@@ -626,12 +666,30 @@ static int tcmu_glfs_write(struct tcmu_device *dev,
...
+			tcmu_dev_warn(dev, "glfs_lstat failed: %m\n");
+			tcmu_notify_conn_lost(dev);
+
+			/* Let the targetcli command return success */
+			ret = 0;
+		} else {
+			if (st.st_size != cfg->data.dev_size) {
+				tcmu_dev_err(dev,
+					     "device size and backing size disagree: "
+					     "device %lld backing %lld\n",
```
mikechristie: Did you want to do your more common one long string instead of breaking it up?

glfs.c
```diff
@@ -510,15 +515,50 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
...
-	return 0;
+	ret = glfs_lstat(gfsp->fs, gfsp->hosts->path, &st);
+	if (ret) {
+		tcmu_dev_warn(dev, "glfs_lstat failed: %m\n");
+		goto close;
+	}
+
+	dev_size = tcmu_get_dev_num_lbas(dev) * tcmu_get_dev_block_size(dev);
+	if (st.st_size != dev_size) {
+		if (!tcmu_dev_flags_test(dev, TCMU_DEV_FLAG_REOPENING)) {
```
mikechristie: We can just add a new reopen arg to the open callout. I think rbd will want something similar so it is going to be common.

glfs.c
```diff
@@ -510,15 +515,50 @@ static int tcmu_glfs_open(struct tcmu_device *dev)
...
+			      "device size and backing size disagree:device %lld backing %lld\n",
+			      dev_size, (long long) st.st_size);
+
+		ret = tcmu_update_num_lbas(dev, st.st_size);
+		if (!ret) {
+			ret = tcmu_set_device_size(dev, false);
+			if (ret) {
+				/* Rolling back */
+				tcmu_update_num_lbas(dev, dev_size);
+				goto close;
```
mikechristie: Let's just put the update_num_lbas, set_dev_size and roll back error handling, in a function like tcmu_dev_update_size().

Also in that new function drop the SIZE_CHANGES bits and just add a call to tcmur_set_pending_ua().

tcmur_device.c
```diff
@@ -86,17 +87,28 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, bool in_lock_thread)
 	pthread_mutex_lock(&rdev->state_lock);
 	rdev->flags &= ~TCMUR_DEV_FLAG_IS_OPEN;
 	ret = -EIO;
+	tcmu_dev_flags_set(dev, TCMU_DEV_FLAG_REOPENING);
 	while (ret != 0 && !(rdev->flags & TCMUR_DEV_FLAG_STOPPING)) {
 		pthread_mutex_unlock(&rdev->state_lock);
 
+		/* Avoid busy loop ? */
+		sleep(1);
```
mikechristie: Let's move the sleep to below the open call and and do it if ret != 0.

configfs.c
lxbsz: Yes, sure, will fix it.

glfs.c
lxbsz: But not sure whether there has some exist handlers are using this, like the qemu-tcmu.

lxbsz: Will this be a problem ? I it's okay I will update this.

glfs.c
lxbsz: Agree and will update it.

tcmur_device.c
lxbsz: Yes, will update it.

glfs.c
mikechristie: We are already not worrying about libtcmu/callout use and non tcmu-runner users. The lib was not being exported anymore.

For the next GPL friendly version of libtcmu we will handle these types of things.

lxbsz: Cool, I will update it.

glfs.c
lxbsz: BTW, How to deal with this issue ? 

Linking C executable consumer
libtcmu.so.2: undefined reference to `tcmur_set_pending_ua'
collect2: error: ld returned 1 exit status
make[2]: *** [consumer] Error 1
make[1]: *** [CMakeFiles/consumer.dir/all] Error 2
make: *** [all] Error 2

The tcmur_set_pending_ua is in tcmur_cmd_handler.c and we will use it in libtcmu.so.

mikechristie: You can just put the tcmu_dev_update_size in tcmur_device.c then add it to main-syms.txt for now.

Again, when the new lib is made we will sort all this out in a proper lib.

lxbsz:
@mikechristie  Update all. Thanks,

mikechristie:
It looks ok to me.

Are you guys done testing it on your side?

pkalever:
@mikechristie I'm planning to test this with HA target configuration and with some decent multipath configuration at the initiator side, but that should not stop this patch from merging.

I will update with my test results here later.

mikechristie:
Thanks.

Adding myself a note about this being merged before the unsigned long vs u64 dev_size fix ups so the changes in this patch need to be handled in the dev size patch.

PR #383, opened 2018-03-14
conifg: fix error ret

pengliang233:
Fix error ret in tcmu_cancel_config_thread.


mikechristie:
Thanks.

PR #388, opened 2018-03-15
main: fix error goto

pengliang233:
Fix error goto in dev_added.


mikechristie:
Thanks.

PR #389, opened 2018-03-15
rbd:fix parse cfgstring optional options

tangwenji:
When cfg two optional options exist at the same time, only the first option can resolve correctly。


mikechristie:
Nice. I guess I just all around messed up on that! Thanks for fixing multiple times!

PR #391, opened 2018-03-27
tcmu-runner: improve the logging always to get daemon's exit state

lxbsz:


main.c
```diff
@@ -190,6 +190,8 @@ static int open_handlers(void)
 
 static gboolean sighandler(gpointer user_data)
 {
+	tcmu_info("Signal killed!\n");
```
mikechristie: How about

"tcmu-runner has received signal! Shutting down.\n"

mikechristie: Ignore the above comment, because we would end up telling the user we are stopping twice.

Do we need a info log message for the above? We should eventually hit the "Exiting..." message below so the user is going to get a log message indicating that it shutdown eventually.

What above making the above log call just a dbg:

tcmu_dbg("Have received signal!\n")

lxbsz: Yes, actually it will log twice. And the Exiting should be what we needed.

Issue #395, opened 2018-04-06
discard/trim support for the block


pranithk:
Is it possible to get a discard/trim support for the block? gluster has the support for this so wondering what should be done in the tcmu layer to get this support.

@mikechristie @lxbsz @pkalever

pranithk:
@pkalever I actually opened it after seeing your tracker because there seems to be changed needed in both the layers as per the pdf. It looks like discard gives bad performance for some cases as per the pdf in the earlier comment. I also saw that the management infra generally has an option to tell the block to do discard or not. So in gluster-block we need to be able to configure the device to support discard and in tcmu-runner we need the changes to call out glfs_discard().

PR #396, opened 2018-04-06
glfs: add discard support

pkalever:
Fixes: #395

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -93,7 +93,8 @@ typedef struct glfs_cbk_cookie {
 	enum {
 		TCMU_GLFS_READ  = 1,
 		TCMU_GLFS_WRITE = 2,
-		TCMU_GLFS_FLUSH = 3
+		TCMU_GLFS_FLUSH = 3,
+		TCMU_GLFS_DISCARD = 4
 	} op;
 } glfs_cbk_cookie;
 
@@ -580,6 +581,7 @@ static void glfs_async_cbk(glfs_fd_t *fd, ssize_t ret, void *data)
 			break;
 		case TCMU_GLFS_WRITE:
 		case TCMU_GLFS_FLUSH:
+		case TCMU_GLFS_DISCARD:
 			ret =  tcmu_set_sense_data(cmd->sense_buf, MEDIUM_ERROR,
 			                           ASC_WRITE_ERROR, NULL);
 			break;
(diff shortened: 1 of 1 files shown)


pkalever:
please-review: @mikechristie @lxbsz @pranithk

glfs.c
```diff
@@ -580,6 +581,7 @@ static void glfs_async_cbk(glfs_fd_t *fd, ssize_t ret, void *data)
 			break;
 		case TCMU_GLFS_WRITE:
 		case TCMU_GLFS_FLUSH:
+		case TCMU_GLFS_DISCARD:
 			ret =  tcmu_set_sense_data(cmd->sense_buf, MEDIUM_ERROR,
 			                           ASC_WRITE_ERROR, NULL);
```
mikechristie: Do you guys have any specific limits on size of alignment? If so call tcmu_set_dev_opt_unmap_gran/tcmu_set_dev_unmap_gran_align in your handler's open() call to override the defaults.

Also, do you have any special error codes that can be returned for discard like if failed due to a resize operation going on? If so we would want to handle it here by returning a different sense code.

pranithk: @mikechristie glfs_discard, eventually calls fallocate with FALLOC_FL_PUNCH_HOLE on the brick, so you can specify any valid offset and length and it will do the right thing. Just adding more info from man page, just in case you have any insight about it based on that.
Deallocating file space
       Specifying  the  FALLOC_FL_PUNCH_HOLE  flag (available since Linux 2.6.38) in mode deallocates space (i.e., creates a hole) in the byte range starting at offset and continuing for
       len bytes.  Within the specified range, partial filesystem blocks are zeroed, and whole filesystem blocks are removed from the file.  After a  successful  call,  subsequent  reads
       from this range will return zeroes.

@mikechristie I am not sure what should be the behaviour with parallel operations. Resize operation in glfs, eventually calls either a truncate/zerofill (based on sparse or not, zerofill is a form of fallocate()). Both these operations take a lock on the file to make sure one operation happens after the other. Could you help us understand the case where we may want to give an error for parallel operations?

mikechristie: For shrink, the resize operation and discard be sent in parallel from runner's perspective, so the discard could by pass the offset and len checks. If glfs.c sent the discard after you did the resize, then would xfs's fallocate call send an error like EFBIG?

We are not supporting shrink in rbd.c so we never handled it in there.

For the zero'ing issue offlist, I think you will be ok.

pranithk: @mikechristie I did some tests with fallocate -p with offset larger then size of the file and it kept succeeding. Let me do more tests and confirm.

mikechristie: Hey,

So was this patch good to go on your side as far as testing goes?

pranithk: @mikechristie Sorry for the delay. Was busy with some other work. Got the results now.

22:16:18 :( ⚡ ls -l /mnt/r3/
total 0
-rw-r--r--. 1 root root 0 May 17 22:15 a

Got the strace and saw that fallocate with offset, length 1024 which is greater than the size of the file (i.e. 0) is succeeding. So we are good to go here

strace.8444:fallocate(19, FALLOC_FL_KEEP_SIZE|FALLOC_FL_PUNCH_HOLE, 1024, 1024) = 0 <0.000011>

mikechristie:
I fixed the conflicts in the attached patch.


[rebase-396.txt](https://github.com/open-iscsi/tcmu-runner/files/2054963/rebase-396.txt)

mikechristie:
@lxbsz you guys were ok with my updated patch above right?

lxbsz:
@mikechristie 
Yes, before I tested it and It looks good to me.

Issue #397, opened 2018-04-11
Dynamic config reloading doesn't work when trying to comment the "loglevel=5" line out


lxbsz:
It should be set the loglevel back to INFO as default.

lxbsz:
https://github.com/open-iscsi/tcmu-runner/issues/397

lxbsz:
https://github.com/open-iscsi/tcmu-runner/pull/398

mikechristie:
#398 is merged. Closing issue. Thanks.

PR #398, opened 2018-04-11
config: reset to default values after the options being set

lxbsz:


lxbsz:
Fixing https://github.com/open-iscsi/tcmu-runner/issues/397

mikechristie:
Thanks.

PR #401, opened 2018-04-17
tcmur: fix ret value in tcmur_dev_update_size

pkalever:

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -2489,7 +2489,8 @@ int tcmur_dev_update_size(struct tcmu_device *dev, unsigned long new_size)
 
 	old_size = tcmu_get_dev_num_lbas(dev) * tcmu_get_dev_block_size(dev);
 
-	if (tcmu_update_num_lbas(dev, new_size)) {
+	ret = tcmu_update_num_lbas(dev, new_size);
+	if (!ret) {
 		ret = tcmu_set_dev_size(dev);
 		if (ret)
 			tcmu_update_num_lbas(dev, old_size); /* Rolling back */


tcmur_cmd_handler.c
```diff
@@ -2485,7 +2485,7 @@ static int handle_try_passthrough(struct tcmu_device *dev,
 int tcmur_dev_update_size(struct tcmu_device *dev, unsigned long new_size)
 {
 	unsigned long old_size;
-	int ret;
+	int ret = -1;
```
mikechristie: For errors we want to return a -errno value, so you could just do

ret = tcmu_update_num_lbas(dev, new_size));

However, I think that function is completely wrong or I'm not sure what we are doing.

tcmu_update_num_lbas returns -errno on failure and 0 on success, so here

```
        if (tcmu_update_num_lbas(dev, new_size)) {
                ret = tcmu_set_dev_size(dev);
```

I think instead of the above we wanted to do:

```
        if (!tcmu_update_num_lbas(dev, new_size)) {
                ret = tcmu_set_dev_size(dev);
```

where we call tcmu_set_dev_size if tcmu_update_num_lbas returns 0/success.

@lxbsz, am I right?

lxbsz: Yeah, right.

        ret = tcmu_update_num_lbas(dev, new_size);
        if (!ret) {
                ret = tcmu_set_dev_size(dev);
                ...
        }

It should be something like this.

pkalever:
@mikechristie good catch, thanks.
@lxbsz thanks.

Updated the PR with the suggestions.

lxbsz:
It looks good to me.

mikechristie:
Thanks.

PR #402, opened 2018-04-19
tcmu-runner: add reset netlink support

lxbsz:
Will do the netlink reset when starting and just before it could
receive and handle the netlink requests from kernel space.

### DIFF
--- a/configfs.c
+++ b/configfs.c
@@ -72,6 +72,93 @@ int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val
 	return tcmu_set_cfgfs_str(path, buf, strlen(buf) + 1);
 }
 
+static bool tcmu_cfgfs_mod_param_is_supported(const char *name)
+{
+	char path[PATH_MAX];
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/%s", name);
+
+	if (access(path, F_OK) == -1)
+		return false;
+
+	return true;
+}
+
+void tcmu_block_netlink(void)
+{
+	char path[PATH_MAX];
+	int rc;
+
+	if (!tcmu_cfgfs_mod_param_is_supported("block_netlink")) {
+		tcmu_warn("Kernel does not support blocking netlink.\n");
+		return;
+	}
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/block_netlink");
+
+	tcmu_dbg("blocking netlink\n");
+	rc = tcmu_set_cfgfs_ul(path, 1);
+	if (rc) {
+		tcmu_warn("Could not block netlink %d.\n", rc);
+		return;
+	}
+	tcmu_dbg("block netlink done\n")
+}
+
+void tcmu_unblock_netlink(void)
+{
+	char path[PATH_MAX];
+	int rc;
+
+	if (!tcmu_cfgfs_mod_param_is_supported("block_netlink")) {
+		tcmu_warn("Kernel does not support unblocking netlink.\n");
+		return;
+	}
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/block_netlink");
+
+	tcmu_dbg("unblocking netlink\n");
+	rc = tcmu_set_cfgfs_ul(path, 0);
+	if (rc) {
+		tcmu_warn("Could not unblock netlink %d.\n", rc);
+		return;
+	}
+	tcmu_dbg("unblock netlink done\n")
+}
+
+/*
+ * Usually this will be used when the daemon is starting just before
+ * it could receive and handle the kernel netlink requests.
+ *
+ * It will reset all the pending netlink msg, of which the reply maybe
+ * lost for some reason, such as the userspace dameon crashed just before
+ * it could reply to it, in kernel space.
+ *
+ * This must be called after blocking the netlink, and after this unblocking
+ * is a must.
+ */
+void tcmu_reset_netlink(void)
+{
+	char path[PATH_MAX];
+	int rc;
+
+	if (!tcmu_cfgfs_mod_param_is_supported("reset_netlink")) {
+		tcmu_warn("Kernel does not support reseting netlink.\n");
+		return;
+	}
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/reset_netlink");
+
+	tcmu_dbg("reseting netlink\n");
+	rc = tcmu_set_cfgfs_ul(path, 1);
+	if (rc) {
+		tcmu_warn("Could not reset netlink: %d\n", rc);
+		return;
+	}
+
+	tcmu_dbg("reset netlink done\n")
+}
+
 /*
  * Return a string that contains the device's WWN, or NULL.
  *
--- a/main.c
+++ b/main.c
@@ -486,11 +486,12 @@ static int load_our_module(void)
 			} else {
 				tcmu_info("no modules directory '/lib/modules/%s', checking module target_core_user entry in '/sys/modules/'\n",
 					  u.release);
-				ret = stat("/sys/module/target_core_user", &sb);
+				ret = stat(CFGFS_TARGET_MOD, &sb);
 				if (!ret) {
 					tcmu_dbg("Module target_core_user already loaded\n");
 				} else {
-					tcmu_err("stat() on '/sys/module/target_core_user' failed: %m\n");
+					tcmu_err("stat() on '%s' failed: %m\n",
+						 CFGFS_TARGET_MOD);
 				}
 			}
 		} else {
--- a/main.c
+++ b/main.c
@@ -1078,6 +1078,10 @@ int main(int argc, char **argv)
 	}
 
 	tcmu_dbg("handler path: %s\n", handler_path);
+
+	tcmu_block_netlink();
+	tcmu_reset_netlink();
+
 	ret = open_handlers();
 	if (ret < 0) {
 		tcmu_err("couldn't open handlers\n");
@@ -1141,6 +1145,7 @@ int main(int argc, char **argv)
 				NULL  // user date free func
 		);
 
+	tcmu_unblock_netlink();
 	g_main_loop_run(loop);
 
 	tcmu_info("Exiting...\n");
@@ -1167,6 +1172,7 @@ err_tcmulib_close:
 err_free_handlers:
 	darray_free(handlers);
 close_fd:
+	tcmu_unblock_netlink();
 	lock_fd.l_type = F_UNLCK;
 	if (fcntl(fd, F_SETLK, &lock_fd) == -1) {
 		tcmu_err("fcntl(UNLCK) on lockfile %s failed: [%m]\n",
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -66,6 +66,7 @@ enum {
 #define CFGFS_CORE CFGFS_ROOT"/core"
 
 #define CFGFS_TARGET_MOD "/sys/module/target_core_user"
+#define CFGFS_MOD_PARAM CFGFS_TARGET_MOD"/parameters"
 
 /* Temporarily limit this to 32M */
 #define VPD_MAX_UNMAP_LBA_COUNT            (32 * 1024 * 1024)
@@ -157,6 +158,9 @@ int tcmu_set_dev_size(struct tcmu_device *dev);
 long long tcmu_get_dev_size(struct tcmu_device *dev);
 char *tcmu_get_wwn(struct tcmu_device *dev);
 int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val);
+void tcmu_reset_netlink(void);
+void tcmu_block_netlink(void);
+void tcmu_unblock_netlink(void);
 int tcmu_get_cdb_length(uint8_t *cdb);
 uint64_t tcmu_get_lba(uint8_t *cdb);
 uint32_t tcmu_get_xfer_length(uint8_t *cdb);
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -65,6 +65,8 @@ enum {
 #define CFGFS_ROOT "/sys/kernel/config/target"
 #define CFGFS_CORE CFGFS_ROOT"/core"
 
+#define CFGFS_TARGET_MOD "/sys/module/target_core_user"
+
 /* Temporarily limit this to 32M */
 #define VPD_MAX_UNMAP_LBA_COUNT            (32 * 1024 * 1024)
 #define VPD_MAX_UNMAP_BLOCK_DESC_COUNT     0x04


lxbsz:
https://www.spinics.net/lists/target-devel/msg16902.html

The kernel patch set has been applied.

configfs.c
```diff
@@ -72,6 +72,54 @@ int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val
...
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/block_netlink");
+
+	return tcmu_set_cfgfs_ul(path, val);
+}
+
+static void tcmu_block_netlink(void)
+{
+	tcmu_set_block_netlink(1);
+}
```
mikechristie: No need for the wrapper and we will probably never need it, so just move the insides of tcmu_set_block_netlink into this function.

Same with unblock below.

configfs.c
```diff
@@ -72,6 +72,54 @@ int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val
...
+static void tcmu_unblock_netlink(void)
+{
+	tcmu_set_block_netlink(0);
+}
+
+/*
+ * Usually this will be used when the daemon is starting just before
+ * it could receive and handle the kernel netlink requests.
+ *
+ * I will reset all the pending netlink msg, of which the reply maybe
```
mikechristie: Change "I" to "It".

configfs.c
```diff
@@ -72,6 +72,54 @@ int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val
...
+ * lost for some reason(such as the userspace dameon crashed just before
```
mikechristie: No need for the ()

main.c
```diff
@@ -1077,6 +1077,9 @@ int main(int argc, char **argv)
 	}
 
 	tcmu_dbg("handler path: %s\n", handler_path);
+
+	tcmu_reset_netlink();
```
mikechristie: I think we need to do this in 2 or 3 stages for the restart case, because we could have apps sending requests while tcmu-runner is reinitializing. If we block, reset and unblock here, apps could send new requests after the unblock and they would be stuck because we have not opened the nl socket yet.

So block nl here first then unblock when the nl socket is setup and the daemon is fully ready to handle nl requests. I think that should be right before we call g_main_loop_run. We would then want to run the reset nl call any time between the block and unblock.

pkalever:
@lxbsz @mikechristie Thanks for all the work involved in getting changes at the kernel land.

This patch looks to me, I will test this later.

Thanks!

lxbsz:
@pkalever @mikechristie 
Updated it with debugging info.

lxbsz:
@mikechristie @pkalever 

Updated it again, here will set the reset netlink to 2 stages:
1, block_netlink and then reset_netlink before the netlink setup right away.
2, unblock_netlink before g_main_loop_run().

because we cannot put the reset_netlink after tcmulib_initialize() --> open_devices() --> add_device(), which will call tcmu_block_device() --> tcmu_cfgfs_file_is_supported() --> "/sys/kernel/config/target/core/user_XX/" and will introduce another stuck. 

This is because after the old tcmu-runner process was crashed or killed, the destroy_device is still in D state for tcmu_destroy_device() --> tcmu_netlink_event() --> wait_for_completion(). So if we come to "/sys/kernel/config/target/core/user_XX/" here again, it will dead lock in VFS. 

So we must reset the netlink and clean the old spending netlink events before any accessing to "/sys/kernel/config/target/core/user_XX/".

Thanks

configfs.c
```diff
@@ -72,6 +72,91 @@ int tcmu_set_control(struct tcmu_device *dev, const char *key, unsigned long val
...
+		tcmu_warn("Kernel does not support reseting netlink.\n");
+		return;
+	}
+
+	snprintf(path, sizeof(path), CFGFS_MOD_PARAM"/reset_netlink");
+
+	tcmu_dbg("reseting netlink\n");
+	rc = tcmu_set_cfgfs_ul(path, 1);
+	if (rc)
+		tcmu_warn("Could not reset netlink: %d\n", rc);
```
pkalever: Should we return here ?

main.c
```diff
@@ -1077,6 +1077,10 @@ int main(int argc, char **argv)
 	}
 
 	tcmu_dbg("handler path: %s\n", handler_path);
+
+	tcmu_block_netlink();
+	tcmu_reset_netlink();
```
pkalever: May be tcmu_reset_netlink() should be executed only if tcmu_block_netlink() is successful ?

configfs.c
lxbsz: Yes, right.

main.c
lxbsz: It will fail too if the block_netlink is failed, so here IMO is not a must.

pkalever: I was thinking of a case where tcmu_block_netlink() failed in tcmu_set_cfgfs_ul->tcmu_set_cfgfs_str(), but I will leave that up to you :-)

lxbsz: Yeah, for both I am okay, currently this just as the reset ring does by @mikechristie.

main.c
```diff
@@ -1140,6 +1144,7 @@ int main(int argc, char **argv)
 				NULL  // user date free func
 		);
 
+	tcmu_unblock_netlink();
```
pkalever: Please add this at the 'close_fd:' label to unblock netlink at failures and other abnormal terminations due to signals.

lxbsz: Good catch, fixed.

lxbsz:
There is a crash problem by doing:
1, pkill -9 tcmu-runner
2, then the gluster-block-target service will be terminated by systemd which will call 'targetctl clear'
3, if no IOs going on, the 'targetctl clear' will stuck in tcmu_destroy_device() --> tcmu_netlink_event() --> wait_for_completion(), this will be okay when 'systemctl restart tcmu-runner'
4, or if there is any IO going on, the 'targetctl clear' will stuck in iscsit_free_session() --> iscsit_cause_connection_reinstatement() firstly, and then after the reset ring cleaning it and then goto stuck in STEP 3.  But in tcmu-runner the reset netlink must ahead of reset ring due to the reason from Comment 23. But the reset ring is done in the open_device(), that's to say in userspace the device is opened and then it will be deleted in kernel space.

And the following upstream kernel patches and its related pathces should be backported to RHEL 7.6 to fix the crash:

https://patchwork.kernel.org/patch/2875871/
https://patchwork.kernel.org/patch/10392845/

mikechristie:
Let's discuss this here instead of a closed bz. Here is what I wrote before:

> https://patchwork.kernel.org/patch/10392845/

This patch is bad. It adds multiple regressions:

https://lkml.org/lkml/2018/6/20/689

> https://patchwork.kernel.org/patch/2875871/

This patch sort of adds a regression. For drivers like tcmu where the parent is not removed every time (it is tied to a module or a pci device) then the memory is not freed when the uio device is unregistered like it was before:

https://lkml.org/lkml/2017/6/14/30

So for tcmu where the uio device is tied to the lun and you would commonly add/remove luns a lot but pretty much never unload the module, then it would look like a memory leak.

For all these issues we need to properly fix the lifetime management to handle where userspace cannot be shutdown cleanly like Brian was working on here:

https://patchwork.kernel.org/patch/6057431/

main.c
mikechristie: It's probably fine. In the unsupported case you get an extra warning about both block and reset. For the case where the write  fails for the block then best case we avoid races and it works ok for the reset and worst case someone is doing a rtstlib call at the same time and we race and you are in the same hang as before.

mikechristie:
One question.

If you do not have these patches or at least this one

https://patchwork.kernel.org/patch/2875871/

Does it still work ok in your updated patch in this PR where you make sure netlink is flushed first? It seems like we would avoid the case where we open a device while it is also being destroyed/unregistered/freed.

lxbsz:
> One question.
> 
> If you do not have these patches or at least this one
> 
> https://patchwork.kernel.org/patch/2875871/
> 
> Does it still work ok in your updated patch in this PR where you make sure netlink is flushed first? It seems like we would avoid the case where we open a device while it is also being destroyed/unregistered/freed.

There are two stuck case:
[1], If there has any uncompleted IO or cmd, 'targetctl clear' process will stuck in iscsit_free_session() --> iscsit_cause_connection_reinstatement() first, then after all the IOs or cmds completed, it will stuck in [2].

[2], If there has no uncompleted IOs or cmds, 'targetctl clear' process will stuck in tcmu_destroy_device() --> tcmu_netlink_event() --> wait_for_completion().

If we are in the [1] case, no matter the netlink flush option is exist or not, it still works well because the reset ring will wake up the 'targetctl clear' process and then tcmu-runner daemon could process the followed deletion netlink event.

If we are in the [2] case, the netlink flush is a must before reset ring.

Before all my test case was concentrated on [2] case, and since @pkalever hit the [1] case and then use the [1] case I hit the crash problem in RHEL, but the upstream code works well.  So this is why we need the uio fixing patches here.

Thanks

lxbsz:
> Let's discuss this here instead of a closed bz. Here is what I wrote before:
> 
> https://patchwork.kernel.org/patch/10392845/
> 
> This patch is bad. It adds multiple regressions:
> 
> https://lkml.org/lkml/2018/6/20/689
> 
> https://patchwork.kernel.org/patch/2875871/
> 
> This patch sort of adds a regression. For drivers like tcmu where the parent is not removed every time (it is tied to a module or a pci device) then the memory is not freed when the uio device is unregistered like it was before:
> 
> https://lkml.org/lkml/2017/6/14/30
> 
> So for tcmu where the uio device is tied to the lun and you would commonly add/remove luns a lot but pretty much never unload the module, then it would look like a memory leak.
> 
> For all these issues we need to properly fix the lifetime management to handle where userspace cannot be shutdown cleanly like Brian was working on here:
> 
> https://patchwork.kernel.org/patch/6057431/

The crash will happen only in the [1] stuck case in the previous comment:
1), 'targetctl clear' process will stuck in iscsit_free_session() --> iscsit_cause_connection_reinstatement().
2), block netlink --> flush netlink --> reset ring in tcmu-runner daemon
3), just before reset ring option, the tcmu-runner daemon will open the /dev/uioX device, because the target device hasn't been deleted yet.
4), at the same time 'targetctl clear' process was already woken up, and then tried to delete the target device, but in step 2) the tcmu-runner has already blocked the netlink, so the deletion in kernel space just returns without successfully notifying to tcmu-runner daemon and then deletes target device's UIO resources in kernel space, but the tcmu-runner in userspace still accessing that.

The kernel code:
```
1626 static int tcmu_init_genl_cmd_reply(struct tcmu_dev *udev, int cmd)
1627 {
...
1636         mutex_lock(&tcmu_nl_cmd_mutex);
1637 
1638         if (tcmu_netlink_blocked) {                                                                                                                   
1639                 mutex_unlock(&tcmu_nl_cmd_mutex);
1640                 pr_warn("Failing nl cmd %d on %s. Interface is blocked.\n", cmd,
1641                         udev->name);
1642                 return -EAGAIN;
1643         }
...
 
1924 static void tcmu_destroy_device(struct se_device *dev)
1925 {
1926         struct tcmu_dev *udev = TCMU_DEV(dev);
1927 
1928         del_timer_sync(&udev->cmd_timer);
1929         del_timer_sync(&udev->qfull_timer);
1930 
1931         mutex_lock(&root_udev_mutex);
1932         list_del(&udev->node);
1933         mutex_unlock(&root_udev_mutex);
1934 
1935         tcmu_send_dev_remove_event(udev);                                                                                                             
1936 
1937         uio_unregister_device(&udev->uio_info);
1938 
1939         /* release ref from configure */
1940         kref_put(&udev->kref, tcmu_dev_kref_release);
1941 }

```
5), then the kernel crashes.

Actually, only this https://patchwork.kernel.org/patch/2875871/ patch could fix the crash problem.

The related code is:
```
@@ -864,13 +860,9 @@  void uio_unregister_device(struct uio_info *info)

 	uio_free_minor(idev);

-	if (info->irq && (info->irq != UIO_IRQ_CUSTOM))
-		free_irq(info->irq, idev);
-
 	uio_dev_del_attributes(idev);

 	device_destroy(&uio_class, MKDEV(uio_major, idev->minor));
-	kfree(idev);

 	return;
 }
```

I have checked the upstream code, for that regression there still hasn't any fix yet, right ?

So currently, we only need the following two patches:
https://patchwork.kernel.org/patch/2875871/
https://lkml.org/lkml/2015/3/19/665

With these two patches I have test for a whole night in RHEL and it works well, no any crash any more.

Thanks,

lxbsz:
> > https://patchwork.kernel.org/patch/10392845/
> 
> This patch is bad. It adds multiple regressions:
> 
> > https://lkml.org/lkml/2018/6/20/689

@mikechristie To be simple, if we switch the spin lock to mutex ? Is that okay ?

Thanks,

mikechristie:
> @mikechristie To be simple, if we switch the spin lock to mutex ? Is that okay ?

I think so. It works for me and it seems we always have proper context to block.

> https://patchwork.kernel.org/patch/2875871/

For that patch did you see my comment about the possible mem leak? Were you going to fix that too or because it is binding resources to the parent the reason that we are not crashing?

Also nice debugging!

lxbsz:
> > @mikechristie To be simple, if we switch the spin lock to mutex ? Is that okay ?
> 
> I think so. It works for me and it seems we always have proper context to block.

But the lock is protecting the idev->info struct data, and this data will also be used in IRQ routine and the mutex lock couldn't be used in IRQ routing too. So there maybe much more work todo.

I have post the related patches to the kernel list, please help review whenever you are availiable.

> 
> > https://patchwork.kernel.org/patch/2875871/
> 
> For that patch did you see my comment about the possible mem leak? 

Yeah, only with this patch there has mem leak as you mentioned before.

> Were you going to fix that too or because it is binding resources to the parent the reason that we are not crashing?
> 
If only with this patch there is no crash but with memory leak exist.
If we sync up to the upstream code, there will introduce new crash, and I have found the root cause.

Will fix it.

Thanks Mike.

> Also nice debugging!

mikechristie:
>>> https://patchwork.kernel.org/patch/2875871/
>> For that patch did you see my comment about the possible mem leak?
> Yeah, only with this patch there has mem leak as you mentioned before.

Ignore my comment. It is fixed upstream. I think I was just looking at the wrong RHEL patchset. I see in the current RHEL V3 where you brought in everything you got the fix for it too.

main.c
```diff
@@ -486,11 +486,11 @@ static int load_our_module(void)
 			} else {
 				tcmu_info("no modules directory '/lib/modules/%s', checking module target_core_user entry in '/sys/modules/'\n",
 					  u.release);
-				ret = stat("/sys/module/target_core_user", &sb);
+				ret = stat(CFGFS_TARGET_MOD, &sb);
 				if (!ret) {
 					tcmu_dbg("Module target_core_user already loaded\n");
 				} else {
-					tcmu_err("stat() on '/sys/module/target_core_user' failed: %m\n");
+					tcmu_err("stat() on '%s' failed: %m\n", CFGFS_TARGET_MOD);
```
mikechristie: Xiubo, you guys have completed testing this and all the kernel patches right?

Patch looks good to me. Just had this minor not.

Put CFGFS_TARGET_MOD on a new line.

lxbsz: Yes, we have completed testing this. Will fix it.

Thanks,

lxbsz:
@mikechristie 
Updated done.

PR #403, opened 2018-04-20
 tcmu-runner: return the correct scsi proto status to kernel

lxbsz:
The SAM_STAT_TASK_SET_FULL status code is returned when the logical
unit lacks the resources to accept a received task from an I_T nexus.

For the other errors of the backend storage, we will return NOT READY
status instead of ILLEGAL REQUEST, because returning ILLEGAL REQUEST
would cause immediate IO errors on some initiators, Returning NOT
READY instead means the operations will be retried a finite number of
times and we can survive intermittent errors.

### DIFF
--- a/api.c
+++ b/api.c
@@ -1169,7 +1169,9 @@ int tcmu_emulate_start_stop(struct tcmu_device *dev, uint8_t *cdb)
 #define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 1)
 #define CDB_FIX_BYTES 64 /* 64 bytes for default */
 #define CDB_FIX_SIZE CDB_TO_BUF_SIZE(CDB_FIX_BYTES)
-void tcmu_cdb_debug_info(struct tcmu_device *dev, const struct tcmulib_cmd *cmd)
+void tcmu_print_cdb_info(struct tcmu_device *dev,
+			 const struct tcmulib_cmd *cmd,
+			 const char *info)
 {
 	int i, n, bytes;
 	char fix[CDB_FIX_SIZE], *buf;
@@ -1191,9 +1193,17 @@ void tcmu_cdb_debug_info(struct tcmu_device *dev, const struct tcmulib_cmd *cmd)
 	for (i = 0, n = 0; i < bytes; i++) {
 		n += sprintf(buf + n, "%x ", cmd->cdb[i]);
 	}
+
+	if (info)
+		n += sprintf(buf + n, "%s", info);
+
 	sprintf(buf + n, "\n");
 
-	tcmu_dev_dbg_scsi_cmd(dev, buf);
+	if (info) {
+		tcmu_dev_warn(dev, buf);
+	} else {
+		tcmu_dev_dbg_scsi_cmd(dev, buf);
+	}
 
 	if (bytes > CDB_FIX_SIZE)
 		free(buf);
--- a/libtcmu_common.h
+++ b/libtcmu_common.h
@@ -192,7 +192,7 @@ int tcmu_emulate_mode_sense(struct tcmu_device *dev, uint8_t *cdb,
 int tcmu_emulate_mode_select(struct tcmu_device *dev, uint8_t *cdb,
 			     struct iovec *iovec, size_t iov_cnt);
 /* SCSI helpers */
-void tcmu_cdb_debug_info(struct tcmu_device *dev, const struct tcmulib_cmd *cmd);
+void tcmu_print_cdb_info(struct tcmu_device *dev, const struct tcmulib_cmd *cmd, const char *info);
 
 #ifdef __cplusplus
 }
--- a/main.c
+++ b/main.c
@@ -624,7 +624,7 @@ static void *tcmur_cmdproc_thread(void *arg)
 		while (!dev_stopping && (cmd = tcmulib_get_next_command(dev)) != NULL) {
 
 			if (tcmu_get_log_level() == TCMU_LOG_DEBUG_SCSI_CMD)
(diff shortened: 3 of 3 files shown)


mikechristie:
This is just getting to nasty. Just define tcmu errors

enum {
      TCMU_CMD_COMPLETED,
      TCMU_CMD_ASYNC_QUEUED,
      TCMU_CMD_NOMEM,
      TCMU_CMD_NOT_SUPP,
.... etc
}

and just use that throughout runner, libtcmu and the handers. Then in the completion function convert that to the proper SCSI.

We then can just drop this mess of codes :)

glfs.c
```diff
@@ -599,6 +599,7 @@ static int tcmu_glfs_read(struct tcmu_device *dev,
 {
 	struct glfs_state *state = tcmu_get_dev_private(dev);
```
mikechristie: For the error code you are trying to change the behavior for did you test this by failing with a NOT_READY-  COMMUNICATION FAILURE error and did you see a retry? Was it using dm-multipath though?

For linux I think this is a hard error and the error is propagated upwards. It looks like it would only be retried if you are using dm-multipath because it is pass up as a generic IO failure.

In windows I think that it is a hard error and the disk will be offlined if we get it.

lxbsz: For the NOT_READY- COMMUNICATION FAILURE in linux there is no retry even using multipath.

 916                 case NOT_READY:
 917                         /* If the device is in the process of becoming
 918                          * ready, or has a temporary blockage, retry.
 919                          */
 920                         if (sshdr.asc == 0x04) {
 921                                 switch (sshdr.ascq) {
 922                                 case 0x01: /* becoming ready */
 923                                 case 0x04: /* format in progress */
 924                                 case 0x05: /* rebuild in progress */
 925                                 case 0x06: /* recalculation in progress */
 926                                 case 0x07: /* operation in progress */
 927                                 case 0x08: /* Long write in progress */
 928                                 case 0x09: /* self test in progress */
 929                                 case 0x14: /* space allocation in progress */
 930                                         action = ACTION_DELAYED_RETRY;
 931                                         break;
 932                                 default:
 933                                         action = ACTION_FAIL;
 934                                         break;
 935                                 }
 936                         } else
 937                                 action = ACTION_FAIL;
 938                         break;

49476 Mar 29 10:08:04 lnx1aocpdck01d kernel: sd 12:0:0:0: [sdh] Add. Sense: Logical unit communication failure
49477 Mar 29 10:08:04 lnx1aocpdck01d kernel: sd 12:0:0:0: [sdh] CDB: Write(10) 2a 00 00 10 45 7b 00 00 80 00
49478 Mar 29 10:08:04 lnx1aocpdck01d kernel: blk_update_request: I/O error, dev sdh, sector 1066363

But from the linux LIO code in some other platform it will retry a infinite number of times.

And the EXSI is reporting some wired sense errors with tcmu-runner much lock switching, from the following doc:
http://www.yellow-bricks.com/2008/12/09/fixed-or-mru-what-will-it-be/

We can know that with the FIXED multipath policy may will retry when receiving of this.

lxbsz: 2937         [TCM_LOGICAL_UNIT_COMMUNICATION_FAILURE] = {
    2938                 /*
    2939                  * Returning ILLEGAL REQUEST would cause immediate IO errors on
    2940                  * Solaris initiators.  Returning NOT READY instead means the
    2941                  * operations will be retried a finite number of times and we                                                                      
    2942                  * can survive intermittent errors.
    2943                  */
    2944                 .key = NOT_READY,
    2945                 .asc = 0x08, /* LOGICAL UNIT COMMUNICATION FAILURE */
    2946         },
    2947 };

mikechristie:
> 49478 Mar 29 10:08:04 lnx1aocpdck01d kernel: blk_update_request: I/O error, dev sdh, sector 1066363

Just a FYI. The code you quoted was for the SCSI layer and if it decides to retry or fail upwards. This error message above just indicates it was only failed at the SCSI level upwards. If dm-mutltipath also failed it then you would see another error message like that but from dm-N.

Check out __scsi_error_from_host_byte to see what error value gets returned to dm, then you can check if dm handles that a a hard failure or retryable.

lxbsz:
@mikechristie @pkalever 

Updated, Please review.

This will also make the logs output more clear and readable for non supported cmds.

mikechristie:
Convert all the code to use the same error codes. No mixing SCSI sense/status, -Eyxz and now TCMU_CMD_*.

Just one error code type to rule them all :)

mikechristie:
> Yeah, will fix it.

Actually, do not worry about it. I will fix it up. This has lots of conflicts with my failover patchset, so I will just handle my own review comment. Will post patch here.

lxbsz:
Sure Mike, Yesterday I was busying for some urgent cases, so didn't have much time to update this.

mikechristie:
> Sure Mike, Yesterday I was busying for some urgent cases, so didn't have much time to update this.

No problem.

I made a patchset here

https://github.com/open-iscsi/tcmu-runner/pull/413

Be warned that the push is all mixed up and contains new patches and also patches from

https://github.com/open-iscsi/tcmu-runner/pull/407

I'm not sure how to make git/github recognize that some patches are in another PR.

api.c
```diff
@@ -1200,7 +1200,9 @@ int tcmu_emulate_start_stop(struct tcmu_device *dev, uint8_t *cdb,
 #define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 1)
```
mikechristie: Rebuild this patch against my set.

lxbsz:
@mikechristie 
Updated this PR. Just keep the second patch.

PR #405, opened 2018-04-27
fix the memory cross boundary bug in tcmu_rbd_detect_device_class

letterwuyu:
the length of the rados_mon_command return mon_buf in memory is mon_buf_len,
So like "mon_buf[mon_buf_len] ='\0'"will memory transboundary.
Signed-off-by: letterwuyu letterwuyu@gmail.com

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -271,7 +271,7 @@ static void tcmu_rbd_detect_device_class(struct tcmu_device *dev)
 	rados_buffer_free(mon_status_buf);
 
 	/* expected JSON output: "{..."crush_rule":"<rule name>"...}" */
-	mon_buf[mon_buf_len] = '\0';
+	mon_buf[mon_buf_len - 1] = '\0';
 	crush_rule = strstr(mon_buf, "\"crush_rule\":\"");
 	if (!crush_rule) {
 		tcmu_dev_warn(dev, "Could not locate crush rule key\n");
@@ -318,7 +318,7 @@ static void tcmu_rbd_detect_device_class(struct tcmu_device *dev)
 	rados_buffer_free(mon_status_buf);
 
 	/* expected JSON output: ["<rule name>",["<rule name>"...]] */
-	mon_buf[mon_buf_len] = '\0';
+	mon_buf[mon_buf_len - 1] = '\0';
 	if (strstr(mon_buf, crush_rule)) {
 		tcmu_dev_dbg(dev, "Pool %s associated to SSD device class.\n",
 			     state->pool_name);


rbd.c
```diff
@@ -260,22 +260,29 @@ static void tcmu_rbd_detect_device_class(struct tcmu_device *dev)
...
 		tcmu_dev_warn(dev, "Could not retrieve pool crush rule (Err %d)\n",
 			      ret);
 		return;
 	}
 	rados_buffer_free(mon_status_buf);
 
 	/* expected JSON output: "{..."crush_rule":"<rule name>"...}" */
-	mon_buf[mon_buf_len] = '\0';
+	mon_buf_len = temp_mon_buf_len + 1;
+	if(!(mon_buf = (char *)malloc(mon_buf_len))) {
```
mikechristie: Hey letterwuyu,

Very nice find!

Just use the standard coding style in the rest of the  file. No need to cast from void to char and just do the alloc and test on 2 lines like:

mon_buf = malloc(mon_buf_len);
if (!mon_buf) {

Also, in this memory allocate failure path, do we also need a:

rados_buffer_free(temp_mon_buf);

?

rbd.c
```diff
@@ -304,27 +311,34 @@ static void tcmu_rbd_detect_device_class(struct tcmu_device *dev)
...
 		tcmu_dev_warn(dev, "Could not retrieve pool crush rule ls-by-class (Err %d)\n",
 			      ret);
 		goto free_crush_rule;
 	}
 	rados_buffer_free(mon_status_buf);
 
 	/* expected JSON output: ["<rule name>",["<rule name>"...]] */
-	mon_buf[mon_buf_len] = '\0';
+	mon_buf_len = temp_mon_buf_len + 1; 
+	if(!(mon_buf = (char *)malloc(mon_buf_len))) {
```
mikechristie: Same comment as above.

rbd.c
```diff
@@ -260,22 +260,31 @@ static void tcmu_rbd_detect_device_class(struct tcmu_device *dev)
...
-	if (ret < 0 || !mon_buf) {
+	if (ret < 0 || !temp_mon_buf) {
 		tcmu_dev_warn(dev, "Could not retrieve pool crush rule (Err %d)\n",
 			      ret);
 		return;
 	}
 	rados_buffer_free(mon_status_buf);
 
 	/* expected JSON output: "{..."crush_rule":"<rule name>"...}" */
-	mon_buf[mon_buf_len] = '\0';
```
dillaman: Nit: why not just set ```mon_buf[mon_buf_len - 1] = '\0';``` and test for ```mon_buf_len == 0``` on line 266 instead of adding all this extra code? It's in JSON format so there is plenty of extra characters that won't be needed.

letterwuyu: think you for your nit.yes, you are right,and I don't think it will affect the performance. I don't know if it fits our coding habits to overlay a character. If it fits, your proposal is very good, and I'll update the code according to your advice.

dillaman: I'll leave it for Mike to decide, but these buffers are now owned by the caller and are already sliced up during processing. Thanks for the catch.

mikechristie: @letterwuyu, do what dillaman suggested.

When I made my review comment I had thought there were going to be valid non null characters up to the end of the buffer and we needed to add a null after what was returned.

letterwuyu: ok, I have update this pr.

letterwuyu:
ok, i have update this pr.

dillaman:
@letterwuyu Why did you close the PR?

PR #406, opened 2018-04-28
libtcmu: drop duplicated ";"

baihuahua:
These ";" are not needed because we will attach one when everytime we call these macros.

Signed-off-by: Yaowei Bai <baiyaowei@cmss.chinamobile.com>


mikechristie:
Thanks.

PR #407, opened 2018-04-30
Explicit fo

mikechristie:
Add back explicit failover support. It requires the ceph-iscsi-config changes here:

https://github.com/ceph/ceph-iscsi-config/pull/54


mikechristie:
One note. I left implicit failover support, but I am planning on removing it when I figure out if we need to make the explicit failover operation execute in a different thread than the cmdproc one, because that would use the same locking/threading as implicit support.

dillaman:
@mikechristie We also need to consider the upgrade path from implicit configurations -> explicit. How would an initiator detect that it should use explicit ALUA on an active session? Would we have to upgrade and restart tcmu-runner's across all gateway nodes prior to upgrading ceph-iscsi-config?

tcmur_device.c
```diff
@@ -238,33 +238,99 @@ int tcmu_cancel_lock_thread(struct tcmu_device *dev)
...
+	ret = rhandler->get_lock_tag(dev, tag);
+	tcmu_dev_dbg(dev, "Got rc %d tag %hu\n", ret, *tag);
+retry:
+	switch (ret) {
+	case 0:
+		break;
+	case -ENOENT:
+		/* No lock holder yet */
+		break;
+	case -ESHUTDOWN:
```
dillaman: Nit: seems to be leaking RBD-specific error codes back to the generic handler. Could you just re-use the existing lock state enums (a la ```TCMUR_DEV_LOCK_FENCED```)?

rbd.c
```diff
@@ -74,6 +74,10 @@
 #endif
 #endif
 
+#define TCMU_RBD_LOCKER_TAG_KEY "tcmu_rbd_locker_tag"
+#define TCMU_RBD_LOCKER_TAG_FMT "tcmu_tag=%hu,rbd_client=%s"
+#define TCMU_RBD_LCOKER_BUF_LEN 256
```
dillaman: Nit: typo

rbd.c
```diff
@@ -612,7 +616,177 @@ static int tcmu_rbd_lock_break(struct tcmu_device *dev, char **orig_owner)
...
+	if (!num_owners) {
+		/* there would be stale metadata due to a crash */
+		ret = -ENOENT;
+		goto done;
+	}
+
+	metadata_owner = strstr(buf, "rbd_client=");
+	if (!metadata_owner) {
+		tcmu_dev_err(dev, "Invalid lock tag %s.\n", buf);
+		/* Force initaitor to retry STPG */
```
dillaman: Nit: typo

rbd.c
```diff
@@ -612,7 +616,177 @@ static int tcmu_rbd_lock_break(struct tcmu_device *dev, char **orig_owner)
...
+		ret = -ENOENT;
+		goto free_owners;
+	}
+
+	metadata_owner += 11;
+	if (strcmp(metadata_owner, owners[0])) {
+		tcmu_dev_dbg(dev, "owner mismatch %s %s\n", metadata_owner,
+			     owners[0]);
+		/*
+		 * A node could be in the middle of grabbing the lock or it
```
dillaman: Any concern about this potential race? I had originally thought of having each target register a metadata key of something like ```tcmu_target_group_<group id>=<client #>``` just after opening the image. Then, once the lock is acquired, the owner is already in the format of ```client.<client #>``` so you would just need to query the 2 - 4 metadata tags for each group to find the matching client id. Your solution is nice since it only involves a single metadata key lookup.

rbd.c
```diff
@@ -612,7 +616,177 @@ static int tcmu_rbd_lock_break(struct tcmu_device *dev, char **orig_owner)
...
+
+static int tcmu_rbd_unlock(struct tcmu_device *dev)
+{
+	struct tcmu_rbd_state *state = tcmu_get_dev_private(dev);
+	int ret;
+
+	if (tcmu_rbd_has_lock(dev) != 1)
+		return TCMUR_DEV_LOCK_UNLOCKED;
+
+	ret = rbd_metadata_remove(state->image, TCMU_RBD_LOCKER_TAG_KEY);
```
dillaman: Is it possible this could race w/ another client stealing the lock? Perhaps just leave it set to the old value and let the mismatch code above detect that it's out-of-date?

rbd.c
dillaman: ... apparently never mind. I thought that the librbd API provided the client name but it looks like it only provides the IP address of the owner.  I guess that will be something to fix when I add shared lock support (along w/ the ability to store user-provided tags on the lock).

tcmur_device.c
mikechristie: I am going to use the TCMUR_DEV_LOCK codes like suggested above.

Note off list I said I was going to use the TCMU error codes being added in 403,  but that is not done yet. Will convert to them later.

rbd.c
mikechristie: Ok sounds good. I was thinking if another client took the lock we would be blacklisted by the time we got here. But yeah your suggestion is nicer because we have that other mismatch code and then it saves the worry here. Will drop.

tcmur_device.c
mikechristie: I flip flopped again on this. In this PR I left the -Exyz codes. In this patchset

https://github.com/open-iscsi/tcmu-runner/pull/413

I went and made all the core and callout IO path code including above use common TCMU_ status codes.

mikechristie:
> @mikechristie We also need to consider the upgrade path from implicit configurations 

I am still working on upgrade instructions. In some cases it is going to be messy because the switch in supported failover types is not automatically detected. It will take some manual commands.

PR #411, opened 2018-05-02
tcmu: Add Media Change

bgly:
This patch series will add media change that is
triggered via targetcli without involving the kernel.

This patch will also need a targetcli change.


bgly:
With the targetcli change on this patch you can run:

sudo targetcli /backstores/user:fbo/ mediachange test size=3012747264 cfgstring=/home/neo/RHEL-ALT-7.4-20170726.0-Server-ppc64le-dvd1_4.11.0-21.iso

main.c
```diff
@@ -235,8 +235,41 @@ on_check_config(TCMUService1 *interface,
...
+	if (ret == 0)
+		reason = "success";
+
+	tcmu_unblock_device(dev);
+	tcmur_set_pending_ua(dev, TCMUR_UA_DEV_MEDIUM_CHANGED);
+	g_dbus_method_invocation_return_value(invocation,
+		g_variant_new("(is)", ret, reason ? : "unknown"));
+
+	if (ret != 0)
+		free(reason);
```
mikechristie: Does dbus allocate memory for "reason", or was this leftover from another version of the code.

main.c
```diff
@@ -235,8 +235,41 @@ on_check_config(TCMUService1 *interface,
...
+	char *reason = NULL;
+	int ret = 0;
+
+	tcmu_set_dev_cfgstring(dev, cfgstring);
+	tcmu_update_num_lbas(dev, size);
+	tcmu_set_dev_size(dev);
+
+	tcmu_block_device(dev);
+	tcmu_flush_device(dev);
+	ret = tcmu_reopen_dev(dev);
```
mikechristie: In my failover patchset I added an argument to this to control retries. I hope to merge your patch then mine over it. For that, how many retries should I set this to? Should it be infinite like it is now, or should there be a limit? What do you want the limit to be?

tcmur_cmd_handler.c
```diff
@@ -2267,6 +2315,12 @@ int tcmur_cmd_passthrough_handler(struct tcmu_device *dev,
 	if (!rhandler->handle_cmd)
 		return TCMU_NOT_HANDLED;
 
+	if (rdev->pending_uas)
+		tcmu_dev_err(dev, "tcmur_cmd_passthrough_handler handle pending ua:%i\n", (ffs(rdev->pending_uas) - 1));
```
mikechristie: Was this debugging code?

main.c
```diff
@@ -338,7 +411,7 @@ on_handler_appeared(GDBusConnection *connection,
 	if (info->register_invocation) {
 		info->connection = connection;
 		tcmur_register_dbus_handler(handler);
-		dbus_export_handler(handler, G_CALLBACK(on_dbus_check_config));
+		dbus_export_handler(handler, G_CALLBACK(on_dbus_check_config), G_CALLBACK(on_dbus_mediachange));
```
mikechristie: Could you break this up to 80 cols here and below.

main.c
```diff
@@ -788,6 +861,7 @@ static int dev_added(struct tcmu_device *dev)
 	if (ret < 0)
 		goto cleanup_io_work_queue;
 
+	rhandler->dev = dev;
```
mikechristie: I don't think you get a different handler instance per device so this overwrites a previous device.

If it's not possible to do a callback per device, then I think for on_mediachange you need to pass in the device name string in the format used in the netlink functions so we can just look up the device and use it. We could put it at the front of the cfgstring string.

I'm guessing it's possible but it would get ugly for with 1000s of devices it would be too many resources, so I guess just passing in the device in the cfg string is ok.

main.c
bgly: I will remove the free, I took it from check_config. They used it because they used strdup which handles the malloc in the background, and I dont use that, so I can remove.

tcmur_cmd_handler.c
bgly: Yes, I will remove.

main.c
bgly: Yes, will change

main.c
bgly: So before, I fix this part, do you want a specific way of exposing tcmu_device to the mediachange function? As of now tcmur_handler has no way of getting to a tcmu_device which would then allow a search for the device via the name passed from targetcli. - One way could possibly be to add a darray into tcmur_handler that would contain a list of the tcmu_device pointers and then search via that...

main.c
bgly: There is nothing in the spec that details the amount of retries so based upon how this is written I think a few retries would be fine then if it fails to just revert the info back?

mikechristie:
Wait a day. I plan on putting it in the 1.4.0 final. I just want Andy or someone to check out the rtslib dbus stuff, because I am a rookie on that :)

bgly:
Sounds good! Thanks

bgly:
Any news on when this can go in?

mikechristie:
Sorry for the delay. I think it looks ok overall. I have not had a chance to review it for bugs. I should get to it this week.

mikechristie:
@bgly,

I updated your patches here

https://github.com/open-iscsi/tcmu-runner/pull/435

I think we will want to update tcmu_set_configfs_dev_params's dev_config handler with a call to tcmu_update_uio_info similar to how you did for the dev_config attribute.

PR #413, opened 2018-05-05
Add TCMU status codes

mikechristie:
This has the tcmu IO path and related callouts use new TCMU_STS status codes instead of mixing SCSI status/sense, -Exyz, TCMU lock codes and the TCMU handler codes.

I made the code still allow handlers to set sense in case out side code wants to do this. I also left file_zbc as this type of handler because it is more the old school sync handle_cmd handler that does most of its own SCSI processing. So I think it made sense to allow it to continue to do lower level sense handling. Other handlers should use the TCMU_STS codes so we do not have lots of different code paths like before.


mikechristie:
Ughhhh, ok I made this patchset over my failover patchset

https://github.com/open-iscsi/tcmu-runner/pull/407

so only half the patches are for the status changes. When 407 is merged, I will repush this to make it easier to review.

PR #415, opened 2018-05-08
rbd: new option for specifying the id

tjakob:
Allows differnt ids to connect to the RADOS cluster


bgly:
Looks good to me

rbd.c
```diff
@@ -423,7 +424,7 @@ static int tcmu_rbd_image_open(struct tcmu_device *dev)
 	struct tcmu_rbd_state *state = tcmu_get_dev_private(dev);
 	int ret;
 
-	ret = rados_create(&state->cluster, NULL);
+	ret = rados_create(&state->cluster, state->id);
```
mikechristie: Thanks for the patch.

Did you want to use the rados_create2 full name support instead of the rados_create id where "client" is assumed to be the prefix?

tjakob: For our current use case, assuming "client" is okay with us, and we don't currently have any plans for setting the cluster name.

PR #393, opened 2018-04-03
alua: reduce the debug logs

lxbsz:
When the loglevel is in debug mode, for each IO/non-generic SCSI cmd
there will be one debug message output, and there will be aweful
amount of logs full of the log file.

Signed-off-by: Xiubo Li <xiubli@redhat.com>

PR #416, opened 2018-05-10
glfs: add a dummy lock handler

pkalever:
pretend failover is supported for implicit alua access type.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -713,6 +713,11 @@ out:
 	return TCMU_STS_NO_RESOURCE;
 }
 
+static int tcmu_glfs_lock(struct tcmu_device *dev)
+{
+	return 0;
+}
+
 /*
  * For backstore creation
  *
@@ -744,6 +749,7 @@ struct tcmur_handler glfs_handler = {
 	.write          = tcmu_glfs_write,
 	.reconfig       = tcmu_glfs_reconfig,
 	.flush          = tcmu_glfs_flush,
+	.lock           = tcmu_glfs_lock,
 };
 
 /* Entry point must be named "handler_init". */


lxbsz:
@mikechristie 

If we have 1000 targets and it's very possibly that all the active paths of the 1000 multipath devices will be located in the same tcmu-runner node, and this is not what we want.

In future we will add the failover feature support for the glusterfs, but the related develop in glusterfs hasn't be finished and not sure the date. So currently we'd like to use the ALUA feature in tcmu-runner to support the load balance of the IOs, so here we need a dummy lock for glfs handler.

Any idea or advice about this ?

mikechristie:
Just fix the code to support what you guys need for now:

- @lxbsz resend the patch to remove the dbg log message for each IO.

- Drop the failover_is_supported check in tcmu_get_alua_grp and in alua_check_state. For the other uses, rename the function to lock_is_required() so it makes more sense for when we now use it.

- In alua_implicit_transition you could just do
```
if !lock_is_required()
   return TCMU_STS_OK;
```

at the top of the function.

lxbsz:
Yeah, thanks.

> @lxbsz resend the patch to remove the dbg log message for each IO.

Sorry Mike, didn't get you about this.

pkalever:
@mikechristie  didn't got you, @lxbsz and myself are confused thinking :-)

https://github.com/open-iscsi/tcmu-runner/blob/master/api.c#L1196
`tcmu_dev_dbg_scsi_cmd(dev, buf);
`
or
https://github.com/open-iscsi/tcmu-runner/blob/master/alua.c#L568
`tcmu_dev_dbg(dev, "lock state %d\n", rdev->lock_state);`

mikechristie:
My fault. Forgot you had that other dbg patch. @lxbsz I meant fix up and resend this patch

https://github.com/open-iscsi/tcmu-runner/pull/393

I was going to remove all the implicit code when I figured out if I needed the ability to take the lock from a thread like we do for implicit (for explicit right now we just take it from the cmd's context which is the cmdproc thread), so PR 393 would not be needed since the code would be removed.

If you guys are going to use implicit for a little bit then I thought you would want that patch. If not, you can ignore the comment about it.

lxbsz:
> My fault. Forgot you had that other dbg patch. @lxbsz I meant fix up and resend this patch
> 
> #393

https://github.com/open-iscsi/tcmu-runner/pull/417

Done by creating a new PR.

lxbsz:
> If you guys are going to use implicit for a little bit then I thought you would want that patch. If not, you can ignore the comment about it.

Yeah, we need implicit currently.

> Just fix the code to support what you guys need for now:
> 
> @lxbsz resend the patch to remove the dbg log message for each IO.
> 
> Drop the failover_is_supported check in tcmu_get_alua_grp and in alua_check_state. For the other uses, rename the function to lock_is_required() so it makes more sense for when we now use it.
> 
> In alua_implicit_transition you could just do
> 
> if !lock_is_required()
>    return TCMU_STS_OK;
> at the top of the function.

@pkalever @mikechristie 

Updated this in https://github.com/open-iscsi/tcmu-runner/pull/424.

Thanks,

mikechristie:
Merged 424 so as far as I can tell we do not need this PR anymore, so closing.

Thanks.

PR #417, opened 2018-05-14
alua: reduce the debug logs

lxbsz:
When the loglevel is in debug mode, for each IO/non-generic SCSI cmd
there will be one debug message output, and there will be aweful
amount of logs full of the log file.


lxbsz:
@mikechristie 

I couldn't reopen that old PR(https://github.com/open-iscsi/tcmu-runner/pull/393), so just create a new one.

PR #418, opened 2018-05-16
runner: don't allow multiple instances of daemon to run

pkalever:
until now there is no check to defend on multiple runs of daemon within
the same node. This patch takes a non blocking lock on 'tcmu.lock' file,
if it succeeds only then daemon is allowed to run, if there is already a
lock on lock-file (taken by some other instance of tcmu-runner) we will exit.

### DIFF
--- a/main.c
+++ b/main.c
@@ -56,6 +56,8 @@
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
 
+# define TCMU_LOCK_FILE   "/var/run/lock/tcmu.lock"
+
 static char *handler_path = DEFAULT_HANDLER_PATH;
 
 static struct tcmu_config *tcmu_cfg;
@@ -994,6 +996,8 @@ int main(int argc, char **argv)
 	GIOChannel *libtcmu_gio;
 	guint reg_id;
 	bool new_path = false;
+	struct flock lock_fd = {0, };
+	int fd;
 	int ret;
 
 	while (1) {
@@ -1057,18 +1061,34 @@ int main(int argc, char **argv)
 	if (tcmu_setup_log())
 		goto destroy_config;
 
-	tcmu_dbg("handler path: %s\n", handler_path);
+	fd = creat(TCMU_LOCK_FILE, S_IRUSR | S_IWUSR);
+	if (fd == -1) {
+		tcmu_err("creat(%s) failed: [%m]\n", TCMU_LOCK_FILE);
+		goto destroy_log;
+	}
+
+	lock_fd.l_type = F_WRLCK;
+	if (fcntl(fd, F_SETLK, &lock_fd) == -1) {
+		if (errno == EAGAIN) {
+			tcmu_err("tcmu-runner is already running...\n");
+		} else {
+			tcmu_err("fcntl(F_SETLK) on lockfile %s failed: [%m]\n",
+			         TCMU_LOCK_FILE);
+		}
+		goto close_fd;
+	}
 
 	ret = load_our_module();
 	if (ret < 0) {
 		tcmu_err("couldn't load module\n");
-		goto destroy_log;
+		goto close_fd;
 	}
 
+	tcmu_dbg("handler path: %s\n", handler_path);
 	ret = open_handlers();
 	if (ret < 0) {
 		tcmu_err("couldn't open handlers\n");
-		goto destroy_log;
+		goto close_fd;
 	}
 	tcmu_dbg("%d runner handlers found\n", ret);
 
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -1057,14 +1061,27 @@ int main(int argc, char **argv)
...
+	fd = creat(TCMU_LOCK_FILE, S_IRUSR | S_IWUSR);
+	if (fd == -1) {
+		tcmu_err("creat(%s) failed: [%m]\n", TCMU_LOCK_FILE);
+		goto destroy_log;
+	}
+
+	lock.l_type = F_WRLCK;
+	if (fcntl(fd, F_SETLK, &lock) == -1) {
+		tcmu_err("tcmu-runner is already running...\n");
+		close(fd);
```
lxbsz: We'd better add one goto for closing the fd to make the code more readable.

Or we should it close(fd) everywhere just after this line when trying to goto error tags.

main.c
```diff
@@ -1057,14 +1061,27 @@ int main(int argc, char **argv)
...
+		goto destroy_log;
+	}
 
 	ret = load_our_module();
 	if (ret < 0) {
 		tcmu_err("couldn't load module\n");
 		goto destroy_log;
```
lxbsz: Just like here.

main.c
```diff
@@ -1057,14 +1061,27 @@ int main(int argc, char **argv)
...
-	tcmu_dbg("handler path: %s\n", handler_path);
+	fd = creat(TCMU_LOCK_FILE, S_IRUSR | S_IWUSR);
+	if (fd == -1) {
+		tcmu_err("creat(%s) failed: [%m]\n", TCMU_LOCK_FILE);
+		goto destroy_log;
+	}
+
+	lock.l_type = F_WRLCK;
+	if (fcntl(fd, F_SETLK, &lock) == -1) {
+		tcmu_err("tcmu-runner is already running...\n");
```
lxbsz: Should we check that only when errno == EACCES or EAGAIN we will be sure that the tcmu-runner main process is already runing ? Or it should be other op errors.

main.c
pkalever: Good catch @lxbsz

main.c
pkalever: Sure, I will defend on EAGAIN

pkalever:
@lxbsz updated with the suggestions. Thanks!

lxbsz:
It looks good to me and test it and the logs are:

 /usr/bin/tcmu-runner --tcmu-log-dir=/var/log/gluster-block/ --debug
The logdir option from the tcmu.conf will be ignored
Inotify is watching "/etc/tcmu/tcmu.conf", wd: 1, mask: IN_ALL_EVENTS
2018-05-16 09:30:46.462 6677 [ERROR] main:1073: tcmu-runner is already running...

main.c
```diff
@@ -56,6 +56,8 @@
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
 
+# define TCMU_LOCK_FILE   "/var/run/tcmu.lock"
```
mikechristie: Is this more normally in the lock dir like

/var/run/lock/tcmu.lock
or
/var/run/lock/tcmu/lock
?

main.c
```diff
@@ -994,6 +996,8 @@ int main(int argc, char **argv)
 	GIOChannel *libtcmu_gio;
 	guint reg_id;
 	bool new_path = false;
+	struct flock lock = {0, };
+	int fd;
```
mikechristie: How about rename to lock_fd.

main.c
```diff
@@ -1146,6 +1173,8 @@ int main(int argc, char **argv)
 	tcmulib_close(tcmulib_context);
 err_free_handlers:
 	darray_free(handlers);
+close_fd:
+	close(fd);
```
mikechristie: Why for the normal shutdown we do a F_UNLCK and a close, but for the error path we only do a close on the fd?

From the man page it sounds like the same thing happens and the lock is released due to the process exiting. Is that correct? Is it just customary to only do a close in error paths liek this?

main.c
pkalever: Yes will pick the first one.

main.c
pkalever: I will change this.

main.c
pkalever: From man:

* As well as being removed by an explicit F_UNLCK, record locks are automatically released when the process terminates.

* If a process closes any file descriptor referring to a file, then all of the process's locks on that file are released


I just wrote more like sequence, like open, grab lock, release lock, close in the normal shutdown.
And forget to add release lock at error path.
 
But, just closing the fd without F_UNLCK will also release the lock anyway.

@mikechristie I'm not sure what would you prefer:
1. remove F_UNLCK completely and just close(fd) in the both normal & error paths or
2. add F_UNLCK at error path, which I forget before ?

Thanks!

mikechristie: Lol, I do not know.

Go for the F_UNLOCK just to make it clear I guess. We seem to do that for some other things so it will at least be consistent.

pkalever: OK - Make-sense :-)

pkalever:
Updated with the suggested changes. Thanks!

PR #425, opened 2018-05-30
ANNOUNCE/RFC: Relicense libtcmu/tcmu-runner

mikechristie:
Hi,

As every person with a commit, except 1 person that has not yet got back to me, know another group of developers wants to use tcmu in their GPL 2 based project qemu-tcmu:

https://lists.gnu.org/archive/html/qemu-devel/2016-10/msg04408.html

The problem is that we are Apache 2 based and they do not mix well when you want to bring Apache 2 code into a GPL 2 project. So I have contact everyone with code still committed in the upstream tree (if you submitted code but we later removed/replaced it I did not bug you)  and asked if relicensing to a dual Apache 2 or LGPL v2.1 license is ok and everyone has agreed so far. This means if you have or are going to make a project that requires Apache 2, no changes are needed. Go about like you have been.

We are just trying to track down 1 committer, so I am posting this now in case I did miss someone or people have questions/comments.


mikechristie:
The last commiter just got back from vacation and gave their OK, so merging.

The next step on the TODO is to add back the stable libtcmu API. We... I guess I did that :) ...  broke that a while back thinking there were no users and to allow us some flexibility. Now that we have runner doing its more advanced commands like the vmware VAAI ones and have a better understanding of what other users want to do with the lib, we will fix up the lib into a proper stable one.

PR #437, opened 2018-06-27
glfs: allow the backend file size not aligns to the block_size

lxbsz:
The glfs allows the backend file size not to align
to the block_size. But the dev_size here in tcmu-runner
will round down and align it to the block_size.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -481,6 +481,7 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
 	struct stat st;
 	int ret = -EIO;
 	long long dev_size;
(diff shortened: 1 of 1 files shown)


pranithk:
@lxbsz How will a block-device that is not aligned with the block-size be useful? Will it be useful in any case? Do you think it is better to not allow a file size creation which is not aligned with block-size?

lxbsz:
@pranithk 
For the block device it should always be aligned to the block-size, currently I haven't encounter the unaligned block device has special use case.

Will try to defend this in the gluster-block at the same time.
Thanks

PR #439, opened 2018-06-29
Wip fix 3

umuzhaohui:
tcmu-runner: fix bug in HA about re-acquire lock
    
    in ha, writing data to multipath device, Wait a minute, after
    close the ip on node1, the write command can not be processed
    on node0 by tcmu-runner rightly. Because the node1's tcmu-runner
    do not drop the lock on the image, node0's tcmu-runner should
    get the lock on the image after rbd return -EROFS error.
    
    Signed-off-by: Zhaohui Yang <280965257@qq.com>


LICENSE.Apache2
```diff
@@ -1,201 +1,339 @@
-Apache License
-                           Version 2.0, January 2004
```
mikechristie: I think you need to reclone your tree, because we only want that bit at the very bottom and not this license stuff.

umuzhaohui: OK，you can just  manually edit the patch and update the EROFS part at the bottom. i think there is some errors in my git

mikechristie: Hey, I fixed up the patch and tested it here. 

One question on the signed off though. The "From" line on the patch has:

From: xinxin shu <shuxinxin@chinac.com>

but the signed off is:

Signed-off-by: Zhaohui Yang <280965257@qq.com>

Are these the same people but just using 2 email addresses and names?

umuzhaohui: the same people ： Zhaohui Yang 280965257@qq.com

mikechristie: Thanks merged in:

https://github.com/open-iscsi/tcmu-runner/commit/fc8b065ab79d32d0494384b228abf9a4fe392206

PR #444, opened 2018-07-14
Reinstall libtcmu API and header files

baihuahua:
The separation work of libtcmu and tcmu-runner seems harder than we thought and is going to take a bit more time to achieve. Let's reinstall libtcmu header files and API first for now as the future sepatation work can still base on this to move on and qemu-tcmu can also make a little forward. @mikechristie @amarts @lxbsz

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ target_link_libraries(tcmu
   ${LIBNL_GENL_LIB}
   ${GLIB_LIBRARIES}
   )
-install(TARGETS tcmu LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} NAMELINK_SKIP)
+install(TARGETS tcmu LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
 # Stuff for building the static library
 add_library(tcmu_static
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,7 +272,7 @@ configure_file (
   )
 install(SCRIPT tcmu.conf_install.cmake)
 
-install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
+install(FILES libtcmu.h libtcmu_common.h libtcmu_priv.h darray.h
   DESTINATION include)
 install(FILES org.kernel.TCMUService1.service
   DESTINATION /usr/share/dbus-1/system-services)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,8 @@ configure_file (
   )
 install(SCRIPT tcmu.conf_install.cmake)
 
+install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
+  DESTINATION include)
 install(FILES org.kernel.TCMUService1.service
   DESTINATION /usr/share/dbus-1/system-services)
 install(FILES tcmu-runner.conf DESTINATION /etc/dbus-1/system.d)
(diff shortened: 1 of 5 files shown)


mikechristie:
Sorry. I misread it the first time. I thought it depended on your discussion with Xiubo on the PR. Will get to it today.

lxbsz:
@mikechristie If we change the whole code one time in a PR it will be very hard to review, it should be okay for me to split this in different small PRs to make it easier.

And IMO, the goal is to make the libtcmu.so include and provide the following features and APIs:
1, the basic helpers of the kernel uio device and target/module configfs.
2, some basic scsi handling helpers, such as the INQUIRY...
3, dbus related code.

baihuahua:
Yes, small PRs should be more easier to review. We could start from this PR and implement other features one by one. This should make sense. @mikechristie

CMakeLists.txt
```diff
@@ -272,7 +272,7 @@ configure_file (
   )
 install(SCRIPT tcmu.conf_install.cmake)
 
-install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
+install(FILES libtcmu.h libtcmu_common.h libtcmu_priv.h darray.h
```
mikechristie: For the final lib we are not going to have users use/include the private definitions. The lib user should also not need the darray.h header unless we expose an api that uses a darrary.

libtcmu_priv.h
```diff
@@ -20,10 +20,7 @@
 #include <gio/gio.h>
 #include <pthread.h>
 
-#include "scsi_defs.h"
 #include "darray.h"
-#include "ccan/list/list.h"
-#include "tcmur_aio.h"
```
mikechristie: This patch is fine.

alua.c
```diff
@@ -23,6 +23,7 @@
 #include "libtcmu_log.h"
 #include "libtcmu_common.h"
```
mikechristie: This patch seems ok.

CMakeLists.txt
```diff
@@ -272,6 +272,8 @@ configure_file (
   )
 install(SCRIPT tcmu.conf_install.cmake)
 
+install(FILES libtcmu.h libtcmu_common.h tcmu-runner.h
```
mikechristie: Won't the final new lib have a new header name so it does not conflict with users of the old libtcmu?

Also wether the final lib needs something like tcmu-runner.h is still up in the air.

CMakeLists.txt
mikechristie: The lib users should not need to ever directly interact with tcmu_device. It is internal to the lib. The only interact with it via libtcmu helpers.

mikechristie: Check out how consumer.c or tcmu-synthesizer worked with libtcmu originally for an example. They never directly interacted with tcmu_device so we could make internal changes and not have to update every libtcmu user.

mikechristie: Oh yeah for libtcmu_context and darray that should not be an issue. The users of libtcmu only pass around a tcmulib_context struct pointer. They never access it, so the forward declaration in the libtcmu.h should be enough.

libtcmu_priv.h
baihuahua: Thanks.

alua.c
baihuahua: Thanks.

CMakeLists.txt
baihuahua: I see. I'll tune this as your suggestion. Thansk.

PR #448, opened 2018-07-27
To enhance the reuseability of code, move function 'tcmu_cancel_thread' from 'main.c' to 'api.c' .

Gangbiao:
Move the definition of function 'void tcmu_cancel_thread(pthread_t thread)' from 'main.c' to 'api.c', 
so this function can be called by 'main.c', 'libtcmu_config.c' or other module. 
There is no need to implement the same code two or more times.


mikechristie:
Nice. Thanks.

PR #409, opened 2018-05-02
extra: fix RPM build error

lxbsz:
```
RPM build errors:
    Directory not found:
/root/tcmu-runner-orig/extra/rpmbuild/BUILDROOT/tcmu-runner-1.3.0.54.g6684a21-0.el7.centos.x86_64/etc/tcmu
    File not found:
/root/tcmu-runner-orig/extra/rpmbuild/BUILDROOT/tcmu-runner-1.3.0.54.g6684a21-0.el7.centos.x86_64/etc/tcmu/tcmu.conf
-----
```

Signed-off-by: Xiubo Li <xiubli@redhat.com>

PR #450, opened 2018-08-07
spec: fix rpm package build error

lxbsz:
cp: cannot stat 'LICENSE': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.McXq6H (%doc)

### DIFF
--- a/tcmu-runner.spec
+++ b/tcmu-runner.spec
@@ -65,7 +65,7 @@ userspace libraries they like.
 %dir %{_usr}/lib64/tcmu-runner/
 %{_usr}/lib64/tcmu-runner/*
 %{_mandir}/man8/*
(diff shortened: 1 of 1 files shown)


mikechristie:
Thanks.

One related question. What happened with this other spec patch

https://github.com/open-iscsi/tcmu-runner/pull/409

Were you able to replicate the problem? Is the patch correct? Do you want me to merge it or just wait on the user to report if it worked for them?

lxbsz:
Yeah, that patch works.

So let's wait for a week to get the user's ack.

Thanks,

lxbsz:
@mikechristie 
For #409 I have found the root cause and fixed it in spec file.
Please review.
Thanks.

PR #451, opened 2018-08-13
rbd: detect pools associated to NVMe device class

dillaman:
A Ceph bluestore-backed OSD will automatically associate itself to the "nvme" device class if it detects it's running on an NVMe. This adds support for this additional device class so that the non-rotational SCSI flag is properly configured.


dillaman:
@mikechristie can you give this a quick once-over? I tested it locally against both an SSD and NVMe labeled device class, so it functionally works.

PR #455, opened 2018-08-16
target:delete device node from devs before reopen device

tangwenji:


lxbsz:
Good catch.

PR #462, opened 2018-08-23
libtcmu:fix valid field in sense data

tangwenji:
If the valid field has been set to 1, it will be overwritten when setting the sense data.


mikechristie:
Nice. Find. Did you hit this with compare and write?

tangwenji:
yes。

PR #464, opened 2018-08-24
add checking of logging formats vs. arguments

DavidButterfield:
Add compiler hint
	__attribute__ ((format (printf, 4, 5)))
to tcmu_err_message(), et. al. so that the compiler will check the
arguments against the printf format string.

Also change several places in various source files to bring the formats
into conformance with the data types being formatted.

Signed-off-by: David Butterfield <David.Butterfield@wdc.com>

### DIFF
--- a/file_zbc.c
+++ b/file_zbc.c
@@ -13,6 +13,7 @@
 #define _GNU_SOURCE
 #include <stddef.h>
 #include <stdint.h>
+#include <inttypes.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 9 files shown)


file_zbc.c
```diff
@@ -1895,7 +1895,7 @@ static int zbc_check_rdwr(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	}
 
 	if (lba + nr_lbas > zdev->capacity || lba + nr_lbas < lba) {
-		tcmu_dev_err(dev, "cmd exceeds last lba %llu (lba %llu, xfer len %lu)\n",
+		tcmu_dev_err(dev, "cmd exceeds last lba %llu (lba %lu, xfer len %lu)\n",
```
mikechristie: For these 64 bit types on 32 bit userspace doesn't it need to be long long and not just long?

Should we be doing PRId64/PRIu64 so we do not have to worry?

DavidButterfield:
You are right -- I blindly followed the compiler's complaints without considering 32-bit.  I've reworked the changes to use the PRI*64 macros for uint64_t, %zu for size_t, etc.

qcow.c
```diff
@@ -1004,7 +1005,7 @@ static uint64_t qcow2_block_alloc(struct qcow_state *s, size_t size)
 	s->first_free_cluster = cluster + s->cluster_size;
 	// this causes a nasty loop
 	// qcow2_set_refcount(s, cluster, 1);
-	tcmu_dbg("  allocating cluster %d\n", cluster / s->cluster_size);
+	tcmu_dbg("  allocating cluster %"PRIu64"\n", cluster / s->cluster_size);
```
mikechristie: cluster_size is a "unsigned int", so it should be "%u" right? If that is correct, I can just fix it up for you when I merge. No need to respin your patches.

mikechristie:
Thanks for all work!

Is there a tool we should be running on new patches or was there a gcc flag we should be using to detect this?

qcow.c
DavidButterfield: The printf argument is cluster / s->cluster_size, and cluster is uint64_t.

PR #471, opened 2018-08-30
runner/rbd: misc implicit failover fixes

mikechristie:
1. Update lock state when getting a RTPG or INQUIRY to make sure when a
path is added back after a failover, we re-grab the lock later.

2. Use BUSY instead of ALUA STATE TRANSITION to work around a bug in the linux alua layer.

### DIFF
--- a/alua.c
+++ b/alua.c
@@ -387,8 +387,8 @@ static int alua_set_state(struct tcmu_device *dev, struct alua_grp *group,
  * @group_list: list of alua groups
  * @enabled_group_id: group id of the local enabled alua group
  *
- * If the handler is not able to update the remote nodes's state during STPG
- * handling we update it now.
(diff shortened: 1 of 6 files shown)


alua.c
```diff
@@ -387,8 +387,8 @@ static int alua_set_state(struct tcmu_device *dev, struct alua_grp *group,
  * @group_list: list of alua groups
  * @enabled_group_id: group id of the local enabled alua group
  *
- * If the handler is not able to update the remote nodes's state during STPG
- * handling we update it now.
+ * If the handler is not able to update the remote nodes's state during alua
```
dillaman: Nit: ```alua``` -> ```ALUA```

mikechristie: Thanks. Fixed.

Patches merged.

PR #473, opened 2018-09-04
log: buffer messages for early log calls by config and deamon setup

lxbsz:
We need to log all the messages into tcmu-runner log file.
...
2018-09-04 00:11:51.376 28011 [INFO] Starting...
2018-09-04 00:11:51.376 28011 [INFO] Inotify is watching "/etc/tcmu/tcmu.conf", wd: 6930144, mask: IN_ALL_EVENTS
2018-09-04 00:11:51.376 28011 [INFO] log_thread_start:587: Log thread finishes initializing!
2018-09-04 00:11:51.547 28011 [ERROR] reconfig_device:183 glfs/block0: Handler reconfig failed with error -95.
2018-09-04 00:11:51.550 28011 [ERROR] reconfig_device:183 glfs/block0: Handler reconfig failed with error -95.
...

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -29,6 +29,11 @@
 #define LOG_MSG_LEN (LOG_ENTRY_LEN - 1) /* the length of the log message */
 #define LOG_ENTRYS (1024 * 32)
 
+struct early_log_buf {
+	int pri;
+	char msg[LOG_MSG_LEN];
+};
+
 struct log_buf {
 	pthread_cond_t cond;
 	pthread_mutex_t lock;
@@ -39,6 +44,7 @@ struct log_buf {
 	unsigned int head;
 	unsigned int tail;
 	char buf[LOG_ENTRYS][LOG_ENTRY_LEN];
+	darray(struct early_log_buf) early_buf;
 	darray(struct log_output) outputs;
 	pthread_t thread_id;
 };
@@ -166,17 +172,24 @@ static void log_cleanup(void *arg)
 
 	pthread_mutex_lock(&tcmu_logbuf_lock);
 
+	if (!tcmu_logbuf)
+		goto unlock;
+
 	pthread_cond_destroy(&tcmu_logbuf->cond);
 	pthread_mutex_destroy(&tcmu_logbuf->lock);
 
-	darray_foreach(output, tcmu_logbuf->outputs)
-		log_cleanup_output(output);
+	if (!darray_empty(tcmu_logbuf->outputs)) {
+		darray_foreach(output, tcmu_logbuf->outputs)
+			log_cleanup_output(output);
+	}
 
 	darray_free(tcmu_logbuf->outputs);
+	darray_free(tcmu_logbuf->early_buf);
 
 	free(tcmu_logbuf);
 	tcmu_logbuf = NULL;
 
+unlock:
 	pthread_mutex_unlock(&tcmu_logbuf_lock);
 }
 
@@ -209,6 +222,27 @@ static void log_output(int pri, const char *msg, bool bypass)
 	}
 }
 
+static void buffer_log_messages(int pri, const char *fmt, va_list args, bool buffer)
+{
+	struct early_log_buf lb;
+	char buf[LOG_MSG_LEN];
+	ssize_t len;
+
+	len = vsnprintf(buf, LOG_MSG_LEN, fmt, args);
+
+	/* Handle early log calls by config and deamon setup */
+	fwrite(buf, 1, len, stderr);
+
+	if (!buffer)
+		return;
+
+	/* Save the log messages in the tmp buf */
+	lb.pri = pri;
+	strcpy(lb.msg, buf);
+
+	darray_append(tcmu_logbuf->early_buf, lb);
+}
+
 static void
 log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	     int linenr, const char *fmt, va_list args)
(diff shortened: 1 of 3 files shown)


lxbsz:
@pkalever  Please review, thanks.

pkalever:
@lxbsz looks good to me. Works now!

libtcmu_log.c
```diff
@@ -209,6 +209,34 @@ static void log_output(int pri, const char *msg, bool bypass)
...
+static void buffer_log_messages(int pri, const char *fmt, va_list args)
+{
+	struct log_buffer lb;
+	char buf[LOG_MSG_LEN];
+	ssize_t len;
+
+	len = vsnprintf(buf, LOG_MSG_LEN, fmt, args);
+
+	/* Handle early log calls by config and deamon setup */
+	write(2, buf, len);
```
mikechristie: Could you use the stdin/out/err def for this.

libtcmu_log.c
```diff
@@ -209,6 +209,34 @@ static void log_output(int pri, const char *msg, bool bypass)
 	}
 }
 
+struct log_buffer {
+	int pri;
+	char msg[LOG_MSG_LEN];
+};
+
+static darray(struct log_buffer) log_tmpbuf = darray_new();
+static pthread_mutex_t log_tmpbuf_lock = PTHREAD_MUTEX_INITIALIZER;
```
mikechristie: Can you move all these definitions to the top of the file with the rest of them?

libtcmu_log.c
```diff
@@ -209,6 +209,34 @@ static void log_output(int pri, const char *msg, bool bypass)
 	}
 }
 
+struct log_buffer {
```
mikechristie: Let's add a "early" prefix or something similar to avoid confusion with the other definitions that are similar. Also try to name them similar to the normal ones, so we would have

struct early_log_buf instead of log_buffer
tcmu_early_logbuf instead of log_tmpbuf
tcmu_early_logbuf_lock instead of log_tmpbuf_lock

main.c
```diff
@@ -972,6 +972,8 @@ int main(int argc, char **argv)
 	int fd;
 	int ret;
 
+	tcmu_info("Starting...\n");
```
mikechristie: Do you need this? Won't you get it from systemd too?

libtcmu_log.c
```diff
@@ -554,6 +582,14 @@ static void *log_thread_start(void *arg)
 	}
 	pthread_mutex_unlock(&tcmu_logbuf->lock);
 
+	pthread_mutex_lock(&log_tmpbuf_lock);
+	darray_foreach(lb, log_tmpbuf)
+		log_output(lb->pri, lb->msg, true);
+	darray_free(log_tmpbuf);
```
mikechristie: I think we have a couple leaks.

1. If we never get the thread started we never free the log_tmpbuf.

2. In main.c's main failure handling path we do:

```
destroy_log:
        tcmu_destroy_log();
destroy_config:
        tcmu_destroy_config(tcmu_cfg);

```

So the goto path in main.c that calls tcmu_destroy_config path could do 
tcmu_destroy_config -> tcmu_cancel_config_thread -> ... log_internal

libtcmu_log.c
```diff
@@ -548,6 +582,14 @@ static void *log_thread_start(void *arg)
...
+	pthread_mutex_unlock(&log_tmpbuf_lock);
+
+	tcmu_info("Log thread finishes initializing!\n");
```
mikechristie: This should be a debug message. Also fix up the wording:

tcmu_dbg("Log thread finished initializing!\n");

libtcmu_log.c
mikechristie: I think we can fix this by doing:

1. Just clear log_tmpbuf under the lock after we free it above. Then in log_cleanup add a check for if it is still set. If so then free it.

2. If log_tmpbuf is cleared then just drop down to vfprintf(stderr, fmt, args); calls in log_internal.

main.c
lxbsz: Yeah, we need it, in some container case, there no syslog service or even the journalctl command.
What we can get will only from the tcmu-runner.log. And mostly as default there isn't any log output for tcmu-runner in that log file, it seems the logger is not working for tcmu-runner, and for this case has been raised one bug by our QE.

libtcmu_log.c
lxbsz: Will fix it.

libtcmu_log.c
lxbsz: Sure.

libtcmu_log.c
lxbsz: Will update this.

libtcmu_log.c
lxbsz: Yeah, Will update this.

libtcmu_log.c
```diff
@@ -224,17 +251,17 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
 	if (!fmt)
 		return;
 
-	if (!tcmu_logbuf) {
-		/* handle early log calls by config and deamon setup */
-		vfprintf(stderr, fmt, args);
+	pthread_mutex_lock(&tcmu_logbuf_lock);
+	if (tcmu_logbuf_init()) {
```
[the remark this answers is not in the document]
lxbsz: Yeah, handled this all and updated it.

pkalever:
@lxbsz may be you want to close this, as the patches are already pulled at https://github.com/open-iscsi/tcmu-runner/pull/480

lxbsz:
> @lxbsz may be you want to close this, as the patches are already pulled at #480

Yeah, will do it.
Thanks.

PR #480, opened 2018-09-21
Log fixes

mikechristie:
A bunch of log fixes.

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -209,12 +209,26 @@ static void log_output(int pri, const char *msg, bool bypass)
 	}
 }
 
+static void log_queue_msg(struct log_buf *logbuf, int pri, char *buf)
+{
+	unsigned int head;
+	char *msg;
+
+	head = logbuf->head;
+	rb_set_pri(logbuf, head, pri);
+	msg = rb_get_msg(logbuf, head);
+	memcpy(msg, buf, LOG_MSG_LEN);
+	rb_update_head(logbuf);
+
+	if (logbuf->thread_active == false)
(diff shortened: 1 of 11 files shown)


mikechristie:
@lxbsz  please review.

It is probably easiest to just apply all the patches and review it, because there is so much churn with the patchset.

libtcmu_log.c
```diff
@@ -245,29 +239,18 @@ log_internal(int pri, struct tcmu_device *dev, const char *funcname,
...
-	 * Bypass the ringbuffer for some cases,
-	 * such as stdout and log file
+	 * Avoid overflowing the log buf with SCSI CDBs.
 	 */
-	log_output(pri, buf, true);
-
-	if (pri >= TCMU_LOG_DEBUG_SCSI_CMD)
-		goto unlock;
+	if (pri < TCMU_LOG_DEBUG_SCSI_CMD)
+		log_queue_msg(tcmu_logbuf, pri, buf);
```
lxbsz: BTW, before we will only output the TCMU_LOG_DEBUG_SCSI_CMD logs to stderr, and after your changes, it will output to the tcmu-runner.log file. Is this what you are trying to do, right ?

Before why I just output this to stderr was trying to avoid the tcmu-runner.log file increase to large. But I am okay with your change here or maybe we need a log file rotate ?

lxbsz:
@mikechristie 

Your changes look handsome :-).

I test it and all work well. Except the above changes all looks good to me and I do not reject that changes,.

Thanks.

libtcmu_log.c
lxbsz: Yeah, that sounds nice. :-)

PR #482, opened 2018-10-03
logger: notification on changing the options for dynamic reloading

lxbsz:

### DIFF
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -441,7 +441,7 @@ static int create_file_output(struct log_buf *logbuf, int pri,
 	if (!output) {
 		close(fd);
 		tcmu_err("Failed to create output file: %s\n", log_file_path);
-		return ret;
+		return -ENOMEM;
 	}
 
 	pthread_cleanup_push(cleanup_file_out_lock, logbuf);
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -95,6 +95,7 @@ void tcmu_set_log_level(int level)
 	else if (level < TCMU_CONF_LOG_LEVEL_MIN)
 		level = TCMU_CONF_LOG_LEVEL_MIN;
 
+	tcmu_info("log level now is %s\n", log_level_lookup[level]);
 	tcmu_log_level = to_syslog_level(level);
 }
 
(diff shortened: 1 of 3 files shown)


libtcmu_log.c
```diff
@@ -107,6 +107,7 @@ void tcmu_set_log_level(int level)
 	else if (level < TCMU_CONF_LOG_LEVEL_MIN)
 		level = TCMU_CONF_LOG_LEVEL_MIN;
 
+	tcmu_info("log level now is %d\n", level);
```
pkalever: Lets covert the 'level' to string form and then print. That way it makes more sense.

lxbsz: Yeah, make sense.

lxbsz:
@pkalever Done. Please review, thanks.

libtcmu_log.c
```diff
@@ -481,6 +482,7 @@ int tcmu_create_file_output(int pri, const char *filename, bool reloading)
 		return ret;
 	}
 
+	tcmu_info("log file path now is \"%s\"\n", log_file_path);
```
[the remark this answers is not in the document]
mikechristie: I think we should move this to the end of the function in case we fail.

If you want a message above I think it makes more sense as a debug message and something like

tcmu_dbg("Attempting to use %s as the log file path\n", log_file_path);

then at the end of the function when we are successful do:

tcmu_info("log file path is now \"%s\"\n", log_file_path);

libtcmu_log.c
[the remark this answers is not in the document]
lxbsz: Yeah, sorry, missed this changes, just forget to amend it.

libtcmu_log.c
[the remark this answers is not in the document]
lxbsz: @pkalever Looks good.

@mikechristie Yeah, make sense.

lxbsz:
@mikechristie @pkalever Updated.

libtcmu_log.c
```diff
@@ -455,7 +458,12 @@ static int create_file_output(struct log_buf *logbuf, int pri,
 	pthread_mutex_unlock(&logbuf->file_out_lock);
 	pthread_cleanup_pop(0);
 
+	tcmu_info("log file path now is '%s'\n", log_file_path);
 	return 0;
+
+err:
+	tcmu_dbg("Attempting to use '%s' as the log file path\n", log_file_path);
```
mikechristie: I think I might have miswrote my review comment last time. I had meant that this debug statement should go before the open() call. If we put it in the error path I think it will be confusing because we would see something like:

Failed to open
Attempting to use '%s' as the log file path

which would be out of order.

lxbsz: Yeah, make sense.

lxbsz:
@mikechristie Updated done.

PR #484, opened 2018-10-09
file_zbc: fix compile error and make build optional

dillaman:


file_zbc.c
```diff
@@ -2060,8 +2060,8 @@ static int zbc_write_check_zones(struct tcmu_device *dev,
 
 		/* Check LBA on write pointer */
 		if (zbc_zone_seq_req(zone) && lba != zone->wp) {
-			tcmu_dev_err(dev, "Unaligned write lba %"PRIu64", wp %llu\n",
-				     lba, zone->wp);
+			tcmu_dev_err(dev, "Unaligned write lba %"PRIu64", wp %"PRIu64"\n",
+				     lba, (uint64_t)zone->wp);
```
mikechristie: Oh nasty, I guess the __u64, uint64_t, llu and PRIu64 do not match up nicely on all archs.

@damien-lemoal, this patch seems ok to me to get it compiling again. I am guessing the __u64 use instead of uint64_t was to match the kernel structs. Let me know if you have issues with the patch.

Issue #485, opened 2018-10-18
Should we update the ringbuffer tail here if the cmd is timedout in ringbuffer in kernel space ?


lxbsz:
If one cmd in ringbuffer in kernel space is timedout, TCMU driver won't clean the idr node and also the ringbuffer entry for this cmd, until the ringbuffer tail is updated by tcmu-runner.

But for the case the cmd in glfs/ceph... backend stuck forever or a very long time and then there will be thousands of cmds are timedout, then will there be out of memories or memories leak?

mikechristie:
The problem with driving cleanup in the kernel when some timer like the tcmu cmd timeout fires is that you do not know what is going on in userspace so you can't free some of the memory. If you were to free the data block memory/page and reallocated it to another cmd, userspace could still be accessing it for the previous command.

The cmd's idr value is also a issue. If we free the tcmu_cmd->cmd_id and reallocate it to another cmd, but the first instance finishes then the reallocated command is going to hit issues.

For ceph we set a ceph timeout, osd_op_timeout, in the handler, so after osd_op_timeout seconds librbd will timeout the command and the handler will fail the cmd back to the kernel.

We could just have a generic per command timer in userspace for this where when it fires tcmu-runner calls into the handler to cleanup the command in the handler and in the backend if needed, then tcmu-runenr fails the command to the kernel.

lxbsz:
> The problem with driving cleanup in the kernel when some timer like the tcmu cmd timeout fires is that you do not know what is going on in userspace so you can't free some of the memory. If you were to free the data block memory/page and reallocated it to another cmd, userspace could still be accessing it for the previous command.
> 
> The cmd's idr value is also a issue. If we free the tcmu_cmd->cmd_id and reallocate it to another cmd, but the first instance finishes then the reallocated command is going to hit issues.
> 

Yeah, agree.

> For ceph we set a ceph timeout, osd_op_timeout, in the handler, so after osd_op_timeout seconds librbd will timeout the command and the handler will fail the cmd back to the kernel.
> 

That's cool in ceph. While there may have another case that when the network is stuck between tcmu-runner <--> ceph cluster, even there has the osd_op_timeout, we will still hitting this, right ?

> We could just have a generic per command timer in userspace for this where when it fires tcmu-runner calls into the handler to cleanup the command in the handler and in the backend if needed, then tcmu-runenr fails the command to the kernel.

[...]

PR #486, opened 2018-10-31
Update the ringbuffer tail here if the cmd is timedout in ringbuffer in kernel space

lxbsz:
Here we add one timer for each scsi command in tcmu-runner, and the time out value is (cmd_time_out + 5) seconds to make sure before the timer is fired in userspace here, the timer in ringbuffer in kernel space has already been fired.

Fixed: #485 

@mikechristie @pkalever Please review.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -1021,6 +1021,9 @@ struct tcmulib_cmd *tcmulib_get_next_command(struct tcmu_device *dev)
 			if (!cmd)
 				return NULL;
 			cmd->cmd_id = ent->hdr.cmd_id;
+			cmd->dev = dev;
+			cmd->timer = NULL;
+			cmd->timeout = false;
 
 			/* Convert iovec addrs in-place to not be offsets */
 			cmd->iov_cnt = ent->req.iov_cnt;
@@ -1176,32 +1179,40 @@ do { \
 void tcmulib_command_complete(
 	struct tcmu_device *dev,
 	struct tcmulib_cmd *cmd,
-	int result)
+	int result, bool timeout)
 {
 	struct tcmu_mailbox *mb = dev->map;
 	struct tcmu_cmd_entry *ent = (void *) mb + mb->cmdr_off + mb->cmd_tail;
 
-	/* current command could be PAD in async case */
-	while (ent != (void *) mb + mb->cmdr_off + mb->cmd_head) {
-		if (tcmu_hdr_get_op(ent->hdr.len_op) == TCMU_OP_CMD)
-			break;
+	if (!cmd->timeout) {
+		/* current command could be PAD in async case */
+		while (ent != (void *) mb + mb->cmdr_off + mb->cmd_head) {
+			if (tcmu_hdr_get_op(ent->hdr.len_op) == TCMU_OP_CMD)
+				break;
+			TCMU_UPDATE_RB_TAIL(mb, ent);
+			ent = (void *) mb + mb->cmdr_off + mb->cmd_tail;
+		}
+
+		/* cmd_id could be different in async case */
+		if (cmd->cmd_id != ent->hdr.cmd_id) {
+			ent->hdr.cmd_id = cmd->cmd_id;
+		}
+
+		ent->rsp.scsi_status = tcmu_sts_to_scsi(result, cmd->sense_buf);
+		if (ent->rsp.scsi_status == SAM_STAT_CHECK_CONDITION) {
+			memcpy(ent->rsp.sense_buffer, cmd->sense_buf,
+					TCMU_SENSE_BUFFERSIZE);
+		}
+
 		TCMU_UPDATE_RB_TAIL(mb, ent);
-		ent = (void *) mb + mb->cmdr_off + mb->cmd_tail;
 	}
 
-	/* cmd_id could be different in async case */
-	if (cmd->cmd_id != ent->hdr.cmd_id) {
-		ent->hdr.cmd_id = cmd->cmd_id;
-	}
+	cmd->timeout = timeout;
 
-	ent->rsp.scsi_status = tcmu_sts_to_scsi(result, cmd->sense_buf);
-	if (ent->rsp.scsi_status == SAM_STAT_CHECK_CONDITION) {
-		memcpy(ent->rsp.sense_buffer, cmd->sense_buf,
-		       TCMU_SENSE_BUFFERSIZE);
+	if (!timeout) {
+		free(cmd->timer);
+		free(cmd);
 	}
-
-	TCMU_UPDATE_RB_TAIL(mb, ent);
-	free(cmd);
 }
 
 void tcmulib_processing_start(struct tcmu_device *dev)
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -814,6 +814,16 @@ uint64_t tcmu_get_dev_num_lbas(struct tcmu_device *dev)
 	return dev->num_lbas;
 }
 
+void tcmu_set_dev_cmd_time_out(struct tcmu_device *dev, unsigned int cmd_time_out)
+{
+	dev->cmd_time_out = cmd_time_out;
+}
+
+unsigned int tcmu_get_dev_cmd_time_out(struct tcmu_device *dev)
+{
+	return dev->cmd_time_out;
+}
+
 /**
  * tcmu_update_num_lbas - Update num LBAs based on the new size.
  * @dev: tcmu device to update
--- a/libtcmu_timer.c
+++ b/libtcmu_timer.c
@@ -1,0 +1,365 @@
+/*
+ * tcmu timer wheel
+ *
+ * Copyright (C) 1991, 1992  Linus Torvalds
+ * Copyright (c) 2018 Red Hat, Inc.
+ *
+ * This file is licensed to you under your choice of the GNU Lesser
+ * General Public License, version 2.1 or any later version (LGPLv2.1 or
+ * later), or the Apache License 2.0.
+ *
+ * Most of the code is from glusterfs project and which is from Linux
+ * kernel's internal timer wheel driver.
+ */
+
+#define _GNU_SOURCE
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <sys/select.h>
+#include <errno.h>
+
+#include "libtcmu_timer.h"
+#include "tcmu-runner.h"
+
+#define TVR_BITS  8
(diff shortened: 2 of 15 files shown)


bgly:
Do you need to consider the case where qfull_timeout hits which then calls tcmu_check_expired_cmd() in kernel? That would then time out all the commands in the queue if that timer is hit right? Do you need to update the tail ringbuffer in userspace in that case?

mikechristie:
For the case where the qfull or cmd timer fires and tcmu_check_expired_cmd runs and there are expired running commands, the kernel does not clear the ring, because we do not have any synchronization with userspace, so we do not know if it is accessing the pages or not. And then we do not clear the ring for non-running commands, because they are not in the ring yet.

We could add some sort of lock/bits for the running cmd case, but if runner is dead then there is normally no point. The commands would just keep timing out until you hit the initiator side retry limit so either way the path/device is going to go down/offline. Note, when runner restarts then it will clean up the ring in newer kernels.

If runner is up and if a command timesout in userspace by the timer in xiubo's patches then it would complete the command with a failure to the kernel. This will free up the space in the ring, and the kernel would take commands from the cmdr_queue and put them in the ring or execute new commands.

lxbsz:
Yeah, that's right.

And also this fix will depend on the following kernel fix: 

```
scsi: tcmu: avoid cmd/qfull timers updated whenever a new cmd comes
https://patchwork.kernel.org/patch/10644889/
```

Without this, there will be the ringbuffer broken issue due to the cmd_time_out timer is not working as it is expected.

libtcmu.c
```diff
@@ -1007,10 +1017,12 @@ struct tcmulib_cmd *tcmulib_get_next_command(struct tcmu_device *dev)
 			}
 
 			/* Alloc memory for cmd itself, iovec and cdb */
-			cmd = malloc(sizeof(*cmd) + sizeof(*cmd->iovec) * ent->req.iov_cnt + cdb_len);
+			cmd = calloc(1, sizeof(*cmd) + sizeof(*cmd->iovec) * ent->req.iov_cnt + cdb_len);
 			if (!cmd)
 				return NULL;
 			cmd->cmd_id = ent->hdr.cmd_id;
+			cmd->dev = dev;
+			pthread_spin_init(&cmd->lock, 0);
```
lxbsz: comment it myself here:

The spin lock is not used, will delete it.

main.c
```diff
@@ -643,7 +644,25 @@ static void *tcmur_cmdproc_thread(void *arg)
...
+					tcmu_dev_err(dev, "malloc timer failed!\n");
+					return NULL;
+				}
+
+				timer->data = cmd;
+				timer->expires = tcmu_get_dev_cmd_time_out(dev) + 5;
+				timer->function = tcmur_cmd_timeout;
+				cmd->timer = timer;
+
+				tcmu_add_timer(timer);
```
lxbsz: There will be one problem here too:

If the glfs or other backend quick enough, and just before here the cmd->done() is called then the cmd maybe released, so the cmd deferences will be segment crash.

lxbsz:
Update it done.

lxbsz:
@mikechristie @pkalever @bgly 
Test done.

tcmur_cmd_handler.c
```diff
@@ -2267,6 +2279,22 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	/* setup timer for async handled commands */
+	if (!rhandler->has_timer) {
+		timer = malloc(sizeof(*timer));
+		if (!timer) {
+			tcmu_dev_err(dev, "malloc timer failed!\n");
+			goto untrack;
+		}
+
+		timer->data = cmd;
+		timer->expires = tcmu_get_dev_cmd_time_out(dev) + 5;
```
mikechristie: Let's go ahead and just add a tcmu-runner specific timer just in case users will need this set differently from the kernel one. We might have where they are doing tape, and they cannot enable the kernel timer, but they want the userspace one, because we have more flexibility and can one day cancel/abort/cleanup commands on the physical backend.

We should be able to add this into the existing dev_config string with no kernel changes.

For rtslib UserBackedStorageObject:__init__() we might want to add a new arg for these type arguments for the daemon.

_configure() can then add these new args before the device ones. Something like:

tcmu-runner_args=cmd_timeout=$N;setting123=$ABC;setting345=$DEF/handler_args=what_is_getting_pased_to_the_handler_today

libtcmu.c_dev_added can then just chop its args out and store them in tcmu_device. tcmu_get/set_dev_cmd_time_out then just interact with the tcmu_device.

tcmur_cmd_handler.c
```diff
@@ -2267,6 +2279,22 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		break;
 	}
```
mikechristie: We need to move this down to the lib level code or move it up the code in tcmulib_command_complete to tcmur_command_complete so they are the same level for when we do the lib split.

The other calls like tcmu_init/cleanup_timer_base then should stay in runner or move to the lib.

Either the lib should do it for the lib user (tcmu-runner) or the lib does not know about it and tcmu-runner does it all.

libtcmu_timer.c
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	pthread_spin_unlock(&timer_base_lock);
+}
+
+void *runner(void *arg)
+{
+	struct timeval tv = {0,};
+
+	while(1) {
```
mikechristie: Add a space between while and the opening "("

tcmur_cmd_handler.c
```diff
@@ -33,33 +34,42 @@ static void _cleanup_spin_lock(void *arg)
...
 		tcmulib_processing_complete(dev);
 		track_aio_wakeup_finish(rdev, &wake_up);
 	}
 }
 
+void tcmur_cmd_timeout(struct tcmu_timer *timer, void *data)
+{
+	struct tcmulib_cmd *cmd = (struct tcmulib_cmd *)data;
+	struct tcmu_device *dev = cmd->dev;
+
```
mikechristie: For glfs, do you need to perform any cleanup in the handler or is ok to leave it running?

tcmur_cmd_handler.c
```diff
@@ -33,33 +34,42 @@ static void _cleanup_spin_lock(void *arg)
...
+	aio_command_finish(dev, cmd, TCMU_STS_HW_ERR, true);
```
mikechristie: I think we will want to use something that is less harsh.

For linux it looks like it is treated as a path error for the normal RW case.

Do you know how windows and ESX handle this error?

tcmur_cmd_handler.c
mikechristie: Actually, I am a dummy :) There is no need to add the extra init() arg and make daemon args and all that extra stuff.

Just add it onto the normal hander args, have the handle parse it and call tcmu_set_dev_cmd_time_out() to set the value.

We can then drop the has_timer field.

tcmur_cmd_handler.c
mikechristie: Oh yeah, ignore this comment. Feel free to keep it as HW_ERR since you are the only user. Maybe you want to change it to LUN_COMM error like is used in the kernel when the tcmu cmd time out fires.

tcmur_cmd_handler.c
mikechristie: We should not make it tcmu.conf based because different devices may need different values.

Just pass it in with the other handler info on the cfg string.

tcmur_cmd_handler.c
mikechristie: No I meant this function will clean up the tcmu-runner/libtcmu bits but it leaves the command running in glfs. Do you want to call into glfs.c and kill the command in there?

For example, if you ran each command in a thread and cmd1 is in thread1, it's timer fires and libtcmu/runner fail the command, but it is still running on thread1. The initiator gets the failed command decides to retry. That cmd instance is run on thread2 at the same time thread1 is still running. Thread2 completes, then the initiator sends a new command to the same sectors so now the old command on thread1 is running at the same time this new command is. Thread1 might then overwrite newer data.

I was just asking if glfs handles those type of things already or if you need to call into glfs.c to tell it to stop/kill the command that is running on thread1?

mikechristie: For a similar problem but in tcmu-runner core, I think we need to add checks to the compound commands's done functions.

For example, for a xcopy we could timeout during the read part of the command, return failure to the initiator, but then glfs returns success from the read part, calls the done function and then starts the write part while possibly another xcopy is now started.

For CAW we could have another problem where the read part timeout, we return failure to the initiator, but we leave the caw_lock held so future CAWs are still blocked on that stuck original CAW.

tcmur_cmd_handler.c
lxbsz: Okay.

tcmur_cmd_handler.c
lxbsz: > No I meant this function will clean up the tcmu-runner/libtcmu bits but it leaves the command running in glfs. Do you want to call into glfs.c and kill the command in there?
> 
> For example, if you ran each command in a thread and cmd1 is in thread1, it's timer fires and libtcmu/runner fail the command, but it is still running on thread1. The initiator gets the failed command decides to retry. That cmd instance is run on thread2 at the same time thread1 is still running. Thread2 completes, then the initiator sends a new command to the same sectors so now the old command on thread1 is running at the same time this new command is. Thread1 might then overwrite newer data.
> 
> I was just asking if glfs handles those type of things already or if you need to call into glfs.c to tell it to stop/kill the command that is running on thread1?

Yeah, while there is one GLUSTER ping timeout in glusterfs backend, I am still investigating whether will the IO threads be cleaned after this timeout. We meant to set the timeout value here will be larger than the GLUSTER ping timeout value.

lxbsz: > For a similar problem but in tcmu-runner core, I think we need to add checks to the compound commands's done functions.
> 
> For example, for a xcopy we could timeout during the read part of the command, return failure to the initiator, but then glfs returns success from the read part, calls the done function and then starts the write part while possibly another xcopy is now started.
> 
> For CAW we could have another problem where the read part timeout, we return failure to the initiator, but we leave the caw_lock held so future CAWs are still blocked on that stuck original CAW.

This also make sense. Will fix it later.

mikechristie:
@lxbsz 

I am closing this for now. Reopen or make a new PR against:

We can add a handler callout that is called in check_for_timed_out_cmds, or call a function to do the ring cleanup from there, etc.

PR #487, opened 2018-11-01
tcmu: Simplify command completion

bgly:
Instead of requiring handlers with nr_thread > 0 to call
cmd->done we can call in in the io_work_queue, which then
allows for the handlers to pass back any return code.
At the moment handlers can only pass back TCMU_STS_OK


bgly:
@mikechristie

mikechristie:
Nice. Thanks.

PR #488, opened 2018-11-01
tcmu: Fix CMake to install file_optical

bgly:


bgly:
@mikechristie

mikechristie:
Can you make the building also installation of file_optical optional like the non example handlers? Do a with-file-optical option like glfs/rbd/qcow/zbc.

bgly:
On vacation, but I’ll do try to do it on the phone. I’ll also make the change for file_example.

bgly:
@mikechristie Changes are made

PR #497, opened 2018-11-15
spec-file: further changes to take options to ignore dependencies

amarts:
One can now pass '--without gluster' or '--without rbd' options
to rpmbuild now.

The same has been adopted in make_runnerrpms.sh file.

Signed-off-by: Amar Tumballi <amarts@redhat.com>

### DIFF
--- a/extra/make_runnerrpms.sh
+++ b/extra/make_runnerrpms.sh
@@ -1,11 +1,11 @@
 #!/bin/sh
 
-if [ $# -gt 0 -o "$1" == "--help" -o "$1" == "help" ]; then
+if [ "$1" == "--help" -o "$1" == "help" ]; then
 	echo ""
 	echo "  USAGE:"
 	echo ""
 	echo "  # cd tcmu-runner/extra/"
-	echo "  # ./make_runnerrpms.sh"
+	echo "  # ./make_runnerrpms.sh [--without (rbd|glfs|qcow|zbc|fbo)]"
 	echo ""
 	echo "  Will build the RPMs in current dir by using the HEAD commit ID as default."
 	echo ""
@@ -62,4 +62,4 @@ cd $TOPDIR/extra
 rm -rf $TMPDIR
 
 # Build the RPMs
-rpmbuild --define="_topdir $rpmbuild_path" -ba $rpmbuild_path/SPECS/tcmu-runner.spec
+rpmbuild --define="_topdir $rpmbuild_path" -ba $rpmbuild_path/SPECS/tcmu-runner.spec "$@"
--- a/tcmu-runner.spec
+++ b/tcmu-runner.spec
@@ -1,3 +1,32 @@
+%global _hardened_build 1
+
+# without rbd dependency
+# if you wish to exclude rbd handlers in RPM, use below command
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without rbd
+%{?_without_rbd:%global _without_rbd -Dwith-rbd=false}
+
+# without glusterfs dependency
+# if you wish to exclude glfs handlers in RPM, use below command
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without glfs
+%{?_without_glfs:%global _without_glfs -Dwith-glfs=false}
+
+# without qcow dependency
+# if you wish to exclude qcow handlers in RPM, use below command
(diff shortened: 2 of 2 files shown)


amarts:
@lxbsz @mikechristie @pkalever PTAL

extra/make_runnerrpms.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/sh
 
-if [ $# -gt 0 -o "$1" == "--help" -o "$1" == "help" ]; then
+if [ "$1" == "--help" -o "$1" == "help" ]; then
 	echo ""
 	echo "  USAGE:"
 	echo ""
 	echo "  # cd tcmu-runner/extra/"
-	echo "  # ./make_runnerrpms.sh"
+	echo "  # ./make_runnerrpms.sh [--without rbd] [--without gluster] [--without qcow] [--without zbc]"
```
lxbsz: Will it be better to using glfs instead of gluster here ? Which will be keep the same with the handler name.

amarts: I thought of it, but somehow gluster sounded better. I can make that change. No issues

lxbsz: And I am also okay with this.

tcmu-runner.spec
```diff
@@ -1,3 +1,29 @@
+%global _hardened_build 1
+
+# uncomment and add '%' to use the prereltag for pre-releases
+# %%global prereltag qa3
+
+# without librbd1 dependency
+# if you wish to compile an rpm without librbd1...
```
pkalever: Probably you mean to say.
\# if you wish to exclude rbd handler at rpm

Same at other places too

tcmu-runner.spec
```diff
@@ -1,3 +1,29 @@
+%global _hardened_build 1
+
+# uncomment and add '%' to use the prereltag for pre-releases
+# %%global prereltag qa3
```
pkalever: where are we using prereltag ?

I understand you want to use prereltag at 'Release:' and/or with 'Source:', but I don't see its usage in this file.

amarts: Done, removed.

tcmu-runner.spec
amarts: Done

tcmu-runner.spec
```diff
@@ -1,3 +1,29 @@
...
+
+# without librbd1 dependency
+# if you wish to compile an rpm without librbd1...
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without rbd
+%{?_without_rbd:%global _without_rbd -Dwith-rbd=false}
+
+# without glusterfs dependency
+# if you wish to compile an rpm without glusterf api...
```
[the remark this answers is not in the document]
amarts: Removed the comment to reflect new wordings.

tcmu-runner.spec
```diff
@@ -1,3 +1,26 @@
...
+# without qcow dependency
+# if you wish to exclude qcow handlers in RPM, use below command
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without qcow
+%{?_without_qcow:%global _without_qcow -Dwith-qcow=false}
+
+# without zbc dependency
+# if you wish to exclude zbc handlers in RPM, use below command
+# rpmbuild -ta @PACKAGE_NAME@-@PACKAGE_VERSION@.tar.gz --without zbc
+%{?_without_zbc:%global _without_zbc -Dwith-zbc=false}
+
```
mikechristie: Another option was added the other day for file_optical.

amarts: Thanks for pointing this out. I checked it out, and currently there is no option to provide a cmake option for that through `CMakeLists.txt` file. Hence I am thinking I should do it as a separate PR to provide option in cmake, and then option in rpm and make_runnerrpms.sh files.

If that if fine, I will create an issue, and work on that separately. So, this patch can be merged as is.

mikechristie: We really just added it the other day, so maybe you need to do a repull/clone. See this commit:

https://github.com/open-iscsi/tcmu-runner/commit/daa00063045fafdfc25a13a3d94e2ea77c8a1c3d

mikechristie: Ah you know what, it is called fbo (file backed optical) in the cmakelist file which is the abbreviation used for it in the file_optical.c file and what is used for its identifier, so that is probably the confusion.

amarts: Oops, sorry about not being on tip of master. Source of all confusion. Added the option, and pushed again.

PR #502, opened 2018-11-23
logger: add critical log level support

lxbsz:
When changing the loglevel in the tcmu.conf for the dynamic logger,
we need to verify it is really working. Currently this hit log is
one INFO level message, which won't work in the ERR and WARN loglevel
cases.

### DIFF
--- a/libtcmu_config.h
+++ b/libtcmu_config.h
@@ -23,7 +23,8 @@ struct tcmu_config {
 };
 
 /*
- * There are 5 logging levels supported in tcmu.conf:
+ * There are 6 logging levels supported in tcmu.conf:
+ *    0: CRIT
  *    1: ERROR
  *    2: WARNING
  *    3: INFO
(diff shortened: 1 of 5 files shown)


lxbsz:
@mikechristie @pkalever 
Please review, thanks.

libtcmu_log.h
```diff
@@ -14,6 +14,7 @@
 #include <stdarg.h>
 #include <stdbool.h>
 
+#define TCMU_LOG_CRIT	LOG_CRIT
```
pkalever: just throw a comment at the end, to look alike others

lxbsz: Sure, will do it.
Thanks

lxbsz:
@pkalever Yeah, update and exposed it.
Thanks.

tcmu.conf
```diff
@@ -2,6 +2,7 @@
 
 # Logging Controls
 # There are 5 logging levels supported:
```
pkalever: 6 logging levels now.

libtcmu_config.h
```diff
@@ -31,7 +31,8 @@ struct tcmu_config {
  *    5: DEBUG SCSI CMD
  */
```
pkalever: Can you add the new loglevel here and change the logging levels to 6

tcmu.conf
lxbsz: will fix it.

libtcmu_config.h
lxbsz: will fix it.
Thanks.

PR #504, opened 2018-11-23
config: do not update the logdir when it's commented in conf

lxbsz:
Or it will always ouput the following logs even we uncomment the
logdir options in tcmu.conf:

2018-11-23 [...] create_file_output:475: log file path now is '/var/log/tcmu-runner.log'


lxbsz:
@pkalever Thanks, just fixed and updated the type errors of the commit comment.

PR #505, opened 2018-11-26
systemd-unit/rpm-spec/cmakeList: add rsyslog to requires list

pkalever:

### DIFF
--- a/tcmu-runner.spec
+++ b/tcmu-runner.spec
@@ -53,7 +53,7 @@ BuildRequires: glusterfs-api-devel
 Requires(pre): glusterfs-api
 %endif
 
-Requires(pre): kmod, zlib, libnl3, glib2, logrotate
+Requires(pre): kmod, zlib, libnl3, glib2, logrotate, rsyslog
 
 %description
 A daemon that handles the userspace side of the LIO TCM-User backstore.
--- a/tcmu-runner.spec
+++ b/tcmu-runner.spec
@@ -121,13 +121,3 @@ userspace libraries they like.
 %dir %{_sysconfdir}/tcmu/
 %config %{_sysconfdir}/tcmu/tcmu.conf
 %config(noreplace) %{_sysconfdir}/logrotate.d/tcmu-runner
-
-%changelog
-* Thu Nov 15 2018 Amar Tumballi <amarts@redhat.com>
-- Add options to exclude few dependencies while doing rpmbuild
-
-* Tue Nov 06 2018 Amar Tumballi <amarts@redhat.com>
-- Fix build errors
-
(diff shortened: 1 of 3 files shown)


pkalever:
@lxbsz @mikechristie PTAL, thanks!

pkalever:
Updated this PR to reflect the dependency in systemd unit, rpm spec and cmakeList

Thanks!

mikechristie:
Why does it require rsyslog? What happens if it is not there?

I ask because in libtcmu_log.c if create_syslog_output fails we just try to continue on, so that makes it look like it is optional.

tcmu-runner.spec
```diff
@@ -123,6 +123,9 @@ userspace libraries they like.
 %config(noreplace) %{_sysconfdir}/logrotate.d/tcmu-runner
 
 %changelog
+* Mon Nov 26 2018 Prasanna Kumar Kalever <prasanna.kalever@redhat.com>
```
mikechristie: I am thinking maybe these changelog additions are not very useful, because we have the git history for all changes already. Are they for us or for users I mean?

For each release we are not updating the changelog so users are not getting the full benefit of seeing what is changing from release to release. They are just seeing what changes in the rpm which is not very useful compared to getting a list of bugs and features in each list.

I would propose just stop adding these lines and maybe for each release I will add a proper changelog for major features/bugs or maybe for each release in the changelog I will add link to the GitHub release notes which has a lis of major changes?

pkalever:
> Why does it require rsyslog? What happens if it is not there?

As we are using syslog service, it should be our init scripts duty to start the service IMO.

> 
> I ask because in libtcmu_log.c if create_syslog_output fails we just try to continue on, so that makes it look like it is optional.

Yes, I agree with you. Probably making rsyslog.service an optional will be good enough.

```
# man systemd.unit
 Wants=
     A weaker version of Requires=. Units listed in this option will be started if the
     configuring unit is. However, if the listed units fail to start or cannot be added to the
     transaction, this has no impact on the validity of the transaction as a whole. This is the
     recommended way to hook start-up of one unit to the start-up of another unit.

     Note that dependencies of this type may also be configured outside of the unit
     configuration file by adding symlinks to a .wants/ directory accompanying the unit file.
     For details, see above.

```

I will update the PR, to reflect this.

Thanks @mikechristie

tcmu-runner.spec
pkalever: As long as we are caring to update the changelog at rpm packaging (i.e. at fedora, centos, rhel), I'm okay removing this.

I will submit a cleanup patch. Thanks!

pkalever:
@mikechristie updated the PR, PTAL. Thanks!

PR #506, opened 2018-11-30
logger: Fix inotify + honor the defaults and precedence of various ways to configure options

pkalever:

### DIFF
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -350,7 +350,7 @@ static void tcmu_parse_options(struct tcmu_config *cfg, char *buf, int len)
 	tcmu_conf_set_options(cfg);
 }
 
-static int tcmu_load_config(struct tcmu_config *cfg)
+int tcmu_load_config(struct tcmu_config *cfg)
 {
 	int ret = -1;
 	int fd, len;
@@ -452,32 +452,6 @@ static void *dyn_config_start(void *arg)
 	return NULL;
 }
 
-struct tcmu_config *tcmu_parse_config(const char *path)
-{
-	struct tcmu_config *cfg;
-
-	cfg = calloc(1, sizeof(*cfg));
-	if (cfg == NULL) {
-		tcmu_err("Alloc TCMU config failed!\n");
-		return NULL;
-	}
-
-	if (!path)
-		path = "/etc/tcmu/tcmu.conf"; /* the default config file */
-
-	snprintf(cfg->path, PATH_MAX, path);
-	if (tcmu_load_config(cfg)) {
-		tcmu_err("Loading TCMU config failed!\n");
-		goto free_cfg;
-	}
-
-	return cfg;
-
-free_cfg:
-	free(cfg);
-	return NULL;
-}
-
 int tcmu_watch_config(struct tcmu_config *cfg)
 {
 	return pthread_create(&cfg->thread_id, NULL, dyn_config_start, cfg);
@@ -486,6 +460,26 @@ int tcmu_watch_config(struct tcmu_config *cfg)
 void tcmu_unwatch_config(struct tcmu_config *cfg)
 {
 	tcmu_thread_cancel(cfg->thread_id);
+}
+
+struct tcmu_config* tcmu_initialize_config(void)
+{
+	struct tcmu_config *cfg;
+	char *log_dir;
+
+	cfg = calloc(1, sizeof(*cfg));
+	if (cfg == NULL) {
+		tcmu_err("allocating TCMU config failed: %m\n");
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	log_dir = getenv("TCMU_LOGDIR");
+	snprintf(cfg->log_dir, PATH_MAX, log_dir ? log_dir : TCMU_LOG_DIR_DEFAULT);
+	cfg->log_level = TCMU_CONF_LOG_INFO;
+	snprintf(cfg->path, PATH_MAX, TCMU_CONFIGFILE_DEFAULT);
+
+	return cfg;
 }
 
 void tcmu_free_config(struct tcmu_config *cfg)
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -88,11 +88,6 @@ struct tcmu_conf_option {
  *		TCMU_PARSE_CFG_INT(cfg, log_level);
  *		TCMU_CONF_CHECK_LOG_LEVEL(log_level);
  *	}
- * 4, Then add your own free method in if it's a STR KEY:
- *	static void tcmu_conf_free_str_keys(struct tcmu_config *cfg)
- *	{
- *		TCMU_FREE_CFG_STR_KEY(cfg, 'STR KEY');
- *	}
  *
  * Note: For now, if the options have been changed in config file, the
  * system config reload thread daemon will try to update them for all the
@@ -143,20 +138,14 @@ do { \
 	struct tcmu_conf_option *option; \
 	char buf[1024]; \
 	option = tcmu_get_option(#key); \
+	memset(cfg->key, 0, sizeof(cfg->key)); \
 	if (option) { \
-		if (cfg->key) \
-			free(cfg->key); \
-		cfg->key = strdup(option->opt_str); \
+		snprintf(cfg->key, sizeof(cfg->key), option->opt_str); \
 		if (option->opt_str) \
 			free(option->opt_str); \
 		sprintf(buf, "%s", def); \
 		option->opt_str = strdup(buf); \
 	} \
-} while (0);
-
-#define TCMU_FREE_CFG_STR_KEY(cfg, key) \
-do { \
-	free(cfg->key); \
 } while (0);
 
 static void tcmu_conf_set_options(struct tcmu_config *cfg)
@@ -166,21 +155,10 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg)
 	tcmu_set_log_level(cfg->log_level);
 
 	/* set log_dir path option */
-	TCMU_PARSE_CFG_STR(cfg, log_dir_path, TCMU_LOG_DIR_DEFAULT);
-	if (cfg->log_dir_path)
-		tcmu_resetup_log_file(cfg->log_dir_path);
+	TCMU_PARSE_CFG_STR(cfg, log_dir, TCMU_LOG_DIR_DEFAULT);
+	tcmu_resetup_log_file(cfg->log_dir);
 
 	/* add your new config options */
-}
-
-static void tcmu_conf_free_str_keys(struct tcmu_config *cfg)
-{
-	/* add your str type config options
-	 *
-	 * For example:
-	 * TCMU_FREE_CFG_STR_KEY(cfg, 'STR KEY');
-	 */
-	TCMU_FREE_CFG_STR_KEY(cfg, log_dir_path);
 }
 
 #define TCMU_MAX_CFG_FILE_SIZE (2 * 1024 * 1024)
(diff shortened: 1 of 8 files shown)


pkalever:
please-review: @mikechristie @lxbsz 

Thanks!

pkalever:
I see there are some more potential issues on the logger code, have the patches locally, will test them and send the patch[s] over the weekends or on Monday. Please hold the review on this PR for now. Thanks!

lxbsz:
> @lxbsz agree. I have a patch to fix this already. Will update after few tests Locally.

Cool Prasanna.

pkalever:
@lxbsz @mikechristie @amarts please review, thanks!

libtcmu_config.c
```diff
@@ -404,54 +404,63 @@ static void *dyn_config_start(void *arg)
...
 
 	monitor = inotify_init();
 	if (monitor == -1) {
 		tcmu_err("Failed to init inotify %m\n");
 		return NULL;
 	}
 
-	wd = inotify_add_watch(monitor, cfg->path, IN_ALL_EVENTS);
+	snprintf(cfg_dir, PATH_MAX, cfg->path);
+	p = strrchr(cfg_dir, '/');
```
lxbsz: Before it will allowed us to use /AA/BB, but now only /AA/BB/ is allowed.
Is this okay ?

pkalever:
V3: just removed free_opt label in main.c ", as part of "config: deprecate tcmu_parse_config"

libtcmu_config.c
pkalever: BTW - this is just conf file's dir path, not the log_dir path.

lxbsz:
Due to the difference of the editor verison, the editor will trigger the MODITY event once or multi time, I am okay with this.

libtcmu_config.c
lxbsz: yeah.

CMakeLists.txt
```diff
@@ -278,11 +278,10 @@ configure_file (
   "${PROJECT_SOURCE_DIR}/version.h"
   )
 
-configure_file (
-  "${PROJECT_SOURCE_DIR}/tcmu.conf_install.cmake.in"
-  "${PROJECT_SOURCE_DIR}/tcmu.conf_install.cmake"
-  )
-install(SCRIPT tcmu.conf_install.cmake)
+if (EXISTS /etc/tcmu/tcmu.conf)
+	file(RENAME /etc/tcmu/tcmu.conf /etc/tcmu/tcmu.conf.old)
```
lxbsz: Maybe we should use the timestamp instead of old:

+ string(TIMESTAMP newname "tcmu.conf-%Y-%m-%dT%H:%M:%S" UTC)                                                                                                            + file(RENAME /etc/tcmu/tcmu.conf /etc/tcmu/${newname})

lxbsz:
Test this and it works to me.

CMakeLists.txt
pkalever: $ cmake . && make install -> will get a new time stamp and works as we expect.
$ make install -> just make install without cmake, will always use old timestamp generated during cmake, and cause override of old original timestamp file with new tcmu.conf

pkalever:
V5: Added 2 other patches
* logger: skip resetting options if they match with current
* load_config: retry opening conf-file if the initial attempt fail

@lxbsz please have a look. Thanks!

libtcmu_log.c
```diff
@@ -692,10 +697,31 @@ int tcmu_setup_log(char *log_dir)
...
+	return st1.st_dev == st2.st_dev && st1.st_ino == st2.st_ino;
+}
+
 int tcmu_resetup_log_file(char *log_dir)
 {
 	int ret;
 
+	if (is_path_same(tcmu_log_dir, log_dir)) {
+		tcmu_dbg("Skipping resetting log_dir, as it matches with "
+		         "current log dir path: %s.\n", log_dir);
```
lxbsz: Usually the log format won't be wrapped and split into 2 or more lines.
We may make it shorter like:
```

tcmu_dbg("Nothing changes for log dir: %s, skip it.\n",
                  log_dir)
```

pkalever: @lxbsz 
actually it prints in one single line. there is no '\n' in between the lines ?
Or are you saying lets just make it short ?

pkalever:
V6: just updated patch "cmake_install: backup old tcmu.conf version as tcmu.conf.old"  to move changes in tcmu.conf_install.cmake.in rather than in CMakeLists.txt

libtcmu_log.c
lxbsz: Yeah, it is. I meant in the code.

pkalever:
V7: adjusted log messages to single line in patch "logger: skip resetting options if they match with current" as per @lxbsz comments. Thanks!

main.c
```diff
@@ -981,10 +1001,10 @@ int main(int argc, char **argv)
 	bool new_path = false;
 	bool watching_cfg = false;
 	struct flock lock_fd = {0, };
-	char *log_dir = NULL;
 	int fd;
 	int ret;
 
+	tcmu_cfg = tcmu_initiatize_options();
```
lxbsz: Please defend the possible NULL pointer, or the crash may happen, the crash will be much harder to debug .

pkalever:
V8: updated patch "config: deprecate tcmu_parse_config", defending for possible NULL pointer return from tcmu_initiatize_options(); as per @lxbsz comment.

Thanks!

pkalever:
@mikechristie we have tested this series thoroughly today. Please take a look. Thanks!

tcmu.conf
```diff
@@ -15,4 +15,4 @@
 # Logging Directory Path
 # The default logging Directory path is /var/log, uncomment it
 # and set your own path:
-# log_dir_path = "/var/log"
+# log_dir = "/var/log"
```
mikechristie: What was the reason for the rename? I just saw the mention in the git commit and not a reason.

Normally I would not care. I like log_dir more, but because it is a user facing value it is a concern.

main.c
```diff
@@ -943,6 +943,27 @@ static int tcmu_set_max_fd_limit(const int nr_files)
...
+	if (cfg == NULL) {
+		tcmu_err("allocating TCMU config failed: %m\n");
+		errno = ENOMEM;
+		return NULL;
+	}
+
+	log_dir = getenv("TCMU_LOGDIR");
+	snprintf(cfg->def_log_dir, PATH_MAX, log_dir ? log_dir : TCMU_LOG_DIR_DEFAULT);
+	cfg->def_log_level = TCMU_CONF_LOG_INFO;
+	snprintf(cfg->path, PATH_MAX, TCMU_CONFIGFILE_DEFAULT);
```
mikechristie: I think we want to rename this to tcmu_initialize_config() and put it in libtcmu_config.c. libtcmu_config.c will then manage both the allocation and freeing of the tcmu_config object.

Also in main() with your changes we can fix the ordering so the cleanup/freeing path is the reverse of the setup/allocation path right?

main()

tcmu_initialize_config()
tcmu_load_config()
tcmu_setup_log()
tcmu_watch_config()

....

tcmu_unwatch_config()
tcmu_destroy_log()
tcmu_free_config()

so then that matches the error unwinding path too.

libtcmu_log.h
```diff
@@ -22,7 +22,8 @@
 #define TCMU_LOG_DEBUG_SCSI_CMD	(LOG_DEBUG + 1)	/* scsi cmd debug-level messages */
 
 /* default tcmu log dir path */
-#define TCMU_LOG_DIR_DEFAULT   "/var/log"
+#define TCMU_LOG_DIR_DEFAULT      "/var/log"
+#define TCMU_CONFIGFILE_DEFAULT   "/etc/tcmu/tcmu.conf"
```
mikechristie: Use tabs instead of spaces.

libtcmu_log.c
```diff
@@ -699,10 +697,31 @@ int tcmu_setup_log(char *log_dir)
...
+	}
+
+	return st1.st_dev == st2.st_dev && st1.st_ino == st2.st_ino;
+}
+
 int tcmu_resetup_log_file(char *log_dir)
 {
 	int ret;
 
+	if (is_path_same(tcmu_log_dir, log_dir)) {
```
mikechristie: Can you change this to is_same_path()

libtcmu_log.c
```diff
@@ -395,7 +400,7 @@ static int output_to_fd(int pri, const char *timestamp,
 	char *buf, *msg;
 	int count, ret, written = 0, r, pid = 0;
 
-	if (fd < 0)
+	if (fd ==  -1)
```
mikechristie: Drop the extra space between "==" and -1.

pkalever: Done

libtcmu_log.c
pkalever: Done

libtcmu_log.c
```diff
@@ -699,10 +697,31 @@ int tcmu_setup_log(char *log_dir)
...
 
+static bool is_path_same(const char* path1, const char* path2)
+{
+	struct stat st1 = {0,};
+	struct stat st2 = {0,};
+
+	if (!path1 || !path2)
+		return false;
+
+	if ((stat(path1, &st1) == -1) || (stat(path2, &st2) == -1)) {
```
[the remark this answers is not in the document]
pkalever: Sure :-)

libtcmu_log.h
pkalever: Sure

main.c
pkalever: Makesense done, I will update the series with a new patch about organizing the cleanup order

main.c
```diff
@@ -981,10 +1002,13 @@ int main(int argc, char **argv)
...
 	bool watching_cfg = false;
 	struct flock lock_fd = {0, };
-	char *log_dir = NULL;
 	int fd;
 	int ret;
 
+	if ((tcmu_cfg = tcmu_initiatize_options()) == NULL) {
+		tcmu_err("initializing the tcmu config failed: %m\n");
+		exit(EXIT_FAILURE);
+	}
```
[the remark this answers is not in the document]
pkalever: Done

tcmu.conf
pkalever: Frankly, the naming looks a bit awkward to me.

We need to support replacing the tcmu.conf (just in case if new variables are introduced or defaults are changed it will be helpful. *Note* there was a recent change to tcmu.conf about introducing CRIT loglevel already)

As we anyway need to support replacing this file, It is fine to tune the option name here, as the new option name is presented to user anyway. We still save one old tcmu.conf as backup after replace (we can increase to hold more later)

pkalever:
V9: Address all the review comments from Mike Christie above.
Note: new patch addition to the series "main: organize the cleanup sequence "

pkalever:
@mikechristie @lxbsz updated with all the comments till date. Please take a look.

Mike, thanks for the review comments.

lxbsz:
Test done, works well.

libtcmu_log.h
```diff
@@ -22,7 +22,8 @@
 #define TCMU_LOG_DEBUG_SCSI_CMD	(LOG_DEBUG + 1)	/* scsi cmd debug-level messages */
 
 /* default tcmu log dir path */
-#define TCMU_LOG_DIR_DEFAULT   "/var/log"
+#define TCMU_LOG_DIR_DEFAULT	"/var/log"
+#define TCMU_CONFIGFILE_DEFAULT	"/etc/tcmu/tcmu.conf"
```
mikechristie: I think this should have gone to libtcmu_config.h.

I will just fix this up when I merge the set.

mikechristie: I cleaned this up here:

https://github.com/open-iscsi/tcmu-runner/pull/509

PR #507, opened 2018-12-03
cmake: drop rsyslog build time check

ddiss:
rsyslog is a *runtime* dependency for the tcmu-runner systemd service.
It doesn't make sense to check for it at build time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,11 +32,6 @@ find_package(PkgConfig)
 pkg_check_modules(GLIB REQUIRED gio-unix-2.0)
 pkg_check_modules(KMOD REQUIRED libkmod)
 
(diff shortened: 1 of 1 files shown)


pkalever:
BTW: @ddiss this was intentional check. Any reason why you want to mop-off it ?

If we don't mandate it while build time, we never gonna take advantage of it.

If you have a better way, please propose.

ddiss:
I'm still completely confused, sorry...
The spec file includes "Requires... rsyslog" which means that rpm will check for and enforce the rsyslog dependency when tcmu-runner is installed.
Unless I'm mistaken, the *build* of tcmu-runner does not rely on anything from rsyslog, so the cmake check is completely unnecessary.

PR #498, opened 2018-11-18
Start to add back libtcmu

mikechristie:
This is based on @baihuahua 's patches at:

https://github.com/open-iscsi/tcmu-runner/pull/474

It starts to add back the stable libtcmu.

The last major TODO items I have:

1. Decide if tcmulib_cmd should work like how tcmu_device does where it is defined in libtcmu_priv.h and the lib user interacts with it via a helper functions instead of directly accessing it.

2. Remove tcmulib_cmd->cmdstate and tcmulib_cmd->done. Add a void *hm_private field that handlers and/or daemons can add their private data on. For runner we would add a struct with ->cmdstate and ->done and other related info on it used in tcmur_cmd_handler.c and by the runner handlers.

PR #515, opened 2018-12-12
Further tune libtcmu

baihuahua:
This is one of the last TODO items from #498 which replaces cmdstate and done in tcmulib_cmd struct with a hm_private pointer. This patchset doesn't touch the access to tcmulib_cmd with helper functions part.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -37,6 +37,10 @@ void tcmur_command_complete(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 			    int rc)
 {
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
+	struct tcmur_cmd_state *rcmd_state = tcmu_cmd_get_private(cmd);
+
+	if (rcmd_state)
+		free(rcmd_state);
 
 	pthread_cleanup_push(_cleanup_spin_lock, (void *)&rdev->lock);
 	pthread_spin_lock(&rdev->lock);
@@ -188,6 +192,7 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
 					     int *return_err)
 {
 	struct unmap_state *state;
+	struct tcmur_cmd_state *rcmd_state = tcmu_cmd_get_private(cmd);
 	int ret;
 
 	*return_err = 0;
@@ -208,7 +213,7 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
 
 	state->refcount = 0;
 	state->error = false;
-	cmd->cmdstate = state;
+	rcmd_state->cmdstate = state;
 	return state;
 
 out_free_state:
@@ -225,13 +230,16 @@ static void unmap_state_free(struct unmap_state *state)
 static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd,
 			     int ret)
 {
-	struct unmap_descriptor *desc = ucmd->cmdstate;
+	struct tcmur_cmd_state *urcmd_state = tcmu_cmd_get_private(ucmd);
+	struct unmap_descriptor *desc = urcmd_state->cmdstate;
 	struct tcmulib_cmd *origcmd = desc->origcmd;
-	struct unmap_state *state = origcmd->cmdstate;
+	struct tcmur_cmd_state *orig_rcmd_state = tcmu_cmd_get_private(origcmd);
+	struct unmap_state *state = orig_rcmd_state->cmdstate;
 	bool error;
 	int status;
 
 	free(desc);
+	free(urcmd_state);
 
 	pthread_mutex_lock(&state->lock);
 	error = state->error;
@@ -261,10 +269,11 @@ static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd,
 static int unmap_work_fn(struct tcmu_device *dev, struct tcmulib_cmd *ucmd)
 {
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
-	struct unmap_descriptor *desc = ucmd->cmdstate;
+	struct tcmur_cmd_state *urcmd_state = tcmu_cmd_get_private(ucmd);
+	struct unmap_descriptor *desc = urcmd_state->cmdstate;
 	uint64_t offset = desc->offset, length = desc->length;
 
-	ucmd->done = handle_unmap_cbk;
+	urcmd_state->done = handle_unmap_cbk;
 
 	return rhandler->unmap(dev, ucmd, offset, length);
 }
@@ -273,13 +282,15 @@ static int align_and_split_unmap(struct tcmu_device *dev,
 				 struct tcmulib_cmd *origcmd,
 				 uint64_t lba, uint64_t nlbas)
 {
-	struct unmap_state *state = origcmd->cmdstate;
+	struct tcmur_cmd_state *orig_rcmd_state = tcmu_cmd_get_private(origcmd);
+	struct unmap_state *state = orig_rcmd_state->cmdstate;
 	uint32_t block_size = tcmu_dev_get_block_size(dev);
 	uint64_t opt_unmap_gran;
 	uint64_t unmap_gran_align, mask;
 	int ret = TCMU_STS_NOT_HANDLED;
 	int j = 0;
 	struct unmap_descriptor *desc;
+	struct tcmur_cmd_state *urcmd_state;
 	struct tcmulib_cmd *ucmd;
 	uint64_t lbas;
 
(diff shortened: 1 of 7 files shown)


tcmur_cmd_handler.c
```diff
@@ -323,17 +334,24 @@ static int align_and_split_unmap(struct tcmu_device *dev,
...
 			ret = TCMU_STS_NO_RESOURCE;
-			goto free_desc;
+			goto free_urcmd;
 		}
 
 		desc->origcmd = origcmd;
 		desc->offset = lba * block_size;
 		desc->length = lbas * block_size;
-		ucmd->cmdstate = desc;
+		urcmd_state->cmdstate = desc;
```
lxbsz: Here you didn't set the urcmd_state to ucmd.

tcmur_cmd_handler.c
```diff
@@ -816,12 +842,18 @@ static int write_verify_init(struct tcmulib_cmd *origcmd, size_t length)
...
+
+	read_rcmd_state = calloc(1, sizeof(*read_rcmd_state));
+	if (!read_rcmd_state)
+		goto out;
 	readcmd = calloc(1, sizeof(*readcmd));
 	if (!readcmd)
-		goto out;
-	readcmd->cmdstate = origcmd;
+		goto free_rcmd;
+	read_rcmd_state->cmdstate = origcmd;
```
lxbsz: It seems the same here forgot to set read_rcmd_state to readcmd.

tcmur_cmd_handler.c
```diff
@@ -1647,24 +1698,29 @@ caw_init_readcmd(struct tcmulib_cmd *origcmd, size_t length)
 	state->requested = length;
 	state->origcmd = origcmd;
 
-	readcmd->cmdstate = state;
+	tcmu_cmd_set_private(readcmd, read_rcmd_state);
+	read_rcmd_state->cmdstate = state;
```
lxbsz: Maybe this line should move ahead of tcmu_cmd_set_private().

tcmur_cmd_handler.c
```diff
@@ -1829,7 +1891,14 @@ int tcmur_handle_caw(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 	if (ret)
 		return ret;
 
-	cmd->cmdstate = caw_fn;
+        rcmd_state = calloc(1, sizeof(*rcmd_state));
+        if (!rcmd_state) {
+                tcmu_dev_err(dev, "Failed to calloc memory for tcmur_cmd_state!\n");
+                return TCMU_STS_NO_RESOURCE;
+        }
+
```
lxbsz: Please replace the spaces with tab for all the above lines.

tcmur_cmd_handler.c
```diff
@@ -1829,7 +1891,14 @@ int tcmur_handle_caw(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
...
+	tcmu_cmd_set_private(cmd, rcmd_state);
+	rcmd_state->cmdstate = caw_fn;
```
lxbsz: The same here maybe this line should move ahead of tcmu_cmd_set_private().

tcmur_cmd_handler.c
```diff
@@ -2220,7 +2310,18 @@ int tcmur_cmd_passthrough_handler(struct tcmu_device *dev,
...
 	 */
 	track_aio_request_start(rdev);
+
+        rcmd_state = calloc(1, sizeof(*rcmd_state));
+        if (!rcmd_state) {
+                tcmu_dev_err(dev, "Failed to calloc memory for tcmur_cmd_state!\n");
+                ret = TCMU_STS_NO_RESOURCE;
+                goto untrack;
+        }
+
```
lxbsz: The same here, please replace the spaces with tab for all the line above.

tcmur_cmd_handler.c
```diff
@@ -2269,6 +2371,15 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		break;
 	}
 
+	rcmd_state = calloc(1, sizeof(*rcmd_state));
+	if (!rcmd_state) {
+                tcmu_dev_err(dev, "Failed to calloc memory for tcmur_cmd_state!\n");
+                ret = TCMU_STS_NO_RESOURCE;
+                goto untrack;
+        }
+
```
lxbsz: The same here with the spaces issue.

tcmur_cmd_handler.c
```diff
@@ -37,6 +37,10 @@ void tcmur_command_complete(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 			    int rc)
 {
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
+	struct tcmur_cmd_state *rcmd_state = tcmu_cmd_get_private(cmd);
+
+	if (rcmd_state)
+		free(rcmd_state);
```
lxbsz: Here is one problem.
The cmd is allocated by using malloc(), so the cmd->hm_private maybe any possible random value. And if the command is any of the sync type which falls in handle_sync_cmd() will it crash here ?

tcmur_cmd_handler.c
baihuahua: Good catch! Will fix that. Thx.

tcmur_cmd_handler.c
baihuahua: Yeah, thanks.

tcmur_cmd_handler.c
baihuahua: OK, i just followed the coding style in this file. :) Will fix this and the similar below.

tcmur_cmd_handler.c
baihuahua: Will fix the spaces issues above. Thanks.

tcmur_cmd_handler.c
baihuahua: Yeah, i think you're right. Could we just change malloc into calloc for tcmulib_cmd allocation to fix this?

lxbsz: Yeah, that's okay for me.

baihuahua:
Fixed issues above. Please review. @lxbsz @mikechristie

tcmur_cmd_handler.c
```diff
@@ -2269,6 +2373,15 @@ static int tcmur_cmd_handler(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	rcmd_state = calloc(1, sizeof(*rcmd_state));
+	if (!rcmd_state) {
+		tcmu_dev_err(dev, "Failed to calloc memory for tcmur_cmd_state!\n");
+ 		ret = TCMU_STS_NO_RESOURCE;
+		goto untrack;
+	}
+
+	tcmu_cmd_set_private(cmd, rcmd_state);
+
 	switch(cdb[0]) {
```
mikechristie: If we fail early in one of the handle_* functions where the cmd is never sent to the handler then we will leak the rcmd_state allocated above.

At the end of this function in the chunk:

```
        if (ret != TCMU_STS_ASYNC_HANDLED) {
                track_aio_request_finish(rdev, NULL);
                free(rcmd_state);
```

we also need to free rcmd_state.

tcmur_cmd_handler.c
```diff
@@ -2220,7 +2312,18 @@ int tcmur_cmd_passthrough_handler(struct tcmu_device *dev,
...
+		ret = TCMU_STS_NO_RESOURCE;
+		goto untrack;
+	}
+
+	tcmu_cmd_set_private(cmd, rcmd_state);
 	ret = handle_passthrough(dev, cmd);
+
+untrack:
 	if (ret != TCMU_STS_ASYNC_HANDLED)
 		track_aio_request_finish(rdev, NULL);
```
mikechristie: Need to add a

free(rcmd_state);

in this if chunk when handle_passthrough returned a failure.

tcmur_cmd_handler.c
```diff
@@ -1829,7 +1893,14 @@ int tcmur_handle_caw(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
...
+	rcmd_state = calloc(1, sizeof(*rcmd_state));
+	if (!rcmd_state) {
+		tcmu_dev_err(dev, "Failed to calloc memory for tcmur_cmd_state!\n");
+		return TCMU_STS_NO_RESOURCE;
+	}
+
+	rcmd_state->cmdstate = caw_fn;
+	tcmu_cmd_set_private(cmd, rcmd_state);
 
 	return async_handle_cmd(dev, cmd, tcmur_caw_fn);
```
mikechristie: If this fails to queue the cmd to the handler we will leak.

You need to do something like

```
ret = async_handle_cmd()
if (ret != TCMU_STS_ASYNC_HANDLED)
      free(rcmd_state)
```

tcmur_cmd_handler.c
```diff
@@ -784,6 +810,7 @@ int tcmur_handle_writesame(struct tcmu_device *dev, struct tcmulib_cmd *cmd,
 			   tcmur_writesame_fn_t write_same_fn)
 {
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	struct tcmur_cmd_state *rcmd_state = tcmu_cmd_get_private(cmd);
```
mikechristie: This is called from

handle_try_passthrough -> handle_cmd -> rbd/glfs -> tcmur_handle_writesame

and it looks like rcmd_state is not allocated yet. Below tcmur_handle_caw has the same call path, but you allocated in that function.

mikechristie:
Hey baihuahua,

Maybe to make the handler's cmd allocation easier we could do the following or something like it.

1. Add a new field, hm_cmd_size, to struct tcmulib_handler.
2. tcmulib_get_next_command willl allocate hm_cmd_size bytes with the tcmulib_cmd and then set the tcmulib_cmd->hm_private pointer.
3. tcmulib_command_complete will then free the hm_cmd_size bytes along with the tcmulib_cmd.

The handler/daemons then do not have to worry about it.

mikechristie:
One change to the above comment.

We can make it a little easier and just have the caller pass in the cmd_size to tcmulib_get_next_command. No need for tcmulib_handler changes.

baihuahua:
Great! This would be more easier for tcmulib_cmd allocation/free. I'll repush this PR.

mikechristie:
I ended up implementing what is needed here

Closing this PR.

PR #518, opened 2018-12-14
debug-log: add helpful hint msg about previous log_dir

pkalever:
Logs on new log file path look like:

2018-12-14 18:30:31.154 14398 [CRIT] create_file_output:480:
  log file path now is '/var/log/gluster-block/tcmu-runner.log'
2018-12-14 18:30:31.154 14398 [CRIT] tcmu_resetup_log_file:748:
  you many find previous logs at '/var/log/tcmu-runner.log'


libtcmu_log.c
```diff
@@ -714,13 +714,17 @@ static bool is_same_path(const char* path1, const char* path2)
 int tcmu_resetup_log_file(char *log_dir)
 {
 	int ret;
+	char old_logdir[PATH_MAX] = {'\0', };
```
mikechristie: I think tcmu_log_dir_set should have been called by the tcmu_log_dir_create caller and not tcmu_log_dir_create. tcmu_log_dir_create just creates the dirs. Calling it in tcmu_log_dir_create leads to the weird dance we do below and also if the resetup fails we were leaving it set to the bad value.

lxbsz: Yeah, agree with this.

Issue #516, opened 2018-12-13
move dbus into libtcmu


baihuahua:
When tcmu-runner is running with other tcmu daemon like qemu-tcmu at the same time on a host with genl version setting 2 in kernel, creating target through qemu-tcmu will fail with message in tcmu-runner's log file:

> 2018-12-12 13:48:27.039 1542 [DEBUG] handle_netlink:195: cmd 1. Got header version 2. Supported 2.
2018-12-12 13:48:27.039 1542 [ERROR] add_device:477: could not find handler for uio0

This's because tcmu-runner and qemu-tcmu both receive the device add event from netlink, while tcmu-runner can't find a handler to handle the device so it fails the creation by returning the kernel error code.

To settle this problem, tcmu-runner needs a way to give other tcmu daemon a chance to handle the device creation before failing it. Or we can lower down dbus from tcmu-runner into libtcmu so all daemons can acquire dbus on their own without dependence.

mikechristie:
It's not clear to me how dbus comes into the problem or how moving it helps? If you moved it to libtcmu would you add a new AddDevice dbus operation, like CheckConfig, which is used by apps like targetcli to talk to specific daemons?

I think if we continue to use netlink we want to fix the tools/daemons/kernels so netlink events are sent to specific daemons or at least add a field in the netlink operation that indicates which daemon the command is for.

mikechristie:
Is this a similar problem as with containers? We could have tcmu-runner or daemon123 running in multiple containers. We only want a specific instance to handle the netlink request. Currently, because we are just multicasting to all listeners.

baihuahua:
> It's not clear to me how dbus comes into the problem or how moving it helps? 

I misunderstood them, please forget about the dbus moving thing to fix this issue.

> 
> I think if we continue to use netlink we want to fix the tools/daemons/kernels so netlink events are sent to specific daemons or at least add a field in the netlink operation that indicates which daemon the command is for.

Yeah, i agree about this. Which one do you think is more easier to implement, changing multicasting or adding a field in netlink operation?

baihuahua:
> Is this a similar problem as with containers? We could have tcmu-runner or daemon123 running in multiple containers. We only want a specific instance to handle the netlink request. Currently, because we are just multicasting to all listeners.

[...]

PR #519, opened 2018-12-25
libtcmu: allow multiple daemons co-exist

baihuahua:
As we support libtcmu again users could want to run several
daemons at the same time. So don't fail the device addition
via netlink reply when can't find a handler in ctx to handle
the device. This just means the device addition event is not
for **THIS** ctx so give other daemons a chance to try.

This fix is for issue #516 .

Signed-off-by: Yaowei Bai <baiyaowei@cmss.chinamobile.com>

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -212,6 +212,8 @@ static int handle_netlink(struct nl_cache_ops *unused, struct genl_cmd *cmd,
 		ret = add_device(ctx, buf,
 				 nla_get_string(info->attrs[TCMU_ATTR_DEVICE]),
 				 false);
+		if (ret == -ENOENT)
+			return 0;
 		break;
 	case TCMU_CMD_REMOVED_DEVICE:
 		reply_cmd = TCMU_CMD_REMOVED_DEVICE_DONE;
@@ -432,8 +434,8 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 
 	dev->handler = find_handler(ctx, dev->cfgstring);
 	if (!dev->handler) {
-		tcmu_err("could not find handler for %s\n", dev->dev_name);
-		goto err_free;
+		tcmu_warn("could not find handler for %s\n", dev->dev_name);
+		goto err_nohandler;
 	}
 
 	if (dev->handler->check_config &&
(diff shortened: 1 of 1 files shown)


libtcmu.c
```diff
@@ -432,8 +434,8 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 
 	dev->handler = find_handler(ctx, dev->cfgstring);
 	if (!dev->handler) {
-		tcmu_err("could not find handler for %s\n", dev->dev_name);
-		goto err_free;
+		tcmu_dbg("could not find handler for %s\n", dev->dev_name);
```
pkalever: should this be a warning tcmu_warn () ?
For tcmu-runner case, it is helpful to have this msg at logs with non-debug log levels too.

libtcmu.c
```diff
@@ -536,8 +538,11 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 		tcmu_cfgfs_dev_exec_action(dev, "block_dev", 0);
 err_free:
 	free(dev);
-
 	return -ENOENT;
+err_nohandler:
+	free(dev);
+
+	return 1;
```
pkalever: In general, It makes more sense to return -1/1 for all those str***() function errors, and return ENOENT for nohandler case.

libtcmu.c
baihuahua: Yeah, that makes sense to me, thanks.

baihuahua:
Updated, please review. Thanks. @pkalever @mikechristie

baihuahua:
Hey Mike, what do you think about this? @mikechristie

mikechristie:
I am on PTO from Jan 28 - Feb 8. I am going to try and get to this next week while on vacation.

bgly:
This looks fine to me.

baihuahua:
Thanks @bgly .Mike, what do you think about this? @mikechristie

bno1:
targetcli hangs if the added device is not found by any daemon. Restarting tcmu-runner solves the problem because it resets the netlink interface, but this is not a good solution. Also consider that a custom daemon might crash or be killed during the handling of TCMU_CMD_ADDED_DEVICE, which would also cause targetcli to hang.

PR #522, opened 2019-01-04
libtcmu: setup_netlink maybe failure but can't sensed

pengliang233:

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -22,6 +22,7 @@
 #include <libnl3/netlink/genl/genl.h>
 #include <libnl3/netlink/genl/mngt.h>
 #include <libnl3/netlink/genl/ctrl.h>
+#include <libnl3/netlink/errno.h>
 
 #include "target_core_user_local.h"
 #include "libtcmu.h"
@@ -240,11 +241,13 @@ static int set_genl_features(struct nl_sock *sock)
 {
 	struct nl_msg *msg;
 	void *hdr;
-	int ret = -ENOMEM;
+	int ret = -NLE_NOMEM;
 
 	msg = nlmsg_alloc();
-	if (!msg)
+	if (!msg) {
+		tcmu_err("Could not allocate a new message.\n");
 		return ret;
+	}
 
 	hdr = genlmsg_put(msg, NL_AUTO_PORT, NL_AUTO_SEQ, tcmu_ops.o_id,
 			  0, NLM_F_ACK, TCMU_CMD_SET_FEATURES, TCMU_NL_VERSION);
@@ -255,7 +258,15 @@ static int set_genl_features(struct nl_sock *sock)
 	if (ret < 0)
 		goto free_msg;
 
+	/*
+	 * Could be a older kernel. Ignore failure and just work in degraded
+	 * mode.
+	 */
 	ret = nl_send_sync(sock, msg);
+	if (ret == -NLE_OPNOTSUPP) {
+		tcmu_warn("Kernel does not support the operation.\n");
+		ret = 0;
+	}
 	goto done;
 
 free_msg:
@@ -313,11 +324,9 @@ static struct nl_sock *setup_netlink(struct tcmulib_context *ctx)
 		goto err_unregister;
 	}
 
-	/*
-	 * Could be a older kernel. Ignore failure and just work in degraded
-	 * mode.
-	 */
-	set_genl_features(sock);
+	ret = set_genl_features(sock);
+	if (ret < 0)
+		goto err_unregister;
 
 	return sock;
 


libtcmu.c
```diff
@@ -256,6 +256,8 @@ static int set_genl_features(struct nl_sock *sock)
 		goto free_msg;
 
 	ret = nl_send_sync(sock, msg);
+	if (ret < 0)
+		goto free_msg;
```
pkalever: It would be nice to add a log msg about this.

libtcmu.c
```diff
@@ -317,7 +319,9 @@ static struct nl_sock *setup_netlink(struct tcmulib_context *ctx)
 	 * Could be a older kernel. Ignore failure and just work in degraded
 	 * mode.
 	 */
-	set_genl_features(sock);
+	ret = set_genl_features(sock);
+	if (ret < 0)
```
pkalever: same here.

libtcmu.c
pengliang233: ok

libtcmu.c
pengliang233: It has record "tcmu_err("Could not set features. Error %d\n", ret);" before set_genl_features return.

libtcmu.c
```diff
@@ -317,7 +321,9 @@ static struct nl_sock *setup_netlink(struct tcmulib_context *ctx)
...
+		goto err_unregister;
```
mikechristie: We shouldn't fail here, because if we are running on a older kernel runner will fail to start which is too extreme for this optional feature. If that function fails then we just cannot use newer features.

Maybe to make it more user friendly we could just log a more verbose message. Replace the message in set_genl_features with:

"Could not setup kernel features. An update kernel is needed. Will run in degraded mode."

libtcmu.c
```diff
@@ -256,6 +256,10 @@ static int set_genl_features(struct nl_sock *sock)
 		goto free_msg;
 
 	ret = nl_send_sync(sock, msg);
+	if (ret < 0) {
+		tcmu_err("Could not send the msg. Error %d\n", ret);
```
mikechristie: No need for this. It is already logged on the done part.

libtcmu.c
mikechristie: The goto free_msg is wrong here. nl_send_sync will always free the msg.

So I guess I am saying those whole chunk is not needed.

pengliang233: This function frees the msg object after transmitting it by calling nlmsg_free(). It means whether the return is 0 on success or a negative error code, it will calling nlmsg_free to free?

libtcmu.c
pengliang233: Do you means the older kernel cannot support libnl-genl? or other reason? 
If we can confirm the older kernel  just log a more verbose message. If it's not running on older kernel maybe it's better to goto err_unregister.

libtcmu.c
mikechristie: https://github.com/thom311/libnl/blob/08160f0555f988db1ec3de65fa188a9d48ebff9d/lib/nl.c#L548

nl_send_sync does:

```
int nl_send_sync(struct nl_sock *sk, struct nl_msg *msg)
{
	int err;

	err = nl_send_auto(sk, msg);
	nlmsg_free(msg);
```

So as long as the function is called nlmsg_free is going to be called. It does not matter if nl_send_sync returns success or failure.

This is why in our code we have that funky goto done which bypasses the nlmsg_free call in our code that is always called after calling nl_send_sync():

```
ret = nl_send_sync(sock, msg);
goto done;
```

pengliang233: Thanks, I've seen the source code in libnl.

libtcmu.c
pengliang233: Thanks for your advice. I made a test, the nl_send_sync will return -10 if the kernel not support the TCMU_CMD_SET_FEATURES. 
The libnl transforms the errno to it's own error code.
```
int nl_sendmsg(struct nl_sock *sk, struct nl_msg *msg, struct msghdr *hdr)
{
	ret = sendmsg(sk->s_fd, hdr, 0);
	if (ret < 0)
		return -nl_syserr2nlerr(errno);
}
```
```
int nl_syserr2nlerr(int error)
{
	switch (error) {
	case EOPNOTSUPP:	return NLE_OPNOTSUPP;
        }
}
```

`#define NLE_OPNOTSUPP		10`

In order to catch the error code, it need to define a error code in tcmu-runner.

mikechristie: Just to make sure we are on the same page. In libtcmu.c you are just going to include the nl errno.h header like this right:

```
#include <libnl3/netlink/errno.h>

Then in set_genl_features() you will do something like

ret = nl_send_sync()
if (ret == -NLE_OPNOTSUPP) {
```

One more comment. Depending on how you fix this you might need to also fix up the nla_put_u8 ret handling because that also returns a -NLE, but in some places set_genl_features returns -ENOMEM, and we will want to return the same type of code (either a -NLE or -errno) and not a fix of the two.

pengliang233: Thanks again. I've thought about this too. If we should return the same type of code, it's better to init the 'ret' to NLE_NOMEM.

mikechristie:
Thanks!

PR #527, opened 2019-01-28
runner:fix the condition enter to tcmu_acquire_dev_lock function

lmh10144360:

### DIFF
--- a/tcmur_device.c
+++ b/tcmur_device.c
@@ -414,7 +414,7 @@ retry:
 	}
 
 drop_conn:
-	if (ret == TCMU_STS_TIMEOUT) {
+	if (ret == TCMU_STS_TIMEOUT || ret == TCMU_STS_FENCED) {
 		tcmu_dev_dbg(dev, "Fail handler device connection.\n");
 		tcmu_notify_conn_lost(dev);
 	}


mikechristie:
You are going to have to add a more detailed description for why we need this patch. I removed that code here

https://github.com/open-iscsi/tcmu-runner/commit/08e3a0e58589d6355642128aecf377d0ef3390da

because during device initialization/discovery probe IO could be sent through all paths. This leads to each path getting fenced/unfenced multiple times and if we escalate to dropping the iscsi session each time that happens those probe IOs could end up failing and the paths not getting added to the initiator side multipath device.

lmh10144360:
@mikechristie Thanks for reply. In some cases, the tcmu_acquire_dev_lock function comes into the tcmu_reopen_dev function, and if image close success, and image open fail. In this case, we change the ret to TCMU_STS_FENCED,and will not come into tcmu_notify_conn_lost function. After soon tcmu-runner comes in an inquiry cmd, when inquiry cmd processed to tcmu_update_dev_lock_state function，
because of the image was not open successed,then the  parameter state->image transfered to librbd is NULL, and this will cause the tcmu-runner core.
Based on what you mentioned above, whether we can change the ret to TCMU_STS_TIMEOUT if  tcmu_reopen_dev not return success?or any other suggestion？ Thanks again.

mikechristie:
Thanks for the info. I think the crash from the null state->image should be fixed with the PR:

https://github.com/open-iscsi/tcmu-runner/commit/c0887e476e324fde5e0d23e434c9664b173bfee9

mikechristie:
Oh yeah, this specific patch should fix the issue with the successful close() but failed open():

https://github.com/open-iscsi/tcmu-runner/pull/525/commits/a320ef8b174d21085221044dcf1e0ae70fa18b1a

lmh10144360:
@mikechristie  Thanks you. I got it!

PR #528, opened 2019-01-28
Drop sense arg from tcmu_explicit_transition

lmh10144360:


mikechristie:
Thanks.

PR #529, opened 2019-01-31
Fix pthread error ret in aio handler

pengliang233:
The pthread functions return a positive errno, but we assumed it was negative in the caller dev_added.


mikechristie:
Thanks.

PR #532, opened 2019-02-09
Fix bugs in unmap

bgly:
First bug is in state->refcount where it was incremented after
the call to async_handle_cmd. This can cause issues if
handle_unmap_cbk gets called first before the return to increment
the value of state->refcount.

Second bug is not returning the right error code when doing a
unmap_state_alloc.

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -172,7 +172,6 @@ static int write_work_fn(struct tcmu_device *dev,
 struct unmap_state {
 	pthread_mutex_t lock;
 	unsigned int refcount;
-	bool error;
 	int status;
 };
 
@@ -206,8 +205,9 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
 		goto out_free_state;
 	}
 
-	state->refcount = 0;
-	state->error = false;
+	/* released by allocator when done submitting unmaps */
+	state->refcount = 1;
+	state->status = TCMU_STS_OK;
 	cmd->cmdstate = state;
 	return state;
 
@@ -216,10 +216,28 @@ out_free_state:
 	return NULL;
 }
 
-static void unmap_state_free(struct unmap_state *state)
+static void unmap_state_put(struct tcmu_device *dev,
+			    struct tcmulib_cmd *origcmd, int ret)
 {
+	struct unmap_state *state = origcmd->cmdstate;
+	int status;
+
+	pthread_mutex_lock(&state->lock);
+
+	if (state->status != TCMU_STS_OK && ret)
+		state->status = ret;
+
+	if (--state->refcount > 0) {
+		pthread_mutex_unlock(&state->lock);
+		return;
+	}
+	status = state->status;
+	pthread_mutex_unlock(&state->lock);
+
 	pthread_mutex_destroy(&state->lock);
 	free(state);
+
+	aio_command_finish(dev, origcmd, status);
 }
 
 static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd,
@@ -227,35 +245,11 @@ static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd,
 {
 	struct unmap_descriptor *desc = ucmd->cmdstate;
 	struct tcmulib_cmd *origcmd = desc->origcmd;
-	struct unmap_state *state = origcmd->cmdstate;
-	bool error;
-	int status;
 
 	free(desc);
-
-	pthread_mutex_lock(&state->lock);
-	error = state->error;
-	/*
-	 * Make sure to only copy the first error
-	 */
-	if (!error && ret) {
-		state->error = true;
-		state->status = ret;
-	}
-
 	free(ucmd);
 
-	if (--state->refcount > 0) {
-		pthread_mutex_unlock(&state->lock);
-		return;
-	}
-	status = state->status;
-	error = state->error;
-	pthread_mutex_unlock(&state->lock);
-
-	unmap_state_free(state);
-
-	aio_command_finish(dev, origcmd, error ? status : ret);
+	unmap_state_put(dev, origcmd, ret);
 }
 
 static int unmap_work_fn(struct tcmu_device *dev, struct tcmulib_cmd *ucmd)
@@ -347,6 +341,10 @@ static int align_and_split_unmap(struct tcmu_device *dev,
 			tcmu_dev_dbg(dev, "There are totally %d splits\n", j);
 		}
 
+		pthread_mutex_lock(&state->lock);
+		state->refcount++;
+		pthread_mutex_unlock(&state->lock);
+	
 		ret = async_handle_cmd(dev, ucmd, unmap_work_fn);
 		if (ret != TCMU_STS_ASYNC_HANDLED)
 			goto free_ucmd;
@@ -356,12 +354,14 @@ static int align_and_split_unmap(struct tcmu_device *dev,
 
 		lbas = min(opt_unmap_gran, nlbas);
 
-		state->refcount++;
 	}
 
 	return ret;
 
 free_ucmd:
+	pthread_mutex_lock(&state->lock);
+	state->refcount--;
+	pthread_mutex_unlock(&state->lock);
 	free(ucmd);
 free_desc:
 	free(desc);
@@ -371,14 +371,16 @@ free_desc:
 static int handle_unmap_internal(struct tcmu_device *dev, struct tcmulib_cmd *origcmd,
 				 uint16_t bddl, uint8_t *par)
 {
-	struct unmap_state *state = origcmd->cmdstate;
+	struct unmap_state *state;
 	uint16_t offset = 0;
-	int ret = TCMU_STS_OK, i = 0, refcount;
+	int ret = TCMU_STS_OK, i = 0;
+
+	state = unmap_state_alloc(dev, origcmd, &ret);
+	if (!state)
+		return ret;
 
 	/* The first descriptor list offset is 8 in Data-Out buffer */
 	par += 8;
-
-	pthread_mutex_lock(&state->lock);
 	while (bddl) {
 		uint64_t lba;
 		uint64_t nlbas;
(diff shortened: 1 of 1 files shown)


bgly:
@mikechristie

tcmur_cmd_handler.c
```diff
@@ -423,9 +424,10 @@ static int handle_unmap_internal(struct tcmu_device *dev, struct tcmulib_cmd *or
 		state->error = true;
 
 	refcount = state->refcount;
+
 	pthread_mutex_unlock(&state->lock);
 
-	if (refcount)
+	if (refcount > 0)
```
mikechristie: You do not need this do you? If refcount has gone negative then we have a bigger problem.

tcmur_cmd_handler.c
```diff
@@ -538,8 +540,11 @@ static int handle_unmap(struct tcmu_device *dev, struct tcmulib_cmd *origcmd)
 	}
 
 	state = unmap_state_alloc(dev, origcmd, &ret);
-	if (!state)
+	if (!state) {
+		tcmu_dev_dbg(dev,"unmap no resources to allocate struct\n");
+		ret = TCMU_STS_NO_RESOURCE;
```
mikechristie: This should be set by unmap_state_alloc on failure. If not then that function needs to be fixed.

tcmur_cmd_handler.c
mikechristie: Ignore that comment. I see the bug. The refcount code is a little crazy. I am back from PTO Monday, so let me try to send a patch that describes the direction for the fix and you can take it from there. It is a little hard to just describe it because as you saw it is messy in there.

mikechristie:
Hey @bgly,

Attached is a completely untested patch. The idea is that unmap_state_put should always do the right thing when refcount=0. The submitting code paths then do not have to do those refcount checks. They just return TCMU_STS_ASYNC_HANDLED and unmap_state_put will complete the command when all refs are released.

[runner-fix-unmap-ref.txt](https://github.com/open-iscsi/tcmu-runner/files/2849418/runner-fix-unmap-ref.txt)

tcmur_cmd_handler.c
bgly: > Hey @bgly,
> 
> Attached is a completely untested patch. The idea is that unmap_state_put should always do the right thing when refcount=0. The submitting code paths then do not have to do those refcount checks. They just return TCMU_STS_ASYNC_HANDLED and unmap_state_put will complete the command when all refs are released.
> 
> [runner-fix-unmap-ref.txt](https://github.com/open-iscsi/tcmu-runner/files/2849418/runner-fix-unmap-ref.txt)

I will test now, ill let you know soon.

bgly:
@mikechristie The patch does not work; do you want me to debug it and fix it based off of your patch?

bgly:
@mikechristie I based the new patch off of the one you posted and made some fixes to it.

tcmur_cmd_handler.c
```diff
@@ -216,35 +218,22 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
...
-	free(desc);
-
 	pthread_mutex_lock(&state->lock);
 	error = state->error;
 	/*
 	 * Make sure to only copy the first error
 	 */
 	if (!error && ret) {
 		state->error = true;
-		state->status = ret;
```
mikechristie: I do not think this will work.

If you have sent 2 unmap calls to handle the command, and if the first one fails then we will set state->error = true, but not set state->status. When the second call completes, it might succeed. Refcount will now be zero and aio_command_finish will see

error = true
status = for  TCMU_STS_ASYNC_HANDLED if we are called from handle_unmap_in_writesame and TCMU_STS_OK if called from handle_unmap->handle_unmap_internal.
ret = TCMU_STS_OK

so because error=true from the first call, TCMU_STS_ASYNC_HANDLED or TCMU_STS_OK will be returned by aio_command_finish.

tcmur_cmd_handler.c
```diff
@@ -206,8 +206,10 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
 		goto out_free_state;
 	}
 
-	state->refcount = 0;
+	/* released by allocator when done submitting unmaps */
+	state->refcount = 1;
 	state->error = false;
+	state->status = TCMU_STS_ASYNC_HANDLED;
```
mikechristie: I think you can change that to TCMU_STS_OK to match state->error. Then if we get an error, unmap_state_put will set error and status.

tcmur_cmd_handler.c
```diff
@@ -253,11 +242,24 @@ static void handle_unmap_cbk(struct tcmu_device *dev, struct tcmulib_cmd *ucmd,
 	error = state->error;
 	pthread_mutex_unlock(&state->lock);
 
-	unmap_state_free(state);
+	pthread_mutex_destroy(&state->lock);
+	free(state);
 
 	aio_command_finish(dev, origcmd, error ? status : ret);
```
mikechristie: I think if you make the 2 other changes then state->status will always hold the correct value and we can just do

aio_command_finish(dev, origcmd, status);

tcmur_cmd_handler.c
bgly: I agree

tcmur_cmd_handler.c
bgly: I see that now, thanks!

bgly:
@mikechristie I have made the changes and also tested.

tcmur_cmd_handler.c
```diff
@@ -216,23 +218,13 @@ static struct unmap_state *unmap_state_alloc(struct tcmu_device *dev,
...
-	struct unmap_descriptor *desc = ucmd->cmdstate;
-	struct tcmulib_cmd *origcmd = desc->origcmd;
 	struct unmap_state *state = origcmd->cmdstate;
 	bool error;
 	int status;
 
-	free(desc);
-
 	pthread_mutex_lock(&state->lock);
 	error = state->error;
```
mikechristie: With your fixes/cleanups in the patch, we can just drop state->error now too right?

Here we can do

```
if (state->status != TCMU_STS_OK && ret)
      state->status = ret;
```

tcmur_cmd_handler.c
```diff
@@ -412,33 +415,19 @@ static int handle_unmap_internal(struct tcmu_device *dev, struct tcmulib_cmd *or
...
-
-	if (ret != TCMU_STS_ASYNC_HANDLED)
-		state->error = true;
-
-	refcount = state->refcount;
+	* unmap_state_put will do the right thing, so always return
+	* TCMU_STS_ASYNC_HANDLED
+	*/
+	if (ret == TCMU_STS_ASYNC_HANDLED) {
+		state->status = TCMU_STS_OK;
```
mikechristie: You could drop this line since you initialized it to that value.

tcmur_cmd_handler.c
```diff
@@ -412,33 +415,19 @@ static int handle_unmap_internal(struct tcmu_device *dev, struct tcmulib_cmd *or
...
+		ret = TCMU_STS_OK;
+	} else {
+		state->status = ret;
+	}
```
mikechristie: You need to put the same if/else block in handle_unmap_in_writesame so it does not pass in TCMU_STS_ASYNC_HANDLED to unmap_state_put

tcmur_cmd_handler.c
bgly: Yes, you can drop all that its not needed anymore after the correct status change. - Sorry I didn't clean it up before, I'm traveling atm. So i'm kinda rushed.

bgly:
@mikechristie I was testing with nr_threads set and unset which puts the code in async/sync cases and I made some fixes to the locking and unlocking aspects of the code in the latest push. I made sure we didnt deadlock. When nr_threads was not set in the handler it seemed like the ordering of calls was: handle_unmap_internal -> align_and_split_unmap -> async_handle_cmd -> unmap_work_fn -> handler_unmap -> handle_unmap_cbk -> unmap_state_put . This case unmap_state_put would get deadlocked from the initial state lock held at handle_unmap_internal. Thus I made some changes to address that issue by moving the locks around and the refcount.

mikechristie:
The patch looks good to me and tested for for me. You are done testing right @bgly?

bgly:
@mikechristie Yes, it has been working for me for awhile. I've been doing other filesystem integrations (zfs) with the unmap code.

mikechristie:
Oh wait, there is one bug. I will send you a patch to merged into your patch.

bgly:
> Oh wait, there is one bug. I will send you a patch to merged into your patch.

okay, ill merge and test quick.

mikechristie:
[extra-unmap-fixes.txt](https://github.com/open-iscsi/tcmu-runner/files/2933575/extra-unmap-fixes.txt)

The attached patch made over your PR, fixes the following issues:

1. Everytime align_and_split_unmap successfully submits a unmap to a handler we need to do a refcount++, to match the cmd->done->unmap_state_put call that the handler will do.

The refcount++ in handle_unmap_internal before "while (bddl) {" would only handle one align_and_split_unmap call. We need the refcount++ done in align_and_split_unmap because handle_unmap_in_writesame uses that function too.
 
2. If handle_unmap_internal never successfully calls align_and_split_unmap then handle_unmap_internal needs to fail the command when it calls unmap_state_put. The extra refcount++ at the beginning of "while (bddl) {" is going to leave an extra refcount from when it is allocated and the command will never be completed.

tcmur_cmd_handler.c
```diff
@@ -347,6 +341,10 @@ static int align_and_split_unmap(struct tcmu_device *dev,
 			tcmu_dev_dbg(dev, "There are totally %d splits\n", j);
 		}
 
+		pthread_mutex_lock(&state->lock);
+		state->refcount++;
+		pthread_mutex_unlock(&state->lock);
+
```
bgly: I moved this above the async_handle_cmd since the refcount is needed if the unmap_state_put is called before we get the ret from async_handle_cmd call. I then decrement it back down if the async_handle_cmd fails so that we dont have the extra refcount. This is the only change I made to your patch. I have tested and things work.

mikechristie: That's not needed because the caller has a refcount on it already. It would not go down to zero until after we return from align_and_split_unmap and the caller releases their refcount.

bgly: You mean the refcount is initialized to 1 thus we already have a refcount? So then we would want to change if(--state->refcount > 0) to if(--state->refcount => 0) ? TCMU-Runner crashes in my setup atm without the change to refcount++ prior to calling the async_handle_cmd

mikechristie: Oh no, wait you are right about needing to do refcount++ before calling async_handle_cmd.

mikechristie: We are ok on your side then right? It works for me now too.

bgly: yes, I am good if you are good!

mikechristie: Thanks!

PR #535, opened 2019-02-20
rpm: remove %clean section

ktdreyer:
This section is not needed for modern RPM versions.


ktdreyer:
Looks like RHEL 5 was the last version to require this.

PR #540, opened 2019-03-14
azblk handler version 1

caavery:
The azure page blob handler is based on and serves as a user-space
alternative to khenidak's dysk kernel block driver
(https://github.com/khenidak/dysk).

An azblk target device will be exposed to the local host as a TCMU scsi
device via the target loopback driver (targetcli loopback transport).

azblk currently supports SAS access credentials, lease ids,
local read only permissions, http, and https.

Each azblk target device must be configured through a configuration
file whose default location is /etc/tcmu/azblk.conf.

Sample config file:

device =
(
  {
    name = "azblk1_http";
    sas = "sp=rcwd&st=2019-02-22T15:00:55Z&se=2019-03-22T22:00:55Z&spr=https,http&sv=2018-03-28&sig=anIkHRlkadfjlkajdQVqIvENpLO9FL1zyQ%3D&sr=b";
    lease = "48e29061-079d-446c-833c-5d884ab5ec1e";
    url = "http://myazureaccount.blob.core.windows.net/test/test1.vhd";
  },
  {
    name = "azblk2_https";
    sas = "sp=rcwd&st=2019-03-07T13:43:02Z&se=2019-04-04T20:43:02Z&spr=https,http&sv=2018-03-28&sig=MW3SK5h99alskdfjlkjnvV3xz%2BLLcAhDVI%3D&sr=b";
    readonly = 1;
    url = "https://myazureaccount.blob.core.windows.net/test/test2.vhd";
  }
);

Signed-off-by: Cathy Avery <cavery@redhat.com>

### DIFF
--- a/azblk.c
+++ b/azblk.c
@@ -1,0 +1,1006 @@
+/*
+ * Copyright (c) 2018 Red Hat, Inc.
+ *
+ * This file is licensed to you under your choice of the GNU Lesser
+ * General Public License, version 2.1 or any later version (LGPLv2.1 or
+ * later), or the Apache License 2.0.
+ */
+
+#define _GNU_SOURCE
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <uv.h>
+#include <errno.h>
+#include <time.h>
+#include <scsi/scsi.h>
+#include <curl/curl.h>
+#include <unistd.h>
+#include <glib.h>
+#include <libconfig.h>
+#include "scsi_defs.h"
+#include "libtcmu_common.h"
+#include "tcmu-runner.h"
+#include "tcmur_cmd_handler.h"
+#include "libtcmu.h"
+#include "tcmur_device.h"
+#include "ccan/list/list.h"
+
+// Http Response processing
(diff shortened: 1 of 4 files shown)


azblk.c
```diff
@@ -0,0 +1,1006 @@
...
+			len = strlen(sas);
+			if (len > (AZ_SAS_LEN - 1)) {
+				asprintf(&err_msg,
+					 "File %s: device %s sas must be less than %d characters\n",
+					 ddev->cfg.config_file,
+					 tcm_dev_name,
+					 AZ_SAS_LEN);
+				goto error;
+			}
+			strncpy(ddev->cfg.sas, sas, len);
```
lxbsz: Here we'd better to use the strlcpy as Mike has introduce it from BSD locally.
https://github.com/open-iscsi/tcmu-runner/blob/master/strlcpy.c

azblk.c
```diff
@@ -0,0 +1,1006 @@
...
+	 * calling curl_multi_cleanup, curl_multi_remove_handle or
+	 * curl_easy_cleanup."
+	 */
+
+	while ((message = curl_multi_info_read(curl_multi, &pending))) {
+		switch (message->msg) {
+		case CURLMSG_DONE:
+			azblk_multi_done(curl_multi, message);
+			break;
+
```
lxbsz: Extra white line here.

PR #545, opened 2019-04-04
Add method to query whether OOOC is supported

geordieintheshellcode:
It's useful for users of libtcmu to be able to query at runtime whether or not the kernel supports out-of-order command (OOOC) completion as this has an impact on how handlers may service SCSI commands. Looking at the kernel version or module version of `target_core_user` is painful as it requires knowing exactly which versions support OOOC, some older kernel modules have the OOOC changes back-ported for example. 

In this commit:
* Expose a new method `tcmu_dev_oooc_supported()` which looks at the mailbox flags to determine whether OOOC is supported.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -381,7 +381,8 @@ void tcmu_dev_flush_ring(struct tcmu_device *dev)
 	tcmu_dev_dbg(dev, "ring clear\n");
 }
 
-bool tcmu_dev_oooc_supported(struct tcmu_device* dev) {
+bool tcmu_dev_oooc_supported(struct tcmu_device* dev)
+{
 	return dev->map->flags & TCMU_MAILBOX_FLAG_CAP_OOOC;
 }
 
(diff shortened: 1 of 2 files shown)


libtcmu.c
```diff
@@ -381,6 +381,10 @@ void tcmu_dev_flush_ring(struct tcmu_device *dev)
 	tcmu_dev_dbg(dev, "ring clear\n");
 }
 
+bool tcmu_dev_oooc_supported(struct tcmu_device* dev) {
```
mikechristie: Just following the existing coding style and put the opening "{" on a new line when declaring a function here. Check out tcmu_dev_get_private in the same file for an example. Thanks.

geordieintheshellcode: Thanks for taking a look Mike. I've updated the PR to fix the code style.

mikechristie:
Thanks.

PR #546, opened 2019-04-16
glfs: retry 5 times to wait the meta data updated successfully

lxbsz:
For the 'gluster-block modify' we can hit one issue in the same
client with multi-precesses, here is the gluster-blockd and
tcmu-runner.

We do glfs_ftruncate in gluster-blockd to resize the volume file
and then in tcmu-runner to check the size by using the glfs_lstat
but get the old size, which hasn't been updated yet.

But from the mountpoint's ls command, we can see that the size is
already upated to new one.

Here we will went to wait for most 5 seconds and retry 5 times to
make sure the cache has been flushed successfully to the volume.

Fixes: https://github.com/gluster/gluster-block/issues/204

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -672,9 +672,11 @@ static int tcmu_glfs_reconfig(struct tcmu_device *dev,
 	struct glfs_state *gfsp = tcmur_dev_get_private(dev);
 	struct stat st;
 	int ret = -EIO;
+	int retry = 0;
 
 	switch (cfg->type) {
 	case TCMULIB_CFG_DEV_SIZE:
+retry:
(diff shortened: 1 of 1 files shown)


lxbsz:
@pkalever @amarts 
Please help review, thanks.

mikechristie:
Patch looks sane to me. Let me know when you guys have reviewed and I will merge it.

lxbsz:
> Patch looks sane to me. Let me know when you guys have reviewed and I will merge it.

Hi Mike,
This is hack code, we have raise one bug in Gluster, let's wait for days.
More detail please see: https://github.com/gluster/gluster-block/pull/209

Thanks.

PR #548, opened 2019-04-29
dyn-config: try to read it again in case the pagecache is not flushed out

lxbsz:
In-case if the editor (vim) follows write to a new file (.swp, .tmp ..)
and move it to actual file name later. There is a window, where we will
encounter one case that the file data is not flushed to the disk, so in
another process(here) when reading it will be empty.

Let just wait and try again.

### DIFF
--- a/libtcmu_config.c
+++ b/libtcmu_config.c
@@ -195,7 +195,7 @@ static void tcmu_conf_set_options(struct tcmu_config *cfg)
 }
 
 #define TCMU_MAX_CFG_FILE_SIZE (2 * 1024 * 1024)
-static int tcmu_read_config(int fd, char *buf, int count)
+static int tcmu_read_config(int fd, char *buf, int count, int *err)
 {
 	ssize_t len;
 	int save = errno;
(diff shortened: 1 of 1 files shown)


libtcmu_config.c
```diff
@@ -355,12 +356,14 @@ int tcmu_load_config(struct tcmu_config *cfg)
...
 	char *buf;
-	int i;
+	int i, count = 0;
 
 	buf = malloc(TCMU_MAX_CFG_FILE_SIZE);
 	if (!buf)
 		return -ENOMEM;
 
+retry:
+    count++;
```
mikechristie: I think the tabbing/spacing got messed up for the count++ line.

libtcmu_config.c
```diff
@@ -376,8 +379,27 @@ int tcmu_load_config(struct tcmu_config *cfg)
...
+	 * another process(here) when reading it will be empty.
+	 *
+	 * And at the same time the old file maybe deleted, then the read will failed
+	 * with errno == ENOENT.
+	 *
+	 * Let just wait and try again.
+	 */
+	if (count <= 5 && (len == 0 || ret == -ENOENT)) {
+		tcmu_dbg("failed to read the config from file, retrying (%d/5) times\n",
+                 count);
```
mikechristie: Same for the "count);" line here. I think the tabbing/spacing got messed up.

PR #549, opened 2019-05-10
glfs: update the glfs logdir

lxbsz:
Update README to add uninstall info.
Update the logdir in glfs handler when the logdir is changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ We encourage pull requests and issues tracking via Github, and the [target-devel
    * *Note:* If using systemd, `-DSUPPORT_SYSTEMD=ON -DCMAKE_INSTALL_PREFIX=/usr` should be passed to cmake, so files are installed to the correct location.
 1. Type `make`
 1. Type `make install`
+1. Type `xargs rm < install_manifest.txt` to uninstall from source
 
 
 ##### Running tcmu-runner
--- a/glfs.c
+++ b/glfs.c
@@ -101,14 +101,12 @@ struct gluster_cacheconn {
 
 static darray(struct gluster_cacheconn *) glfs_cache = darray_new();
 
-
 const char *const gluster_transport_lookup[] = {
 	[GLUSTER_TRANSPORT_TCP] = "tcp",
 	[GLUSTER_TRANSPORT_UNIX] = "unix",
 	[GLUSTER_TRANSPORT_RDMA] = "rdma",
 	[GLUSTER_TRANSPORT__MAX] = NULL,
 };
-
 
 static void gluster_free_host(gluster_hostdef *host)
 {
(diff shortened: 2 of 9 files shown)


lxbsz:
@mikechristie @amarts @pkalever @nixpanic
Please review.
Thanks.

README.md
```diff
@@ -40,6 +40,7 @@ We encourage pull requests and issues tracking via Github, and the [target-devel
    * *Note:* If using systemd, `-DSUPPORT_SYSTEMD=ON -DCMAKE_INSTALL_PREFIX=/usr` should be passed to cmake, so files are installed to the correct location.
 1. Type `make`
 1. Type `make install`
+1. Type `xargs rm < install_manifest.txt` to uninstall from source
```
nixpanic: Is `make uninstall` not functional?

glfs.c
```diff
@@ -196,6 +194,33 @@ static int gluster_cache_add(gluster_server *dst, glfs_t *fs, char* cfgstring)
 	return -1;
 }
 
+static bool gluster_cache_update_logdir(const char *logdir)
```
nixpanic: Should `logdir` not be used somewhere in this function? Now it is only printed when there is an error.

README.md
lxbsz: @nixpanic
Yeah, for the cmake this is not functional as default, please see [1].

[1] https://gitlab.kitware.com/cmake/community/wikis/FAQ#can-i-do-make-uninstall-with-cmake

Or just add one as the wiki does ? Not sure whether will it make sense here by make it more complicated.

Thanks.

glfs.c
lxbsz: Yeah, right and will fix it.
This is from the old version and forget to remove it.
Thanks.

nixpanic:
> +1. Type `xargs rm < install_manifest.txt` to uninstall from source
> 
> @nixpanic
> Yeah, for the cmake this is not functional as default, please see [1].

In that case, I think the `xargs rm < install_manifest.txt` is fine.

lxbsz:
> +1. Type `xargs rm < install_manifest.txt` to uninstall from source @nixpanic Yeah, for the cmake this is not functional as default, please see [1].
> In that case, I think the `xargs rm < install_manifest.txt` is fine.

Cool, Thanks.
BRs

libtcmu_log.h
```diff
@@ -14,6 +14,8 @@
 #include <stdarg.h>
 #include <stdbool.h>
 
+#include "libtcmu_config.h"
```
mikechristie: Just do a forward declartion for struct tcmu_config like we do for struct tcmu_device.

lxbsz: Sure, will fix it.
Thanks.

PR #550, opened 2019-05-10
lock: create the lock directory when it is not exist

lxbsz:
We are hiting the following errors when the lock dir is
not exist:

2019-05-10 05:45:51.186 348 [CRIT] main:981: Starting...
2019-05-10 05:45:51.186 348 [ERROR] main:985: creat(/var/run/lock/tcmu.lock) failed: [No such file or directory]

### DIFF
--- a/main.c
+++ b/main.c
@@ -48,7 +48,7 @@
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
 
-# define TCMU_LOCK_FILE   "/var/run/lock/tcmu.lock"
+#define TCMU_LOCK_FILE   "/run/tcmu.lock"
 
 static char *handler_path = DEFAULT_HANDLER_PATH;
 


lxbsz:
@mikechristie @amarts @pkalever 
Please review.

Or maybe can we just use the /run/ instead of /var/run/lock/ and will this make sense ?

Thanks.

mikechristie:
Can we just want to use the systemd RuntimeDirectory variable? Will that not work correctly with containers?

lxbsz:
> Can we just want to use the systemd RuntimeDirectory variable? Will that not work correctly with containers?

Yeah, it works without any problem.

Currently only the lock/ dir in /var/run need to be created by tcmu-runner.

If there is no need to consider none systemd case, this will be most optimal resolution here.

Thanks.
BRs

lxbsz:
@mikechristie 
BTW, should we also set RuntimeDirectoryPreserve=yes ?

As far as I can see this will make sense here, for the /var/run/lock/ or /run/lock/ as one temporary dir, it will also be used by others services, without this option it seems will be removed when the tcmu-runner is stopped.

Thanks.

lxbsz:
Maybe we can just switch to use the /run/tcmu.lock or /var/run/tcmu.lock instead of /var/run/lock/tcmu.lock without considering backward compatibility.

```
# ls -l /var/
[...]
lrwxrwxrwx.  1 root root    6 Apr 20  2018 run -> ../run
[...]
```

lxbsz:
> @mikechristie
> BTW, should we also set RuntimeDirectoryPreserve=yes ?
> 

```
diff --git a/tcmu-runner.service b/tcmu-runner.service
index 1951a87..062a712 100644
--- a/tcmu-runner.service
+++ b/tcmu-runner.service
@@ -9,6 +9,8 @@ Type=dbus
 KillMode=process
 ExecStart=/usr/bin/tcmu-runner
+RuntimeDirectory=lock
+RuntimeDirectoryPreserve=yes
 
 [Install]
 WantedBy=multi-user.target

```

I test this but it seemed it not work for me. The runtime directorie '/run/lock/' will always be deleted when the service was stopped.

Thanks.
BRs

lxbsz:
@mikechristie 
Let's just switch to use the "/run/tcmu.lock" instead.  This will be okay for the container case about the backward compatibility.

Thanks.

lxbsz:
@nixpanic

Please help review and thanks.
BRs

nixpanic:
Using `/run/tcmu.lock` is reasonable. Otherwise you could include a `/etc/tmpfiles.d/tcmu-runner.conf` drop-in file that creates `/run/lock/tcmu.lock` before the tcmu-runner service is started.

lxbsz:
> Using `/run/tcmu.lock` is reasonable. Otherwise you could include a `/etc/tmpfiles.d/tcmu-runner.conf` drop-in file that creates `/run/lock/tcmu.lock` before the tcmu-runner service is started.

Cool, let's keep to use '/run/tcmu.lock'.

Thanks.
BR

SaravanaStorageNetwork:
Using /var/run/lock looks cleaner for me.   

I could see other storage related lock files/folders here. It makes sense to have tcmu lock too here.
 
\# tree /var/run/lock
/var/run/lock
├── asound.state.lock
├── iscsi
│   └── lock
├── kdump
├── lockdev
├── lvm
├── ppp
├── subsys
│   └── network
├── vdo
│   └── _etc_vdoconf.yml.lock
└── vdo-config-singletons

6 directories, 6 files

lxbsz:
> Otherwise you could include a `/etc/tmpfiles.d/tcmu-runner.conf` drop-in file that creates `/run/lock/tcmu.lock` before the tcmu-runner service is started.

For this, before we supposed to use the following options in tcmu-runner.service.

```
+RuntimeDirectory=lock
+RuntimeDirectoryPreserve=yes   <== this means the /var/run/lock/ dir won't be deleted when the service is stopped
```

But it do not work for me here. And the /var/run/lock/ dir will always be deleted when the tcmu-runner service is stopped.

@nixpanic @SaravanaStorageNetwork 
Any idea about this ? Did I miss something ?

> Using /var/run/lock looks cleaner for me.
> I could see other storage related lock files/folders here. It makes sense to have tcmu lock too here.

Currently it seems most of the services are using the /var/run/lock/, and some others, such as the rpcbind,  still using the /run/ directly.

Thanks

nixpanic:
It might be nicer to use `/run/lock/` , but some services place their `.lock` file under `/run/` directly too. The easy approach is to use `/run/tcmu.lock`, alternatively I'd recommend to use a `man tmpfiles.d` file that creates the directory+file. Doing it in the `.service` might work too, but it is not the solution I would pick (just because I like the other options more).

lxbsz:
> It might be nicer to use `/run/lock/` , but some services place their `.lock` file under `/run/` directly too. The easy approach is to use `/run/tcmu.lock`, alternatively I'd recommend to use a `man tmpfiles.d` file that creates the directory+file. Doing it in the `.service` might work too, but it is not the solution I would pick (just because I like the other options more).

@nixpanic @SaravanaStorageNetwork 

Thanks for you detail info.

If we won't hit any potential issue to use "/run/tcmu.lock", let's keep to use this to make it simple.

Thanks.
BRs

PR #554, opened 2019-06-04
glfs: add lock support

lxbsz:
The fencing doc: https://review.gluster.org/#/c/glusterfs-specs/+/21925/6/accepted/fencing.md

The fencing test example: https://review.gluster.org/#/c/glusterfs/+/21496/12/tests/basic/fencing/fence-basic.c


lxbsz:
@pkalever @amarts @spalai @mikechristie 

This still need test more later and any idea is welcome.

Thanks.

lxbsz:
When testing this, the glfs_file_lock() get stucked and filed one bug: https://bugzilla.redhat.com/show_bug.cgi?id=1717824

Thanks.

glfs.c
```diff
@@ -836,6 +874,56 @@ static int tcmu_glfs_handle_cmd(struct tcmu_device *dev, struct tcmulib_cmd *cmd
...
+	struct glfs_state *state = tcmur_dev_get_private(dev);
+	struct flock lock;
+	int ret;
+
+	lock.l_type = F_WRLCK | F_RDLCK;
+	lock.l_whence = SEEK_SET;
+	lock.l_start = 0;
+	lock.l_len = 0;
+
+	ret = glfs_file_lock(state->gfd, F_SETLK, &lock, GLFS_LK_MANDATORY);
```
mikechristie: If you already have the lock, then call this again will it return success or failure?

lxbsz: If the lock is already took by any of the other nodes, it will try to break it and take it in local node.

In the local node, all the cases are the same with the linux mandatory lock. Here it will be blocked if any other process has took it.

lxbsz: For this the test is not finished yet due to there is one bug in the gluster backend, which is till not fixed yet by the gluster core team.

lxbsz:
@spalai @nixpanic 

Please help review the GFAPI stuff, thanks very much.
BRs

glfs.c
```diff
@@ -530,6 +534,14 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
 		goto unref;
 	}
 
+#if GFAPI_VERSION770
+	ret = glfs_fsetxattr(gfsp->gfd, GF_ENFORCE_MANDATORY_LOCK, "set", 8, 0);
```
nixpanic: isn't size 8 meant to be the size of the value that you are setting (`"set"` would be 4 with the `'\0'`).

lxbsz: Hmm, yeah that's true, I will fix it.
Thanks :-)

lxbsz: @nixpanic

BTW, which version should I check here ? It seems the 7dev is not correct.

Thanks.

amarts: 7dev (or now 8dev) is only set on development branch. I would say for now (till 8dev becomes 8.0) it should be 8dev??

Let @nixpanic confirm.

lxbsz: Yeah,  the 8.0 seems the correct one.

lxbsz:
@nixpanic @amarts 
Updated and please review again.
Thanks.

nixpanic:
`GF_ENFORCE_MANDATORY_LOCK` was introduced with GlusterFS 6.0. That would be (pkg-config) glusterfs-api version 7.6. What is the reason to go for GlusterFS version 8?

lxbsz:
> `GF_ENFORCE_MANDATORY_LOCK` was introduced with GlusterFS 6.0. That would be (pkg-config) glusterfs-api version 7.6. What is the reason to go for GlusterFS version 8?

This is because the fencing patches will goto version 8, which still not merged yet, please see: https://review.gluster.org/23088.

Will it make sense ?
Thanks.

nixpanic:
Fencing support has been added to GlusterFS 6.0 with https://review.gluster.org/c/glusterfs/+/21496 . If there are bugfixes that you depend on, those should ideally be backported to maintained releases. Maybe check with @spalai about that?

lxbsz:
@nixpanic 
Okay, thanks very much for your detail explanation of this.

BRs

lxbsz:
@spalai

As we disscussed, the fencing fix patches will be backported since the release 6.6.

I have updated it here. Could you confirm it here ?

Thanks.
BRs

spalai:
Will 

> @spalai
> 
> As we disscussed, the fencing fix patches will be backported since the release 6.6.
> 
> I have updated it here. Could you confirm it here ?
> 
> Thanks.
> BRs

Sure will backport to rel-6 & 7,

lxbsz:
> Will
> 
> > @spalai
> > As we disscussed, the fencing fix patches will be backported since the release 6.6.
> > I have updated it here. Could you confirm it here ?
> > Thanks.
> > BRs
> 
> Sure will backport to rel-6 & 7,

@spalai,

Thanks very much.

The currently release is 6.5, so will depend on 6.6 then.

`pkg_check_modules(GFAPI766 glusterfs-api>=7.6.6 QUIET)`

Thanks.

lxbsz:
@mikechristie 

This is ready now, please review again.

Thanks.
BRs

mikechristie:
Patch looks ok to me. Just 2 questions/comments.

1. get_lock_tag is a hack for explicit failover with ESX. We currently do not enable explicit failover, because it is buggy, so no need to implement it.

2. get_lock_state is not required but it might be useful. During a failback, the gw we are failing back to might still think it has the lock at the runner level, so we will send the IO down to the handler. glfs's handler would then send the IO down to whatever handles lock checking states. It looks like glfs would get an error and then you will call the lock lost function, so new IO will kick of a lock acquisition.

get_lock_state is normally called before IO is sent during a failback. So when a path is added back on the multipath side that layer would do a RTPG and/or inquiry to reprobe the path. At this time we call that callout to update the lock state.

So the callout is not required since you will handle the issue as part of the normal error handling, but it might speed things up because there will be less IO in flight to flush in runner core and the handler before doing the lock acquisition since the first IO would kick off acquisition.

lxbsz:
> Patch looks ok to me. Just 2 questions/comments.
> 
> 1. get_lock_tag is a hack for explicit failover with ESX. We currently do not enable explicit failover, because it is buggy, so no need to implement it.

Currently we supposed to support the implicit without the ESX usage only, so yes this shouldn't be problem for glfs for now.

> 2. get_lock_state is not required but it might be useful. During a failback, the gw we are failing back to might still think it has the lock at the runner level, so we will send the IO down to the handler. glfs's handler would then send the IO down to whatever handles lock checking states. It looks like glfs would get an error and then you will call the lock lost function, so new IO will kick of a lock acquisition.

Yeah, this is really a very useful callout, for the currently implementation there is no any directly API could do this, I have tried the glfs_file_lock(F_GETLK) which is equal to fcntl(F_GETLK), but currently this won't work, because from my test the PID member in 'struct flock {}' won't work as expect as what we needs in distribute case for gusterfs.

> 
> get_lock_state is normally called before IO is sent during a failback. So when a path is added back on the multipath side that layer would do a RTPG and/or inquiry to reprobe the path. At this time we call that callout to update the lock state.
> 
> So the callout is not required since you will handle the issue as part of the normal error handling, but it might speed things up because there will be less IO in flight to flush in runner core and the handler before doing the lock acquisition since the first IO would kick off acquisition.
> 

Yeah, I had hit the similar issue you mentioned above:

When I test the Fencing stuff, I found there is one problem, every first time when I mkfs of the new mpath device, it will fail for the first time only.

[root@client ~]# mkfs.xfs -f /dev/mapper/mpathv
meta-data=/dev/mapper/mpathv     isize=512    agcount=4, agsize=12800 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=0, sparse=0
data     =                       bsize=4096   blocks=51200, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0 ftype=1
log      =internal log           bsize=4096   blocks=855, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
existing superblock read failed: Input/output error
======================

And in the gluster-block/tcmu-runner node side, I can see there has the -11(-EAGAIN) errno:

2019-08-01 07:31:47.230 60714 [DEBUG] tcmu_acquire_dev_lock:440 glfs/block1: lock call done. lock state 1
2019-08-01 07:31:47.230 60714 [DEBUG] tcmu_cfgfs_dev_exec_action:329 glfs/block1: executing action block_dev
2019-08-01 07:31:47.230 60714 [DEBUG] tcmu_cfgfs_dev_exec_action:331 glfs/block1: action block_dev done
2019-08-01 07:31:47.338 60714 [INFO] glfs_async_cbk:646 glfs/block0: lxb ------------------ ret = -1, err = -11.
2019-08-01 07:31:47.348 60714 [INFO] glfs_async_cbk:646 glfs/block0: lxb ------------------ ret = -1, err = -11.
2019-08-01 07:31:59.606 60714 [INFO] glfs_async_cbk:646 glfs/block1: lxb ------------------ ret = -1, err = -11.
2019-08-01 07:31:59.612 60714 [INFO] glfs_async_cbk:646 glfs/block1: lxb ------------------ ret = -1, err = -11.
++++++++++++++

So after confirmed with @palai EAGAIN and EBUSY should be for non-blockiing locks in general. And I fixed it with tigger the lost connection when there is EBUSY too. And certainly the first tens of IOs will be retried and will somehow delay the mkfs or the applications who are trying to use the new devices.

Except this issue I have's hit any others till now.


> It does not matter to me, so let me know and I can merge what you have and you add the callout later or never.

For now all my tests are working well, so let's do this later if we really need it.
Thanks Mike,
BRs

tcmur_device.c
```diff
@@ -437,7 +437,7 @@ int tcmu_acquire_dev_lock(struct tcmu_device *dev, uint16_t tag)
 	else
 		rdev->lock_state = TCMUR_DEV_LOCK_UNLOCKED;
 
-	tcmu_dev_dbg(dev, "lock call done. lock state %d\n", rdev->lock_state);
+	tcmu_dev_info(dev, "lock call done. lock state %d\n", rdev->lock_state);
```
mikechristie: If we are going to make it a "info" level message then let's make it more user friendly.  Something like:

tcmu_dev_info(dev, "Lock acquisition %s\n", rdev->lock ==  TCMUR_DEV_LOCK_LOCKED ? "successful" : "unsuccessful");

lxbsz: @mikechristie 
Done.
Thanks.

PR #555, opened 2019-06-06
Optionally link against tcmalloc for improved small IO performance

dillaman:
Under high-throughput, small IO size workloads, librbd can see up to a
20% performance increase when using tcmalloc instead of the default
glibc allocator.


bgly:
You probably want to modify the install_dep.sh script to include gperftools-devel?

dillaman:
> You probably want to modify the install_dep.sh script to include gperftools-devel? @dillaman

#557

PR #559, opened 2019-06-13
cmake: add a minor .so version number to libtcmu

ddiss:
The minor .so version number should be bumped when backwards compatible
ABI changes are made.

This commit results in the following build output:
  libtcmu.so -> libtcmu.so.2
  libtcmu.so.2 -> libtcmu.so.2.1
  libtcmu.so.2.1

Link: https://www.spinics.net/lists/target-devel/msg17636.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ add_library(tcmu
   )
 set_target_properties(tcmu
   PROPERTIES
+  VERSION 2.1
   SOVERSION "2"
   )
 target_include_directories(tcmu


mikechristie:
@ddiss,

Just to make sure I understand this correctly for future updates.

We are working on a stable lib (hoping to do this in the next next release), but until that time we are going to break the API. For every release that happens, then I should update the VERSION value being added in this PR correct? So for the next release, I just bump it to 2.2.

ddiss:
@mikechristie IIUC, backwards compatible ABI changes (e.g. new symbols) should lead to a bump of the second digit. If an ABI change isn't backwards compatible (e.g. symbol sig change or removal) then the library's first (major) version number should be bumped.

mikechristie:
@ddiss 

How do you guys handle libmultipath changes where they also do not have a stable ABI? Do you have similar issues and are you packaging it differently?

ddiss:
@mikechristie I've thankfully managed to avoid that. @mwilk might be able to comment - by the looks of  https://build.opensuse.org/package/view_file/Base:System/multipath-tools/multipath-tools.spec?expand=1 we do something to minimize the impact of the unstable ABI:
```
# Currently, it makes no sense to split out libmpathpersist and libmpathcmd
# separately. libmultipath has no stable API at all, and it depends
# on libmpathcmd (to be fixed). libmpathpersist depends on libmultipath
# and it loads prioritizers (to be fixed) and checkers.
```

PR #560, opened 2019-06-14
configfs: fix index out of bounds

pengliang233:
The ret of function call "read(fd, buf, sizeof(buf))" maybe 0, so the array "buf[ret-1]" will out of bounds.

### DIFF
--- a/configfs.c
+++ b/configfs.c
@@ -101,8 +101,12 @@ uint64_t tcmu_cfgfs_dev_get_info_u64(struct tcmu_device *dev, const char *name,
 			 strerror(errno));
 		*fn_ret = -EINVAL;
 		return 0;
+	} else if (ret == 0) {
+		tcmu_err("Invalid device info.\n");
+		*fn_ret = -EINVAL;
+		return 0;
 	}
-	buf[sizeof(buf)-1] = '\0'; /* paranoid? Ensure null terminated */
+	buf[ret-1] = '\0'; /* paranoid? Ensure null terminated */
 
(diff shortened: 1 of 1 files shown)


configfs.c
```diff
@@ -189,7 +189,7 @@ char *tcmu_cfgfs_dev_get_wwn(struct tcmu_device *dev)
 	}
 
 	/* Kill the trailing '\n' */
-	buf[ret-1] = '\0';
+	buf[sizeof(buf)-1] = '\0';
```
mikechristie: I am not sure that will do what we want. The kernel returns a string with a prefix string, the wwn, then a newline:

T10 VPD Unit Serial Number: 163c422a-4fbf-426e-9757-51ef08cc9861

If buf is 1024 chars, then the null is going to go at 1023. We want it to go at

strlen(T10 VPD Unit Serial Number: 163c422a-4fbf-426e-9757-51ef08cc9861) - 1

Have you got ret=0 for this function? It seems like it should never happen and we should always at least get that prefix and a newline even if the serial is not set: "T10 VPD Unit Serial Number:\n".

If we are getting ret=0 then I think you just want to put a

```
if (!ret) {
      tcmu_err("Invalid empty T10 VPD serial number.\n")l 
      return NULL
}
```

in there or the end of that function is going to just return a bogus value.

I think you also want to fix up tcmu_cfgfs_dev_get_info_u64, so it catches ret=0 and return -EINVAL for that case.

pengliang233: Yes, I want to fix up ret=0 just like what you said at first until I see tcmu_cfgfs_dev_get_info_u64.

PR #563, opened 2019-06-20
tcmu: reconfig_device convert err log to debug

pkalever:
These logs looks annoying with ERROR tags:

2019-06-14 17:56:33.567 1887 [ERROR] reconfig_device:195 glfs/blockvol:
Handler reconfig failed with error -95.
2019-06-14 17:56:33.569 1887 [ERROR] reconfig_device:195 glfs/blockvol:
Handler reconfig failed with error -95.
2019-06-14 17:56:33.577 1887 [ERROR] reconfig_device:195 glfs/blockvol:
Handler reconfig failed with error -95.
2019-06-14 17:56:33.578 1887 [ERROR] reconfig_device:195 glfs/blockvol:
Handler reconfig failed with error -95.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -154,7 +154,7 @@ static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 	}
 
 	if (!dev->handler->reconfig) {
-		tcmu_dev_err(dev, "Reconfiguration is not supported with this device.\n");
+		tcmu_dev_dbg(dev, "Reconfiguration is not supported with this device.\n");
 		return -EOPNOTSUPP;
 	}
 
@@ -177,8 +177,8 @@ static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 
 	ret = dev->handler->reconfig(dev, &cfg);
 	if (ret < 0) {
-		tcmu_dev_err(dev, "Handler reconfig failed with error %d.\n",
-			     ret);
+		tcmu_dev_dbg(dev, "Handler reconfig failed with error %s.\n",
+		             strerror(-ret));
 		return ret;
 	}
 
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -137,6 +137,12 @@ lookup_dev_by_name(struct tcmulib_context *ctx, char *dev_name, int *index)
 	return NULL;
 }
 
+static const char *const tcmulib_cfg_type_lookup[] = {
+	[TCMULIB_CFG_DEV_CFGSTR]  = "TCMULIB_CFG_DEV_CFGSTR",
+	[TCMULIB_CFG_DEV_SIZE]    = "TCMULIB_CFG_DEV_SIZE",
+	[TCMULIB_CFG_WRITE_CACHE] = "TCMULIB_CFG_WRITE_CACHE",
+};
+
 static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 			   struct genl_info *info)
 {
@@ -151,11 +157,6 @@ static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 		tcmu_err("Could not reconfigure device %s: not found.\n",
 			 dev_name);
 		return -ENODEV;
-	}
-
-	if (!dev->handler->reconfig) {
-		tcmu_dev_dbg(dev, "Reconfiguration is not supported with this device.\n");
-		return -EOPNOTSUPP;
 	}
 
 	if (info->attrs[TCMU_ATTR_DEV_CFG]) {
@@ -175,10 +176,16 @@ static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 		return -EOPNOTSUPP;
 	}
 
+	if (!dev->handler->reconfig) {
+		tcmu_dev_dbg(dev, "Reconfiguration is not supported with this device. "
+		             "Request for %s.\n", tcmulib_cfg_type_lookup[cfg.type]);
+		return -EOPNOTSUPP;
+	}
+
 	ret = dev->handler->reconfig(dev, &cfg);
 	if (ret < 0) {
-		tcmu_dev_dbg(dev, "Handler reconfig failed with error %s.\n",
-		             strerror(-ret));
+		tcmu_dev_dbg(dev, "Handler reconfig for %s failed with error %s.\n",
+		             tcmulib_cfg_type_lookup[cfg.type], strerror(-ret));
 		return ret;
 	}
 
--- a/api.c
+++ b/api.c
@@ -348,7 +348,7 @@ void tcmu_cdb_print_info(struct tcmu_device *dev,
 	sprintf(buf + n, "\n");
 
 	if (info) {
-		tcmu_dev_warn(dev, "%s", buf);
+		tcmu_dev_dbg(dev, "%s", buf);
 	} else {
 		tcmu_dev_dbg_scsi_cmd(dev, "%s", buf);
 	}


pkalever:
@bgly added one another case. Plenty of these logs together annoy!

libtcmu.c
```diff
@@ -177,8 +177,8 @@ static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,
 
 	ret = dev->handler->reconfig(dev, &cfg);
 	if (ret < 0) {
-		tcmu_dev_err(dev, "Handler reconfig failed with error %d.\n",
-			     ret);
+		tcmu_dev_dbg(dev, "Handler reconfig failed with error %s.\n",
+		             strerror(-ret));
```
lxbsz: There also has some other errors like -EINVAL in glfs.c 
Maybe only when ret == -EOPNOTSUPP, make this message as DEBUG or WARNING, and all the others as what it is.

mikechristie: Is 100% of the annoying logging due to not supported errors (-95)? If so then I think you want to change the tcmu_dev_err at the begging of the function that checks for if a reconfig callout is set.

mikechristie: I think it might be ok to just make them all dbg statements, because the caller will log the same info that a reconfigure operation failed and the reason why.

mikechristie: For the glfs resize failure case for example, we would want to leave the tcmu_dev_err call in tcmu_glfs_reconfig because it has extra info we can't express with an error code. We do not need to log a message above though since  tcmu_glfs_reconfig logs a detailed error and then the caller (rtslib/targetcli normally) will also log the same info as above.

lxbsz: No, as mentioned above, we also possible will have -EINVAL errno.

But even the reconfig callout is support, it will be like:

```
690 static int tcmu_glfs_reconfig(struct tcmu_device *dev,
691                               struct tcmulib_cfg_info *cfg)       
692 {  
693         struct glfs_state *gfsp = tcmur_dev_get_private(dev); 
694         struct stat st;        
695         int ret = -EIO;        
696    
697         switch (cfg->type) {   
698         case TCMULIB_CFG_DEV_SIZE:
699                 ret = glfs_lstat(gfsp->fs, gfsp->hosts->path, &st);                                                                                                      
700                 if (ret) {     
701                         tcmu_dev_warn(dev, "glfs_lstat failed: %m\n");
702                         tcmu_notify_conn_lost(dev);         
703    
704                         /* Let the targetcli command return success */
705                         ret = 0;
706                 } else if (st.st_size != cfg->data.dev_size) {
707                         tcmu_dev_err(dev,                   
708                                      "device size and backing size disagree: device %"PRId64" backing %lld\n",
709                                      cfg->data.dev_size, (long long) st.st_size);
710                         ret = -EINVAL;                      
711                 }
712                 return ret;    
713         case TCMULIB_CFG_DEV_CFGSTR:        
714         case TCMULIB_CFG_WRITE_CACHE:       
715         default:
716                 return -EOPNOTSUPP;
717         }
718 }  
```

Here the most cases we hit this annoying logging is due to TCMULIB_CFG_WRITE_CACHE no support.

lxbsz: Yeah, if so, that looks good to me.

mikechristie: I think we are talking about the same thing and also different things :)

I was just asking about the really annoying messages. The ones due to target_core_user exporting all attrs even if tcmu-runner handlers do not support them, then rtslib always trying to init them during device creation. For those cases we get the op not supported (-95 errors) here:

https://github.com/open-iscsi/tcmu-runner/blob/b1145f595d960b654295779e3a6429506e202490/libtcmu.c#L156

and

https://github.com/open-iscsi/tcmu-runner/blob/b1145f595d960b654295779e3a6429506e202490/libtcmu.c#L180

pkalaver fixed the second one which glfs hits, but did not fix the first one which can be hit by other handlers.

lxbsz: Yeah, make sense to me.

Currently we are only hitting the second one.

mikechristie:
@pkalever  when you repush the patches to handle that one comment, could you add more useful git commit messages. For example for the first patch just saying the log message is annoying is not helpful. At least tell us its annoying because reason "xyz" and is not needed because reason "abc". For the second patch tell us that it's logged on the initiator or the initiator does not care if those commands fail due to not being supported. Something other than you just don't like it :)

pkalever:
@mikechristie @lxbsz updated the commit msgs.

I had intentionally left tcmu_dev_err at:
```
static int reconfig_device(struct tcmulib_context *ctx, char *dev_name,                                                                                                                                            
         struct genl_info *info)                                                 
{ 
[...]
	if (!dev->handler->reconfig) {
		tcmu_dev_err(dev, "Reconfiguration is not supported with this device.\n");
		return -EOPNOTSUPP;
	}
[...]
}
```

1. I feel every handler should register the reconfig with bare min things that they can support. If nothing is supported they should just handle that in the handler code and return EOPNOTSUPP;

```
static int tcmu_MyHandler_reconfig(struct tcmu_device *dev, struct tcmulib_cfg_info *cfg)                                         
{                                                                                
  switch (cfg->type) {                                                           
  case TCMULIB_CFG_DEV_SIZE:                                                     
  case TCMULIB_CFG_DEV_CFGSTR:                                                   
  case TCMULIB_CFG_WRITE_CACHE:                                                  
  default:                                                                       
    return -EOPNOTSUPP;                                                          
  }                                                                              
}

struct tcmur_handler tcmu_MyHandler_handler = { 
[...]
  .reconfig      = tcmu_MyHandler_reconfig,
[...]
};

```

This ensure that the handler maintainers are knowingly not supporting the reconfig.

2. For the same reason why Mike suggested to put tcmu_dev_err() in the tcmu_glfs_reconfig for resize failure case. But here, as the handlers didn't handle them, lets throw the some error for notifying purpose.

I'm okay to change it to  tcmu_dev_warn(), but modifying it to dbg is not a good idea IMHO.

Thoughts are welcome :-)

mikechristie:
I disagree with the catching CFG commands explicitly in the handlers, because 

1. Supporting reconfig is optional and supporting each operation is optional.

Note:
While reviewing your patch I made a patch to remove the explicit checks in each handler's reconfig. 

2. We do not have control over every handler at the lib level or tcmu-runner level, so we have to handle it in those layers and not rely on the handlers doing it. By just handling it there we also avoid duplicating the work. There is no need for the extra clutter in each handler.

3. It fits with how we do other operations in the handler callouts where we do not check for every operation. We just check for what we support then default out of a switch for the rest.

4. There is no code to enforce having to add a reconfig callout. If we added one it would seem odd to add a check for it, but allow handlers to just have a empty handler callout so they full fill a requirement to do nothing. It just adds extra code. How would you even enforce that the handler had a specific check for each CFG operation.

If the callout was required because we needed it to do some work like opening a device which has to be done I would be with you.

libtcmu.c
lxbsz: @pkalever @mikechristie 

Will it make sense to print the cfg.type here like:

```
+               tcmu_dev_dbg(dev, "Handler reconfig for %u failed with error %d.\n",
+                            (unsigned int)cfg.type, ret);
```

Or we must guess what the unsupport type is every time.

Thanks.

lxbsz: With the latest kernel which has support the write_cache and dev_config attrs, when restarting the tcmu-runner it will be:

2019-06-22 12:08:14.534 9999 [INFO] tcmu_glfs_reconfig:689: TCMULIB_CFG_DEV_CFGSTR not sup!
2019-06-22 12:08:14.534 9999 [ERROR] reconfig_device:196 glfs/block3: Handler reconfig failed with error -95.
2019-06-22 12:08:14.534 9999 [INFO] tcmu_glfs_reconfig:692: TCMULIB_CFG_WRITE_CACHE not sup!
2019-06-22 12:08:14.534 9999 [ERROR] reconfig_device:196 glfs/block3: Handler reconfig failed with error -95.

mikechristie: Yeah.

It is ok if pkalever wants to sneak it in here or we can do it in another patch/PR since pkalever's original goal of this PR/patch was making the logs nicer from a flooding standpoint.

lxbsz: I am okay with this.
Thanks.

pkalever:
@mikechristie @lxbsz updated the PR. Please take a look.

From a different view, I think Mike's comments about handling 'reconfig not sup' in a common place also makesense.

Thanks!

PR #570, opened 2019-07-08
alua: fix pthread attr destroy

pengliang233:


alua.c
```diff
@@ -587,6 +587,8 @@ int alua_implicit_transition(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 		ret = TCMU_STS_BUSY;
 	}
 
+	pthread_attr_destroy (&attr);
```
mikechristie: Drop the extra space so instead of

pthread_attr_destroy (&attr);

its

pthread_attr_destroy(&attr);

Thanks.

pengliang233: OK

PR #580, opened 2019-07-30
Add link to tcmu-runner manpage from tcmu-runner.service

sathieu:


sathieu:
Caught by lintian (Debian static `.deb` checker).

PR #583, opened 2019-07-31
spec: fix build error

lxbsz:
Since the header files won't be installed, so we need to remove it
from the spec file too.


mikechristie:
Sorry about that @lxbsz. I threw in that last change and forgot to update the spec.

I made a new release so rpm based builds will work:

https://github.com/open-iscsi/tcmu-runner/releases/tag/v1.5.1

PR #588, opened 2019-08-23
tcmu_dev_get_memory_info()

dfsmith:
Our handler/device needs more information about the tcmu shared memory block (so that it can pass offsets to another process).  This patch adds a function to retrieve that information about the shared memory.

Also updated install_dep.sh so that it works with Debian installations.

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -43,9 +43,9 @@ static struct nla_policy tcmu_attr_policy[TCMU_ATTR_MAX+1] = {
 	[TCMU_ATTR_SUPP_KERN_CMD_REPLY] = { .type = NLA_U8 },
 };
 
-static int add_device(struct tcmulib_context *ctx, char *dev_name,
+static int device_add(struct tcmulib_context *ctx, char *dev_name,
 		      char *cfgstring, bool reopen);
-static void remove_device(struct tcmulib_context *ctx, char *dev_name,
+static void device_remove(struct tcmulib_context *ctx, char *dev_name,
 			  bool should_block);
 static int handle_netlink(struct nl_cache_ops *unused, struct genl_cmd *cmd,
 			  struct genl_info *info, void *arg);
@@ -217,13 +217,13 @@ static int handle_netlink(struct nl_cache_ops *unused, struct genl_cmd *cmd,
 	switch (cmd->c_id) {
 	case TCMU_CMD_ADDED_DEVICE:
 		reply_cmd = TCMU_CMD_ADDED_DEVICE_DONE;
-		ret = add_device(ctx, buf,
+		ret = device_add(ctx, buf,
 				 nla_get_string(info->attrs[TCMU_ATTR_DEVICE]),
 				 false);
 		break;
 	case TCMU_CMD_REMOVED_DEVICE:
 		reply_cmd = TCMU_CMD_REMOVED_DEVICE_DONE;
-		remove_device(ctx, buf, false);
+		device_remove(ctx, buf, false);
 		ret = 0;
 		break;
 	case TCMU_CMD_RECONFIG_DEVICE:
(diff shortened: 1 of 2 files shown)


libtcmu.c
```diff
@@ -491,8 +492,9 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 	}
 
 	snprintf(str_buf, sizeof(str_buf), "/dev/%s", dev_name);
+	dev->mmap_name = strndup(str_buf, sizeof(str_buf));
```
lxbsz: Maybe we can just delete the str_buf here and use asprintf(&dev->map_name, "/dev/%s", dev_name);

libtcmu_common.h
```diff
@@ -109,6 +109,8 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *priv);
 void *tcmu_get_daemon_dev_private(struct tcmu_device *dev);
 void tcmu_set_daemon_dev_private(struct tcmu_device *dev, void *priv);
 int tcmu_dev_get_fd(struct tcmu_device *dev);
+const char *tcmu_dev_get_memory_info(struct tcmu_device *dev,
+	void **base, size_t *len, off_t *offset);
```
mikechristie: Follow the formatting in the file where this line should be tabbed and possibly spaced over more to match up with the char after the "(".

libtcmu.c
```diff
@@ -924,6 +928,27 @@ int tcmu_dev_get_fd(struct tcmu_device *dev)
...
+ * information in struct iovec can be converted using these values.  Returns
+ * the filename of the mmap() that fd uses, or NULL if information is not
+ * available.
+ * @dev: tcmu device
+ * @base: receives the base address of the shared memory region
+ * @len: receives the size of the shared memory region
+ * @offset: receives the offset within fd to the shared memory region (conventionally 0)
+ */
+const char *
+tcmu_dev_get_memory_info(struct tcmu_device *dev, void **base, size_t *len, off_t *offset)
```
mikechristie: We try to do 80 chars. Checkout the other functions in the file.

libtcmu.c
```diff
@@ -924,6 +928,27 @@ int tcmu_dev_get_fd(struct tcmu_device *dev)
...
+{
+	if (!dev->mmap_name) return NULL;
+	if (base) *base = dev->map;
+	if (len) *len = dev->map_len;
+	if (offset) *offset = FDOFFSET;
```
mikechristie: Just follow the coding style in the rest of the code and put what comes after the test on a new line.

libtcmu.c
mikechristie: Also add a check for the failover case. You will probably want to rename the err_free tag to err_free_dev.

If (asprintf failed)
     goto err_free_dev;

And then add a new err_free_mmap_name goto and fix up the existing goto err_free lines.

libtcmu.c
```diff
@@ -556,6 +558,7 @@ static int add_device(struct tcmulib_context *ctx, char *dev_name,
 	if (reopen && reset_supp)
 		tcmu_cfgfs_dev_exec_action(dev, "block_dev", 0);
 err_free:
+	if (dev->mmap_name) free(dev->mmap_name);
```
mikechristie: See above comment. Fix up the gotos, so it goes to a specific unwinding and not a bulk free.

lxbsz:
@dfsmith 
BTW, in both of your commits there are no "Signed-off-by". 
Is that okay @mikechristie ?

All the others looks good to me.
Thanks.

extra/install_dep.sh
```diff
@@ -1,15 +1,23 @@
-#!/bin/sh
+#!/bin/bash
 #
```
mikechristie: Could you drop this patch from this set since its not related to the memory helper?

We have 2 other PRs trying to do something similar, so just comment on them or if you do not like them do a new PR with just this patch.

https://github.com/open-iscsi/tcmu-runner/pull/589
https://github.com/open-iscsi/tcmu-runner/pull/590

libtcmu.c
```diff
@@ -393,63 +393,186 @@ bool tcmu_dev_oooc_supported(struct tcmu_device* dev)
...
 
-static int add_device(struct tcmulib_context *ctx, char *dev_name,
-		      char *cfgstring, bool reopen)
+static ssize_t read_size(const char *filename)
 {
-	struct tcmu_device *dev;
-	struct tcmu_mailbox *mb;
-	char str_buf[256];
-	bool reset_supp = true;
+	/* Read a size_t from a file.  Returns -1 on error. */
```
mikechristie: Just stick the function description before start of it like:

```
/* Read a size_t from a file.  Returns -1 on error. */
static ssize_t read_size(const char *filename)

```

libtcmu.c
```diff
@@ -393,63 +393,186 @@ bool tcmu_dev_oooc_supported(struct tcmu_device* dev)
...
+	tcmu_warn("cannot read size from %s\n", filename);
+	return -1;
+}
+
+static bool device_parse_cfg(struct tcmu_device *dev,
+			     const char *dev_name, const char *cfgstring)
+{
+	/*
+	 * Extract configuration parameters into dev.
+	 */
```
mikechristie: Same as in read_size.

libtcmu.c
```diff
@@ -393,63 +393,186 @@ bool tcmu_dev_oooc_supported(struct tcmu_device* dev)
...
+err_badcfg:
+	tcmu_err("invalid cfgstring: expecting "
+		 "\"tcm-user/<hba_name>/<tcm_device_name>/<handler_name>\"\n");
+err_recompile: /* consider expanding string lengths in dev */
+	return false;
+}
+
+static void device_close_shm(struct tcmu_device *dev)
+{
+	int ret;
```
mikechristie: newline between variable declarations and start of code.

I will not flood you with the same comment. Check the other functions too.

libtcmu.c
```diff
@@ -393,63 +393,186 @@ bool tcmu_dev_oooc_supported(struct tcmu_device* dev)
...
+	if (len >= sizeof(dev->cfgstring)) {
+		tcmu_warn("additional handler cfgstring was truncated\n");
+		/* not a terminal error, but terminate anyway */
+		dev->cfgstring[sizeof(dev->cfgstring)-1] = '\0';
+	}
+	return true;
+
+err_badcfg:
+	tcmu_err("invalid cfgstring: expecting "
+		 "\"tcm-user/<hba_name>/<tcm_device_name>/<handler_name>\"\n");
```
mikechristie: I know it goes against the 80 char rule I wrote before, but for strings like this that get printed in log messages you can just keep them as one long string and not break it up:

`tcmu_err("invalid cfgstring: expecting "tcm-user/<hba_name>/<tcm_device_name>/<handler_name>\"\n");`

It seems to make it easier to grep for a log message in the source.

libtcmu.c
```diff
@@ -1015,15 +1004,45 @@ tcmu_dev_get_memory_info(struct tcmu_device *dev, void **base,
...
 	if (base)
 		*base = dev->map;
-	if (len)
-		*len = dev->map_len;
+	if (len) {
+		if (dev->map_len != 0) {
+			/* cached */
+			*len = dev->map_len;
+		}
+		else {
```
mikechristie: We do the else on the "}" line:

`} else {`

mikechristie:
Just some minor coding style nits. I swear I am not trying to be a pain in the  butt about that :) Thanks for the updated patch with all the cleanups!

extra/install_dep.sh
dfsmith: Strange coincidences!  Dropped.

dfsmith:
Thank you for spotting the nits.  Keep it up!

mikechristie:
Thanks!

PR #589, opened 2019-08-25
Add support for Debian

JKJameson:


JKJameson:
ACK

mikechristie:
Thanks. Could could you just roll up all the patches into one patch? We don't want to see the intermediate steps you took to fix up your patches in the main git tree.

JKJameson:
Patch has been rolled up. If you choose #590 over this push, please update the shell in extra/install_dep.sh from sh to bash. The script uses stuff that is only in BASH.

mikechristie:
You responded first so I will merge what you have :) Thanks!

@DifanZhang rebase and update your patches to take into account this PR.

PR #590, opened 2019-09-03
Support Ubuntu and Debian Package creation with Cmake

DifanZhang:
Verified on Debian buster and Ubuntu bionic. The following patch creates Debian package for Debian based OS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,17 @@ pkg_check_modules(KMOD REQUIRED libkmod)
 find_library(PTHREAD pthread)
 find_library(DL dl)
 
+# Stuff for building the Debian package
+set(CPACK_GENERATOR "DEB")
+set(CPACK_DEB_COMPONENT_INSTALL ON)
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "libnl-3-200, libnl-genl-3-200")
+set(CPACK_DEBIAN_PACKAGE_SECTION "net")
+set(CPACK_DEBIAN_PACKAGE_PRIORITY "optional")
+set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "A daemon that handles the userspace side of the LIO TCM-User backstore")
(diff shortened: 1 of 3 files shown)


mikechristie:
It looks like you and @JKJameson are trying to do the same things here and in

https://github.com/open-iscsi/tcmu-runner/pull/589

Please take a look at each others PRs.

CMakeLists.txt
```diff
@@ -47,6 +47,11 @@ set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Target Devel <target-devel@vger.kernel.org>
 set(CPACK_DEBIAN_PACKAGE_HOMEPAGE "https://github.com/open-iscsi/tcmu-runner")
 include(CPack)
```
mikechristie: I think you need to rebase and repush by doing a forced push, so we do not end up seeing this in the final pull and commit.

If everyone adds junk like this then it gets messy in the main git tree.

CMakeLists.txt
```diff
@@ -225,7 +225,7 @@ endif (with-zbc)
 
 if (with-rbd)
 	find_library(LIBRBD rbd)
-  set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, librados2, librbd1")
+	set(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, librados2, librbd1")
```
mikechristie: Just make the first patch "Add preliminary support to build Debian packages with Cmake infrastru…" use tabs from the beginning.

JuniorJPDJ:
This probably should be closed as #589 is merged already.

lxbsz:
> This probably should be closed as #589 is merged already.

The #589 is only supporting the `extra/install_dep.sh`, not sure whether will it work for the cmake ? I never test it yet on ubuntu/debian.

PR #591, opened 2019-09-20
tcmur: fix unlock for handle_caw

pengliang233:

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -1682,8 +1682,10 @@ static int handle_caw(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 
 	ret = aio_request_schedule(dev, tcmur_cmd, caw_work_fn,
 				   tcmur_cmd_complete);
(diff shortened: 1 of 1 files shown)


tcmur_cmd_handler.c
```diff
@@ -1682,8 +1682,10 @@ static int handle_caw(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 
 	ret = aio_request_schedule(dev, tcmur_cmd, caw_work_fn,
 				   tcmur_cmd_complete);
-	if (ret == TCMU_STS_ASYNC_HANDLED)
+	if (ret == TCMU_STS_ASYNC_HANDLED) {
+		pthread_mutex_unlock(&rdev->caw_lock);
```
mikechristie: We hold the lock while executing the IO so other CAWs are not running concurrently. The lock will be dropped in handle_caw_read_cbk or handle_caw_write_cbk depending on if the CAW fails early or completes successfully.

pengliang233: I get it. Thanks for your reply.

PR #592, opened 2019-09-27
logger: fix posible crash when loging to file

lxbsz:
When creating the file output object, we allow it failing and then
the logger should skip it later. If the failing really happen for
some reason, the tcmu-runner daemon will crash.


lxbsz:
Additional:

This is also the same with the syslog output object.

lxbsz:
@pkalever 
Please help review.
Thanks.

PR #594, opened 2019-10-03
Fix build on Fedora 31 beta

yuri-per:
Build is failing because of missing `-lpthread` option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ target_link_libraries(tcmu
   ${LIBNL_LIB}
   ${LIBNL_GENL_LIB}
   ${GLIB_LIBRARIES}
+  ${PTHREAD}
   ${TCMALLOC_LIB}
   )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -135,6 +135,7 @@ target_include_directories(tcmu-synthesizer
   )
 target_link_libraries(tcmu-synthesizer
   ${GLIB_LIBRARIES}
+  ${PTHREAD}
```
mikechristie: It's the files libtcmu_config.c and libtcmu_log.c in libtcmu in the tcmu chunk that need the pthread addition, right? I think we want to add this under the "target_link_libraries(tcmu" line so anything using tcmu gets it automatically like is done with the other libtcmu dependencies.

Same comment for the addition below.

yuri-per: Thanks, you are right. I have updated the code.

mikechristie:
Thanks.

PR #598, opened 2019-10-29
tcmu-runner: set a name for each pthread

lxbsz:
This will be helpful when debugging.

### DIFF
--- a/file_optical.c
+++ b/file_optical.c
@@ -1004,6 +1004,8 @@ static void *fbo_async_sync_cache(void *arg)
 	struct fbo_state *state = tcmur_dev_get_private(dev);
 	uint8_t sense[SENSE_BUFFERSIZE];
 
+	tcmu_set_thread_name("fbo-cache", dev);
+
 	pthread_mutex_lock(&state->state_mtx);
 	state->async_cache_count++;
 	state->flags |= FBO_BUSY_EVENT;
@@ -1318,6 +1320,8 @@ static void *fbo_async_format(void *arg)
 	struct tcmu_device *dev = (struct tcmu_device *)arg;
 	struct fbo_state *state = tcmur_dev_get_private(dev);
 	uint8_t sense[SENSE_BUFFERSIZE];
+
+	tcmu_set_thread_name("fbo-format", dev);
 
 	pthread_mutex_lock(&state->state_mtx);
 	state->flags |= FBO_BUSY_EVENT | FBO_FORMAT_IMMED;
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -835,6 +835,49 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *private)
 	dev->hm_private = private;
 }
 
+const char *tcmu_dev_get_uio_name(struct tcmu_device *dev)
+{
+	return dev->dev_name;
+}
+
+void tcmu_set_thread_name(const char *prefix, struct tcmu_device *dev)
+{
+	const char *uio = dev ? tcmu_dev_get_uio_name(dev) : NULL;
+	char *pname;
+	int end;
+
+	if (!prefix) {
+		tcmu_dev_err(dev, "failed to set name for thread %lu\n",
+			     pthread_self());
+		return;
+	}
+
+	end = strlen(prefix);
+
+	if (uio)
+	       end += strlen(uio) + 1; // extra 1 byte for '-'
+
+	if (end >= TCMU_THREAD_NAME_LEN) {
+		tcmu_dev_warn(dev, "the length for prefix(%s) + uio(%s) is (%d) should < (%d)\n",
+			      prefix, uio, end, TCMU_THREAD_NAME_LEN);
+		tcmu_dev_warn(dev, "will truncate the thread name to %d\n",
+			      TCMU_THREAD_NAME_LEN - 1);
+
+		/* truncate the thread name to 15 characters */
+		end = TCMU_THREAD_NAME_LEN - 1;
+	}
+
+	if (asprintf(&pname, "%s%s%s", prefix, uio ? "-" : "",  uio ? uio : "") == -1) {
+		tcmu_dev_err(dev, "no memory for pname\n");
+		return;
+	}
+
+	pname[end] = '\0';
+
+	pthread_setname_np(pthread_self(), pname);
+	free(pname);
+}
+
 void tcmu_dev_set_num_lbas(struct tcmu_device *dev, uint64_t num_lbas)
 {
 	dev->num_lbas = num_lbas;
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -521,6 +521,8 @@ static void *log_thread_start(void *arg)
 {
 	tcmu_logbuf = arg;
 
+	tcmu_set_thread_name("logger", NULL);
+
 	pthread_cleanup_push(log_cleanup, arg);
 
 	while (1) {
--- a/libtcmu_log.c
+++ b/libtcmu_log.c
@@ -403,6 +403,7 @@ static int output_to_fd(int pri, const char *timestamp,
 	int fd = (intptr_t) data;
 	char *buf, *msg;
 	int count, ret, written = 0, r, pid = 0;
+	char pname[TCMU_THREAD_NAME_LEN];
 
 	if (fd == -1)
 		return -1;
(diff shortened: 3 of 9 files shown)


pkalever:
@lxbsz this looks good to me, won't it be nice to use them in the debug logs too, may be using pthread_getname_np() ?

lxbsz:
> @lxbsz this looks good to me, won't it be nice to use them in the debug logs too, may be using pthread_getname_np() ?

@pkalever 
Yeah, this sounds a good idea.

lxbsz:
@pkalever 
Done. Thanks.

libtcmu.c
```diff
@@ -835,6 +835,11 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *private)
 	dev->hm_private = private;
 }
 
+const char *tcmu_dev_get_td_name(struct tcmu_device *dev)
```
mikechristie: Is "td" short of tcmu_dev? Could you rename to tcmu_dev_get_lio_name or tcmu_dev_get_kern_name to reflect is the lio/kernel backend name.

target.c
```diff
@@ -219,6 +219,8 @@ static void *tgt_port_grp_recovery_thread_fn(void *arg)
 	bool enable_tpg = false;
 	int ret;
 
+	pthread_setname_np(pthread_self(), "tpg");
```
mikechristie: Could you rename to "tpg-recovery"?

main.c
```diff
@@ -753,8 +753,16 @@ static void *tcmur_cmdproc_thread(void *arg)
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
 	struct pollfd pfd;
-	int ret;
 	bool dev_stopping = false;
+	char *pname;
+	int ret;
+
+	if (asprintf(&pname, "cmdproc-%s", tcmu_dev_get_td_name(dev)) == -1) {
```
mikechristie: The man page for pthread_setname_np states that the name can't be more than 16 chars including the terminating null. The tcmu lio device names are probably limited by what configfs can support and that seems to be the normal linux name length.

So maybe we want to use a different identifier.

libtcmu_log.c
```diff
@@ -411,10 +412,13 @@ static int output_to_fd(int pri, const char *timestamp,
 	if (pid <= 0)
 		return -1;
 
+	pthread_getname_np(pthread_self(), pname, 128);
```
mikechristie: Same comment as in the other thread. The man page says this can only be 16 chars, so no need to do 128.

When we figure out what name we want to use for the device id, then we should maybe do a
#define TCMU_THREAD_NAME_LEN 16

main.c
lxbsz: The hba name length range will be [6, 25], so this still won't work.

```
# mkdir /sys/kernel/config/target/core/user_12345678901234567890
# mkdir /sys/kernel/config/target/core/user_123456789012345678901
mkdir: cannot create directory ‘/sys/kernel/config/target/core/user_123456789012345678901’: Numerical result out of range

```

The dev_name will be the okay, it will be "uio%d" and the "%d" is from idev->minor.

```
 69 struct uio_device {            
 70         struct module           *owner;     
 71         struct device           *dev;       
 72         int                     minor;                                                                                                                                 
 73         atomic_t                event;      
 74         struct fasync_struct    *async_queue;
 75         wait_queue_head_t       wait;       
 76         struct uio_info         *info;      
 77         struct kobject          *map_dir;   
 78         struct kobject          *portio_dir;
 79 };

```
But the minor max length will be 1048576, which if from:

#define UIO_MAX_DEVICES         (1U << MINORBITS)
#define MINORBITS 20

In Arch 64, the "uio%d" length will range in [4, 11]

lxbsz:
@mikechristie 
Fix them all. Thanks.

libtcmu.c
```diff
@@ -835,6 +835,11 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *private)
 	dev->hm_private = private;
 }
 
+const char *tcmu_dev_get_uio_name(struct tcmu_device *dev)
+{
+	return dev->tcm_dev_name;
```
mikechristie: tcm_dev_name is the LIO name, but the function is named after the uio name. You probably wanted:

return dev->dev_name;

main.c
```diff
@@ -753,8 +753,16 @@ static void *tcmur_cmdproc_thread(void *arg)
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
 	struct pollfd pfd;
-	int ret;
 	bool dev_stopping = false;
+	char *pname;
+	int ret;
+
```
mikechristie: Make a function so we can handle errors once. Here we are just at the string len limit. We will probably never have more than 9999 LUNs (probably really 4k) so the uio name len would be at most "uio9999" (7 chars), and then for "cmdproc-uio9999" we are at 16 including the null.

We can do:

```
void tcmu_set_thread_name(const char *prefix, struct tcmu_device *dev)
{
        if ((strlen(tcmu_dev_get_uio_name(dev)) + strlen(prefix) > TCMU_THREAD_NAME_LEN - 1) ||
           ((asprintf(&pname, "cmdproc-%s", tcmu_dev_get_uio_name(dev)) == -1) {
            pthread_setname_np(pthread_self(), "cmdproc");
                pthread_setname_np(pthread_self(), "cmdproc");
       } else {
		pthread_setname_np(pthread_self(), pname);
 		free(pname);
     }
}
```

libtcmu.c
lxbsz: Yeah, it is. Checked it locally, the diff code is still in the stash.
Will push it again.

main.c
lxbsz: Make sense and done.

lxbsz:
@mikechristie 
Done, this in version added the tcmu_set_thread_name() and tcmu_get_thread_name() helpers.
Thanks.

file_optical.c
```diff
@@ -1319,6 +1321,8 @@ static void *fbo_async_format(void *arg)
 	struct fbo_state *state = tcmur_dev_get_private(dev);
 	uint8_t sense[SENSE_BUFFERSIZE];
 
+	tcmu_set_thread_name("fbo-async-format", NULL);
```
mikechristie: You can just name this "fbo-format" so it is "handler_name-command" like the fbo-sync-cache one.

libtcmu.c
```diff
@@ -835,6 +835,48 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *private)
...
+
+		len = TCMU_THREAD_NAME_LEN;
+	}
+
+	if (asprintf(&pname, "%s%s%s", prefix, uio ? "-" : "",  uio ? uio : "") == -1) {
+		tcmu_dev_err(dev, "no memory for pname\n");
+		return;
+	}
+
+	pname[len] = '\0';
```
mikechristie: If len=TCMU_THREAD_NAME_LEN will this put the null at one char past the max length. The max length is 16 including the terminating null, so you would want to put it at pname[15] right?

libtcmu_log.c
```diff
@@ -411,10 +412,15 @@ static int output_to_fd(int pri, const char *timestamp,
 	if (pid <= 0)
 		return -1;
 
+	if (tcmu_get_thread_name(&pname))
```
mikechristie: We can probably just drop the function and make this a lot more simple for now. Just do a:

char pname[TCMU_THREAD_NAME_LEN];
...
pthread_getname_np(pthread_self(), pname, TCMU_THREAD_NAME_LEN);

in this function.

lxbsz: sure.

lxbsz:
Resolved the conflicts.

libtcmu.c
```diff
@@ -835,6 +835,49 @@ void tcmu_dev_set_private(struct tcmu_device *dev, void *private)
...
+	}
+
+	if (asprintf(&pname, "%s%s%s", prefix, uio ? "-" : "",  uio ? uio : "") == -1) {
+		tcmu_dev_err(dev, "no memory for pname\n");
+		return;
+	}
+
+	pname[end] = '\0';
+
+	pthread_setname_np(pthread_self(), pname);
```
mikechristie: @lxbsz 

Sorry. I forgot about this patch. I was wondering why you never responded to my last comments, but I guess I never posted them :)

I attached a patch with some fixes for this function. We do not need to be so finely grained in the error message for when it is too long, so we can simplify the function. Also it adds a check for if pthread_setname_np fails.

If it's ok with you then I will just merge your patches as is, then merge my patch over it.


[xiubo-thread-name-simplify.txt](https://github.com/open-iscsi/tcmu-runner/files/4189636/xiubo-thread-name-simplify.txt)

lxbsz: @mikechristie ,

I almost forgot this too. I thought it was already merged :-)
This looks good to me.

Thanks

lxbsz: @mikechristie 
What's the status of this PR?
Thanks.

PR #599, opened 2019-10-31
glfs: check EOF for read

lxbsz:

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -657,6 +657,13 @@ static void glfs_async_cbk(glfs_fd_t *fd, ssize_t ret, void *data)
 		/* Read/write/flush failed */
 		switch (cookie->op) {
 		case TCMU_GLFS_READ:
+			/* ENOENT for READ operation means EOF,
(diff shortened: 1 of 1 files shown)


lxbsz:
@pkalever Please review. Thanks.

glfs.c
```diff
@@ -667,6 +672,7 @@ static void glfs_async_cbk(glfs_fd_t *fd, ssize_t ret, void *data)
 			break;
 		}
 	} else {
+out:
 		ret = TCMU_STS_OK;
```
mikechristie: Did you guys want me to merge this up? It seems ok. Maybe going by the coding style of the current code you would want to rename the goto to something like "success:" because you guys seem to use "out:" for the completion. Or maybe just do a

```
if (err == -ENOENT) {
      ret = TCMU_STS_OK;
      break;
}
```

In the first chunk instead of going the goto route.

It does not matter to me. Keep it like it is, or change it. I am ok with whatever you guys approve.

lxbsz: Yeah, this looks nicer, will fix it.

PR #602, opened 2019-12-05
Refactor use of size_t

javacruft:
Use uint64_t in preference to size_t to avoid size mismatches
on architectures that are not 64bit.  This generates compilation
errors on most modern gcc compiler version.

Signed-off-by: James Page <james.page@ubuntu.com>

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -86,7 +86,7 @@ void tcmur_cmd_complete(struct tcmu_device *dev, void *data, int rc)
 }
 
 static void tcmur_cmd_iovec_reset(struct tcmur_cmd *tcmur_cmd,
-				  size_t data_length)
+				  uint64_t data_length)
 {
 	tcmur_cmd->iovec->iov_base = tcmur_cmd->iov_base_copy;
 	tcmur_cmd->iovec->iov_len = data_length;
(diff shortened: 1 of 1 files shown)


tcmur_cmd_handler.c
```diff
@@ -86,7 +86,7 @@ void tcmur_cmd_complete(struct tcmu_device *dev, void *data, int rc)
 }
 
 static void tcmur_cmd_iovec_reset(struct tcmur_cmd *tcmur_cmd,
-				  size_t data_length)
+				  uint64_t data_length)
```
mikechristie: size_t is 32 bits on 32bit archs, right? If so, above we will use a uint64_t, but the iov_len field on a struct iovec is size_t, so will that cause a compilation error?

mikechristie:
Hey, thanks for the patch. One other question I had is are using tcmu-runner on a 32bit arch or just handling build issues?

We had another 32 bit issue here:

https://github.com/open-iscsi/tcmu-runner/pull/360

so I am just wondering if you also hit that.

javacruft:
I'm mainly just dealing with build issues in Ubuntu; the packaging had disabled errors relating to type mismatches due to size_t != uint64_t on 32 bit architectures.

I had not hit the issue in #360.

tcmur_cmd_handler.c
javacruft: Hmm - it does not create a compile error but I think that the code needs alignment around this fact - i.e size_t should be the basic assumption rather than uint64_t

PR #608, opened 2020-01-03
Fix i586 length error

gonzoleeman:
This fix is actually from @ddiss but I'm submitting the pull request since I tested the fix.

This allows tcmu-runner to compile on 32-bit i586 architectures.


gonzoleeman:
the merge tirds are from git, not me. That was just me syncing upstream to my branch. Most easily fixed by me deleting and recreating my clone, I suspect.

As for handle_format_unit(), strange that the compiler doesn't care, but I'm fine with adding that in my resubmission.

PR #610, opened 2020-01-07
extra script: Fix install_dep.sh

g199209:
In `CMakeLists.txt`:

```
find_library(LIBNL_LIB nl-3)
find_library(LIBNL_GENL_LIB nl-genl-3)
set(LIBNL_LIBS
  ${LIBNL_LIB}
  ${LIBNL_GENL_LIB}
  )
```

So we should install `libnl-genl-3-dev` package, which is not included in `install_dep.sh`.

### DIFF
--- a/extra/install_dep.sh
+++ b/extra/install_dep.sh
@@ -28,8 +28,8 @@ if [ y`uname`y = yLinuxy ]; then
 		$SUDO apt update
 		
 		# for generic
-		$SUDO apt install -y cmake make gcc libnl-3-dev libglib2.0-dev zlib1g kmod
-		$SUDO apt install -y libnl-3-dev libglib2.0-0 libkmod-dev libgoogle-perftools-dev
+		$SUDO apt install -y cmake make gcc zlib1g kmod
+		$SUDO apt install -y libnl-3-dev libnl-genl-3-dev libglib2.0-0 libkmod-dev libgoogle-perftools-dev
 		
 		# for glusterfs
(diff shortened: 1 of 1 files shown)


extra/install_dep.sh
```diff
@@ -28,8 +28,8 @@ if [ y`uname`y = yLinuxy ]; then
 		$SUDO apt update
 		
 		# for generic
-		$SUDO apt install -y cmake make gcc libnl-3-dev libglib2.0-dev zlib1g kmod
-		$SUDO apt install -y libnl-3-dev libglib2.0-0 libkmod-dev libgoogle-perftools-dev
+		$SUDO apt install -y cmake make gcc zlib1g kmod
+		$SUDO apt install -y libnl-3-dev libnl-genl-3-dev libglib2.0-0 libkmod-dev libgoogle-perftools-dev
```
mikechristie: Do you want to add this so it matches the CMakeList?

I had thought libnl-genl-3-dev was brought in automatically when you install libnl-3-dev.

g199209: `libnl-genl-3-dev` and `libnl-3-dev` are two different package in Debian, installing `libnl-3-dev` will NOT install `libnl-genl-3-dev` automatically.

I got a cmake error "LIBNL_GENL_LIB not found" after I run `install_dep.sh`, so I add this package to this script.

mikechristie:
Thanks.

Issue #613, opened 2020-03-11
Stack buffer overflow in tcmu_cdb_print_info


chrisccoulson:
tcmu_cdb_print_info has a fixed size 193 bytes scratch buffer on the stack, defined here:
```
#define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 1)
#define CDB_FIX_BYTES 64 /* 64 bytes for default */
#define CDB_FIX_SIZE CDB_TO_BUF_SIZE(CDB_FIX_BYTES)
void tcmu_cdb_print_info(struct tcmu_device *dev,
                         const struct tcmulib_cmd *cmd,
                         const char *info)
{
        int i, n, bytes;
        char fix[CDB_FIX_SIZE], *buf;

        buf = fix;
```
Further down, it uses sprintf to write some bytes in 2-digit hex format to buf, with each byte occupying 3 bytes in buf:
```
        for (i = 0, n = 0; i < bytes; i++) {
                n += sprintf(buf + n, "%x ", cmd->cdb[i]);
        }
```
To avoid overflowing the stack buffer, there is a check earlier in the function which performs a heap allocation if bytes is too large:
```
        if (bytes > CDB_FIX_SIZE) {
                buf = malloc(CDB_TO_BUF_SIZE(bytes));
                if (!buf) {
                        tcmu_dev_err(dev, "out of memory\n");
                        return;
                }
        }
```
But this check looks wrong. If bytes > 64, then the previously mentioned for loop will overflow the 193 byte stack buffer. I think it should be:
```
        if (bytes > CDB_FIX_BYTES) {
```
Correcting this isn't sufficient on its own though, as this check doesn't take in to account that an additional string (info) is written to buf and can also result in the stack buffer overflowing:
```
        if (info)
                n += sprintf(buf + n, "%s", info);
```
I would normally report this kind of issue privately, but it looks like triggering this code relies on changing the log level and AFAICT there isn't really a mechanism for an attacker to exploit this anyway.

mikechristie:
Thanks! Agree on both cases. I made a PR that should hopefully address all your issues here:

https://github.com/open-iscsi/tcmu-runner/pull/615

PR #615, opened 2020-03-13
Fix possible cdb printing overflow

mikechristie:
Reported by Chris Coulson (chrisccoulson) at:
https://github.com/open-iscsi/tcmu-runner/issues/613

tcmu_cdb_print_info has a fixed size 193 bytes scratch buffer on the stack,
defined here:

void tcmu_cdb_print_info(struct tcmu_device *dev,
                         const struct tcmulib_cmd *cmd,
                         const char *info)
{
        int i, n, bytes;
        char fix[CDB_FIX_SIZE], *buf;

        buf = fix;
Further down, it uses sprintf to write some bytes in 2-digit hex format to
buf, with each byte occupying 3 bytes in buf:

        for (i = 0, n = 0; i < bytes; i++) {
                n += sprintf(buf + n, "%x ", cmd->cdb[i]);
        }
To avoid overflowing the stack buffer, there is a check earlier in the
function which performs a heap allocation if bytes is too large:

        if (bytes > CDB_FIX_SIZE) {
                buf = malloc(CDB_TO_BUF_SIZE(bytes));
                if (!buf) {
                        tcmu_dev_err(dev, "out of memory\n");
                        return;
                }
        }
But this check looks wrong. If bytes > 64, then the previously mentioned for
loop will overflow the 193 byte stack buffer. I think it should be:

        if (bytes > CDB_FIX_BYTES) {
Correcting this isn't sufficient on its own though, as this check doesn't
take in to account that an additional string (info) is written to buf and
can also result in the stack buffer overflowing:

        if (info)
                n += sprintf(buf + n, "%s", info);

Signed-off-by: Mike Christie <mchristi@redhat.com>

### DIFF
--- a/api.c
+++ b/api.c
@@ -314,14 +314,14 @@ size_t tcmu_memcpy_from_iovec(
 	return copied;
 }
 
-#define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 1)
+#define CDB_TO_BUF_SIZE(bytes) ((bytes) * 3 + 2)
 #define CDB_FIX_BYTES 64 /* 64 bytes for default */
 #define CDB_FIX_SIZE CDB_TO_BUF_SIZE(CDB_FIX_BYTES)
 void tcmu_cdb_print_info(struct tcmu_device *dev,
 			 const struct tcmulib_cmd *cmd,
 			 const char *info)
 {
-	int i, n, bytes;
+	int i, n, bytes, info_len = 0;
 	char fix[CDB_FIX_SIZE], *buf;
 
 	buf = fix;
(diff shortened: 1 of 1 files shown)


chrisccoulson:
This fix looks ok, although I think there is still one minor issue - the size of the buffer in both cases (both the on-stack one and the heap one) are one byte too short, as the numbers don't take in to account the null terminating byte that sprintf writes.

mikechristie:
I thought that's covered with CDB_TO_BUF_SIZE with the +1 when it does "((bytes) * 3 + 1)"?

chrisccoulson:
Unless I'm missing something, the +1 in CDB_TO_BUF_SIZE gets taken up by a new line here: https://github.com/open-iscsi/tcmu-runner/blob/73602a8dbd1de41ca62974cc7c1b369706649cee/api.c#L348

mikechristie:
Doh, you are right.

mikechristie:
@chrisccoulson, it should be fixed now.

mikechristie:
I'm assuming I got it right this time, and merging it.

Just let me know if there are still issues and I will do another PR.

chrisccoulson:
> I'm assuming I got it right this time, and merging it.
> 
> Just let me know if there are still issues and I will do another PR.

Oh, sorry, yes - it looks fine now. Thanks for fixing this!

PR #619, opened 2020-03-31
glfs: allow to load the dev when its size disagree with the backend size

lxbsz:
The old code will just fail the glfs device creating, but on the
resize operation, say the backend file expand succees but the if
the config change on tcmu-runner node fails. In this case, after
the node rebooted and retoring it will always fails.

Here will allow the sizes disagree and just apply the new size to
kernel.

### DIFF
--- a/glfs.c
+++ b/glfs.c
@@ -563,23 +563,19 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
 		if (round_down(st.st_size, block_size) == dev_size)
 			goto out;
 
-		if (!reopen) {
-			ret = -EINVAL;
-			goto close;
-		}
+		if (reopen)
+			goto out;
 
-		/*
-		 * If we are here this should be in reopen path,
-		 * then we should also update the device size in
-		 * kernel.
-		 */
-		tcmu_dev_info(dev,
-			      "device size and backing size disagree:device %lld backing %lld\n",
-			      dev_size, (long long) st.st_size);
+		tcmu_dev_warn(dev,
+			      "device size (%lld) and backing file size (%lld) not matching, updating it to kernel\n",
+			      (long long)dev_size, (long long) st.st_size);
 
+		/* Update the device size in kernel. */
 		ret = tcmur_dev_update_size(dev, st.st_size);
 		if (ret)
 			goto close;
+
+		tcmu_dev_info(dev, "loaded with size (%lld)\n", (long long) st.st_size);
 	}
 
 out:
@@ -772,10 +768,15 @@ static int tcmu_glfs_reconfig(struct tcmu_device *dev,
 			/* Let the targetcli command return success */
 			ret = 0;
 		} else if (st.st_size != cfg->data.dev_size) {
-			tcmu_dev_err(dev,
-				     "device size and backing size disagree: device %"PRId64" backing %lld\n",
-				     cfg->data.dev_size, (long long) st.st_size);
-			ret = -EINVAL;
+			/*
+			 * Currently we cannot update the size to kernel here,
+			 * because it will be overrided by kernel with the old
+			 * value after it gets the genl reply.
+			 */
+			tcmu_dev_warn(dev,
+				      "device size (%lld) and backing file size (%lld) not matching, and ignoring it\n",
+				      (long long)cfg->data.dev_size, (long long) st.st_size);
+			return -EINVAL;
 		}
 		return ret;
 	case TCMULIB_CFG_DEV_CFGSTR:


lxbsz:
@pkalever Please review, thanks.

glfs.c
```diff
@@ -563,16 +563,10 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
 		if (round_down(st.st_size, block_size) == dev_size)
 			goto out;
 
-		if (!reopen) {
-			ret = -EINVAL;
-			goto close;
-		}
+		if (!reopen)
+		    tcmu_dev_warn(dev, "unmatched sizes in none resize path!\n");
```
mikechristie, line 567: I think you wanted "non resize path". Maybe a better warning would be:

Device size mismatch during initialization.

Or, you probably do not need the extra log statement. You could just modify the one below.

lxbsz: Yeah, will fix the warning.
This case is rare, normally it shouldn't be here when initializing the device, so give a warning will be more friendly.
Thanks

pkalever:
@lxbsz I have tested this and looks like we still need to do some changes in `tcmu_glfs_reconfig`

```
[root@server1 tcmu-runner]# targetctl restore                                                                      
Storage Object user/block: Cannot set attribute dev_config: [Errno 95] Operation not supported, skipped            
Storage Object user/block: Cannot set attribute emulate_write_cache: [Errno 95] Operation not supported, skipped   
Storage Object user/block: Cannot set attribute dev_size: [Errno 22] Invalid argument, skipped                     
[root@server1 tcmu-runner]#                                                                                        

```

```
2020-04-20 09:27:02.783 20301:tcmu-runner [WARN] tcmu_glfs_open:567 glfs/block: Device size mismatch during initialization.
2020-04-20 09:27:02.783 20301:tcmu-runner [INFO] tcmu_glfs_open:572 glfs/block: device size and backing size disagree:device 1048576 backing 2097152
2020-04-20 09:27:02.791 20301:tcmu-runner [ERROR] tcmu_glfs_reconfig:771 glfs/block: device size and backing size disagree: device 1048576 backing 2097152

```


Thanks!

lxbsz:
In tcmu-runner/glfs handler though the updating the dev_size to kernel will success, the kernel will override the dev_size again with the old size. This is because the updating from tcmu-runner is just before the TCMU_CMD_RECONFIG_DEVICE genl event gets its reply, and after reply received it will update the dev_size with the old size.

This PR will try to allow the device to be loaded when it detect the sizes are mismatched. And then we could allow the gluster-block tool to adjust the size later.

lxbsz:
@mikechristie 

If we want to make the above to work smoothly, is there objection to the following change in kernel space ?

```
git diff drivers/target/target_core_user.c
diff --git a/drivers/target/target_core_user.c b/drivers/target/target_core_user.c
index 0b9dfa6..a53b011 100644
--- a/drivers/target/target_core_user.c
+++ b/drivers/target/target_core_user.c
@@ -2417,22 +2417,25 @@ static ssize_t tcmu_dev_size_store(struct config_item *item, const char *page,
        struct se_dev_attrib *da = container_of(to_config_group(item),
                                                struct se_dev_attrib, da_group);
        struct tcmu_dev *udev = TCMU_DEV(da->da_dev);
-       u64 val;
+       u64 val, old;
        int ret;
 
        ret = kstrtou64(page, 0, &val);
        if (ret < 0)
                return ret;
 
+       old = udev->dev_size;
+       udev->dev_size = val;
+
        /* Check if device has been configured before */
        if (target_dev_configured(&udev->se_dev)) {
                ret = tcmu_send_dev_size_event(udev, val);
                if (ret) {
                        pr_err("Unable to reconfigure device\n");
+                       udev->dev_size = old;
                        return ret;
                }
        }
-       udev->dev_size = val;
        return count;
 }
 CONFIGFS_ATTR(tcmu_, dev_size);
```

glfs.c
```diff
@@ -563,20 +563,24 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
 		if (round_down(st.st_size, block_size) == dev_size)
 			goto out;
 
-		if (!reopen) {
-			ret = -EINVAL;
-			goto close;
-		}
+		if (!reopen)
+			tcmu_dev_warn(dev, "Device size mismatch during initialization.\n");
```
pkalever, line 567: may start with the small case just like in other places:
"Device size..." -> "device size..."

glfs.c
```diff
@@ -772,10 +776,16 @@ static int tcmu_glfs_reconfig(struct tcmu_device *dev,
...
+			/*
+			 * Currently we cannot update the size to kernel during
+			 * initialization, because it will be overrided by kernel
+			 * after it get the genl reply. Here will just allow it
+			 * to load the device and we will run gluster-block tool
+			 * to update the size.
+			 */
+			tcmu_dev_info(dev,
+				      "device size and backing size disagree:device %lld backing %lld\n",
+				      (long long)cfg->data.dev_size, (long long) st.st_size);
```
pkalever, line 788: shouldn't this be warning? something like below would be nice:
```
tcmu_dev_warn(dev,
              "device size (%lld) and backing file size (%ldd) not matching\n"
              (long long)cfg->data.dev_size, (long long) st.st_size);
```

glfs.c
lxbsz: Checked all the other in tcmu-runner and glfs, they are all start with upper case. And only some start with function name will the be small case.

glfs.c
lxbsz: Make sense and will fix it.

mikechristie:
Hey @lxbsz 

I'm not sure I understood the issue. Is one thread updating the dev_size via the configfs info file, and then another thread is updating the dev_size via sysfs?

I can see a race like that happening in the kernel code. However, I had thought for targetcli though the initial dev creation is serialized. It does the creation, when that is done then that code to write the sysfs attrs runs.

Also, we have this same bug with other attrs/settings in tcmu right?

lxbsz:
> Hey @lxbsz
> 
> I'm not sure I understood the issue. Is one thread updating the dev_size via the configfs info file, and then another thread is updating the dev_size via sysfs?

This use case here, such as when running the` targetctl restore` command, it will do:

1, kernel remove/add the tcmu devices back and will send a CFG DEV SIZE netlink request to tcmu-runner.
2, tcmu-runner/glfs get the netlink request and if glfs detects the sizes are mismatched and it will update the new size to kernel.
3, tcmu-runner will reply the netlink request to kernel .
4, after kernel get the reply it will update the dev_size with the old size, which will override the new size we updated in Step2.

```
        /* Check if device has been configured before */
        if (target_dev_configured(&udev->se_dev)) {
                ret = tcmu_send_dev_size_event(udev, val);  ===> netlink request , tcmu-runner update the dev_size if mismatched and get the reply
                if (ret) {
                        pr_err("Unable to reconfigure device\n");
                        return ret;
                }
        }
       udev->dev_size = val;            ===> update the dev_size with the old size value
```

> 
> I can see a race like that happening in the kernel code. However, I had thought for targetcli though the initial dev creation is serialized. It does the creation, when that is done then that code to write the sysfs attrs runs.
> 
> Also, we have this same bug with other attrs/settings in tcmu right?

Currently for the other attrs/settings should be okay, but potentially have the same issue someday in future.

lxbsz:
@mikechristie

This PR will just allow it to load the device when the sizes disagree. The old code will fail it in this case, which makes it's not possible to recover from it.

mikechristie:
Maybe, I am starting to understand :)

Is the issue that in tcmu_glsfs_reconfig you want to be able to call tcmur_dev_update_size()? If you call that function from the reconfig callout, the kernel will set the dev_size set to N. But tcmu_glfs_reconfig will return and tcmu-runner returns a NL response. tcmu_send_dev_size_event then sets to the dev_size to M.

lxbsz:
@mikechristie

This PR will just allow it to load the device when the sizes disagree. The old code will fail it in this case, which makes it's not possible to recover from it. 

> Maybe, I am starting to understand :)
> 
> Is the issue that in tcmu_glsfs_reconfig you want to be able to call tcmur_dev_update_size()? If you call that function from the reconfig callout, the kernel will set the dev_size set to N. But tcmu_glfs_reconfig will return and tcmu-runner returns a NL response. tcmu_send_dev_size_event then sets to the dev_size to M.

Yeah, right.

I was thinking in tcmu-runner just to wait a few seconds in a separate thread after it returns the NL response, and then update the new size to kernel, but it seems buggy because we didn't know whether will the NL response will delay and how long will it stuck. And it will be race with others too.

lxbsz:
@mikechristie 
Could merge this PR ? This will as a workaround and just allow the tcmu-device to be created/loaded when it see the sizes mismatched. And this will also allow us to use the gluster-block resize command to do the resize again to fix it.

glfs.c
```diff
@@ -563,20 +563,24 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
...
 
 		/*
-		 * If we are here this should be in reopen path,
-		 * then we should also update the device size in
-		 * kernel.
+		 * Currently we cannot update the size to kernel during
+		 * initialization, because it will be overrided by kernel
+		 * after it get the genl reply. Here will just allow it
+		 * to load the device and we will run gluster-block tool
+		 * to update the size.
```
mikechristie, line 578: I'm not sure this is correct.

In targetcli/rtslib we do:

1. make the device in configfs (do the mkdirs in /sys/kernel/config/core to setup user_N and whatever the device name is under that dir, etc).
2. Set dev_config, dev_size via the configfs control/info file.
3. enable the device (write 1 to the enable file in the device dir). This calls into the tcmu_configure_device callout for tcmu and this sends the TCMU_CMD_ADDED_DEVICE event.
4. runner handles that event by creating the device in the daemon and calling the handler->open callout.

Here if you set the tcmur_dev_update_size it will call into the kernel via the configfs control/info file and set dev_size.

5. runner returns NL response to tcmu.
6. tcmu returns response to lio core which returns to rtslib.
7. targetcli/rtslib then loops over the device attrs and writes to them assuming they are not the same as setup in the info/control file, which triggers the reconfig callout below for dev_size. In that callout you cannot update the size because you would hit the bug where it gets reset like you describe.

For the reconfig you would want to ignore the size if its not correct like you did below. But the dev_size should be the correct/updated value from step 4.

mikechristie: > For the reconfig you would want to ignore the size if its not correct like you did below. But the dev_size should be the correct/updated value from step 4.

I mean the dev_size in tcmu-runner and the kernel will have been updated from step 4. The value being passed into the reconfig callout will be the old value from the /etc/target/saveconfig.json attributes array or wherever you are getting the storied attributes from.

lxbsz: I try to catch some logs when restarting the tcmu-runner and gluster-blockd daemons at the same time:

```
tcmu-runner [INFO] handle_netlink:229: TCMU_CMD_REMOVED_DEVICE
tcmu-runner [INFO] handle_netlink:222: TCMU_CMD_ADDED_DEVICE, cfgstring = [tcm-user/0/block0/glfs/repvol@localhost/block-store/0f9483be-31a1-4ecb-8e3a-6137bedc6217]
tcmu-runner [INFO] tcmu_glfs_open:559 glfs/block0: device size (4294967296) and backing file size (5368709120)
tcmu-runner [INFO] handle_netlink:235: TCMU_CMD_RECONFIG_DEVICE
tcmu-runner [INFO] reconfig_device:163: TCMU_ATTR_DEV_CFG
tcmu-runner [INFO] handle_netlink:235: TCMU_CMD_RECONFIG_DEVICE
tcmu-runner [INFO] reconfig_device:172: TCMULIB_CFG_WRITE_CACHE
tcmu-runner [INFO] handle_netlink:235: TCMU_CMD_RECONFIG_DEVICE
tcmu-runner [INFO] reconfig_device:168: TCMULIB_CFG_DEV_SIZE, size = 4294967296

```
The dev_size could be set after the enable happens, see the last logs. And in the last step(my logs) above when setting the dev_size, it will call:

```
git diff drivers/target/target_core_user.c
diff --git a/drivers/target/target_core_user.c b/drivers/target/target_core_user.c
index 0b9dfa6..a53b011 100644
--- a/drivers/target/target_core_user.c
+++ b/drivers/target/target_core_user.c
@@ -2417,22 +2417,25 @@ static ssize_t tcmu_dev_size_store(struct config_item *item, const char *page,
        struct se_dev_attrib *da = container_of(to_config_group(item),
                                                struct se_dev_attrib, da_group);
        struct tcmu_dev *udev = TCMU_DEV(da->da_dev);
-       u64 val;
+       u64 val, old;
        int ret;
 
        ret = kstrtou64(page, 0, &val);
        if (ret < 0)
                return ret;
 
+       old = udev->dev_size;
+       udev->dev_size = val;
+
        /* Check if device has been configured before */
        if (target_dev_configured(&udev->se_dev)) {
                ret = tcmu_send_dev_size_event(udev, val);
                if (ret) {
                        pr_err("Unable to reconfigure device\n");
+                       udev->dev_size = old;
                        return ret;
                }
        }
-       udev->dev_size = val;
        return count;
 }
 CONFIGFS_ATTR(tcmu_, dev_size);
```
 As the code, the kernel/tcmu will send the dev_size reconfig NL request ---> tcmu-runner get the request and if we call `tcmur_dev_update_size(5G),` then the kernel/tcmu will update the udev->dev_size ---> tcmu-runner reply the NL request --> kernel/tcmu continue and will update` udev->dev_size = val;`, which the` val` here is the old size, 4G.

mikechristie: @lxbsz 

There are 3 cases we are discussing right?

1. reconfig.

I agree it can set/reset the dev size after addition. As I said in the last comment that happens in step #7. We get the resize events because targetcli/rtslib loops over the attributes after the enablement.

That part of the patch is fine.

2. initial open (reopen=false):

I am saying that when the device is first added and runner calls handler->open() you can set the dev_size to the correct value and it will not get reset like in the reconfig case. 

After the initial open then we can hit #1, like I described in my last comment, but if you ignore like you are doing in your patch then that is fine.

3. reopen=true

We have a race where tcmu_glfs_reconfig could call tcmu_notify_conn_lost and in another thread glfs->open is called with reopen=true. If that thread calls tcmur_dev_update_size before the reconfig thread has returned the NL response then we hit issue 1.

So you could do the opposite in the open() call and we will use the correct size. See below for an example of what I mean.

You would need to still run targetlci or whatever tool to update the stored attribtutes in /etc/target/saveconfig.json.
```
@@ -558,20 +558,20 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
                if (round_down(st.st_size, block_size) == dev_size)
                        goto out;

-               if (!reopen) {
-                       ret = -EINVAL;
-                       goto close;
-               }
+               tcmu_dev_warn(dev,
+                             "device size (%lld) and backing file size (%lld) not matching\n",
+                             (long long)dev_size, (long long) st.st_size);

                /*
-                * If we are here this should be in reopen path,
-                * then we should also update the device size in
-                * kernel.
+                * If reopen is processed while a reconfig we can race
+                * and the reconfig NL response may reset the size.
                 */
-               tcmu_dev_info(dev,
-                             "device size and backing size disagree:device %lld backing %lld\n",
-                             dev_size, (long long) st.st_size);
+               if (reopen)
+                       goto out;

+               tcmu_dev_warn(dev, "device size mismatch during initialization. Updating the kernel to  %lld. Please run \n",
+                            (long long) st.st_size);
+               /* Update the device size in kernel. */
                ret = tcmur_dev_update_size(dev, st.st_size);
                if (ret)
                        goto close;
@@ -762,10 +762,16 @@ static int tcmu_glfs_reconfig(struct tcmu_device *dev,
                        /* Let the targetcli command return success */
                        ret = 0;
                } else if (st.st_size != cfg->data.dev_size) {
-                       tcmu_dev_err(dev,
-                                    "device size and backing size disagree: device %"PRId64" backing %lld\n",
-                                    cfg->data.dev_size, (long long) st.st_size);
-                       ret = -EINVAL;
+                       /*
+                        * Currently we cannot update the size to kernel during
+                        * initialization, because it will be overrided by kernel
+                        * after it get the genl reply. Here will just allow it
+                        * to load the device and we will run gluster-block tool
+                        * to update the size.
+                        */
+                       tcmu_dev_warn(dev,
+                                     "device size (%lld) and backing file size (%lld) does not match\n",
+                                     (long long)cfg->data.dev_size, (long long) st.st_size);
                }
                return ret;
        case TCMULIB_CFG_DEV_CFGSTR:  
```

mikechristie: >                 } else if (st.st_size != cfg->data.dev_size) {


One correction here. You need to add a

`                           ret = -EINVAL;
`


> -                       tcmu_dev_err(dev,
> -                                    "device size and backing size disagree: device %"PRId64" backing %lld\n",
> -                                    cfg->data.dev_size, (long long) st.st_size);
> -                       ret = -EINVAL;
> +                       /*
> +                        * Currently we cannot update the size to kernel during
> +                        * initialization, because it will be overrided by kernel
> +                        * after it get the genl reply. Here will just allow it
> +                        * to load the device and we will run gluster-block tool
> +                        * to update the size.
> +                        */
> +                       tcmu_dev_warn(dev,
> +                                     "device size (%lld) and backing file size (%lld) does not match\n",
> +                                     (long long)cfg->data.dev_size, (long long) st.st_size);
>                 }
>                 return ret;
>         case TCMULIB_CFG_DEV_CFGSTR:  
> ```

lxbsz: > @lxbsz
> 
> There are 3 cases we are discussing right?
> 

Right.

> 1. reconfig.
> 
> I agree it can set/reset the dev size after addition. As I said in the last comment that happens in step # 7. We get the resize events because targetcli/rtslib loops over the attributes after the enablement.
> 
> That part of the patch is fine.
> 
> 1. initial open (reopen=false):
> 
> I am saying that when the device is first added and runner calls handler->open() you can set the dev_size to the correct value and it will not get reset like in the reconfig case.
> 
> After the initial open then we can hit # 1, like I described in my last comment, but if you ignore the bad size in the reconfig like you are doing in your patch then that is fine.
> 
> 1. reopen=true
> 
> We have a race where tcmu_glfs_reconfig could call tcmu_notify_conn_lost and in another thread glfs->open is called with reopen=true. If that thread calls tcmur_dev_update_size before the reconfig thread has returned the NL response then we hit issue 1.
> 
> So you could do the opposite in the open() call and we will use the correct size. See below for an example of what I mean.
> 
> You would need to still run targetlci or whatever tool to update the stored attribtutes in /etc/target/saveconfig.json.

Yeah, agree.

Updated it now. Thanks Mike.

glfs.c
```diff
@@ -772,10 +766,17 @@ static int tcmu_glfs_reconfig(struct tcmu_device *dev,
...
 			ret = 0;
 		} else if (st.st_size != cfg->data.dev_size) {
+			/*
+			 * Currently we cannot update the size to kernel during
+			 * initialization, because it will be overrided by kernel
+			 * after it get the genl reply. Here will just allow it
+			 * to load the device and we will run gluster-block tool
+			 * to update the size.
+			 */
 			tcmu_dev_err(dev,
```
pkalever, line 776: @lxbsz , as I mentioned before, will warn level log is not enough?

Also please fix the code comment above.

lxbsz: I just following the others, which is also as ERRORS:
```

Storage Object user/block0: Cannot set attribute dev_config: [Errno 95] Operation not supported, skipped
Storage Object user/block0: Cannot set attribute emulate_write_cache: [Errno 95] Operation not supported, skipped
Storage Object user/block0: Cannot set attribute dev_size: [Errno 22] Invalid argument, skipped
```

pkalever: @lxbsz I get you:
but look at the below logs:

```
2020-04-27 02:50:58.501 4589 [CRIT] main:1187: Starting...
2020-04-27 02:50:58.616 4589 [WARN] tcmu_glfs_open:558 glfs/block1: device size (1073741824) and backing file size (2147483648) not matching
2020-04-27 02:50:58.623 4589 [ERROR] tcmu_glfs_reconfig:714 glfs/block1: device size (1073741824) and backing file size (2147483648) not matching
```


They don't give any kind of impression that the device is loaded with the size same as backend anywhere. Rather they say we errored already.

Maybe some logs mentioning that we are ignoring the mismatch and continuing to load the storage with size same as backend file would be helpful IMHO.

Thanks!

lxbsz: Yeah, sounds good.

lxbsz: @pkalever  Updated it. Thanks

pkalever:
@lxbsz the previous change looks good to me.
I was also pointing, like can add something similar to:

```
diff --git a/glfs.c b/glfs.c                                                       
index 9274daf..588a2fa 100644                                                      
--- a/glfs.c                                                                       
+++ b/glfs.c                                                                       
@@ -574,6 +574,9 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
                ret = tcmur_dev_update_size(dev, st.st_size);                      
                if (ret)                                                           
                        goto close;                                                
+                                                                                  
+               tcmu_dev_info(dev, "device: %s loaded with size: (%lld)",          
+                  dev->tcm_dev_name, (long long) st.st_size);                     

```

Thanks!

lxbsz:
> @lxbsz the previous change looks good to me.
> I was also pointing, like can add something similar to:
> 
> ```
> diff --git a/glfs.c b/glfs.c                                                       
> index 9274daf..588a2fa 100644                                                      
> --- a/glfs.c                                                                       
> +++ b/glfs.c                                                                       
> @@ -574,6 +574,9 @@ static int tcmu_glfs_open(struct tcmu_device *dev, bool reopen)
>                 ret = tcmur_dev_update_size(dev, st.st_size);                      
>                 if (ret)                                                           
>                         goto close;                                                
> +                                                                                  
> +               tcmu_dev_info(dev, "device: (%s) loaded with size: (%lld)",          
> +                  dev->tcm_dev_name, (long long) st.st_size);                     
> ```
> 
> Thanks!

```
16400 2020-04-27 03:37:07.164 28318:tcmu-runner [CRIT] main:1269: Starting...
16401 2020-04-27 03:37:07.177 28318:tcmu-runner [ERROR] open_handlers:160: Could not open handler at /usr/lib64/tcmu-runner/handler_rbd.so:       /usr/lib64/tcmu-runner/handler_rbd.so: undefined symbol: rados_service_register
16402 2020-04-27 03:37:18.344 28318:tcmu-runner [WARN] tcmu_glfs_open:571 glfs/block0: device size (16106127360) and backing file size (171      79869184) not matching, and updating it to kernel
16403 2020-04-27 03:37:18.345 28318:tcmu-runner [INFO] tcmu_glfs_open:578 glfs/block0: loaded with size (17179869184)
16404 2020-04-27 03:37:18.353 28318:tcmu-runner [WARN] tcmu_glfs_reconfig:778 glfs/block0: device size (16106127360) and backing file size       (17179869184) not matching, and ingoring it 
```

pkalever:
@lxbsz tested, looks good to me. But there is a typo:

```
2020-04-27 04:08:46.677 7156 [WARN] tcmu_glfs_open:558 glfs/block1: device size (1073741824) and backing file size (2147483648) not matching, and updating it to kernel
2020-04-27 04:08:46.677 7156 [INFO] tcmu_glfs_open:565 glfs/block1: loaded with size (2147483648)
2020-04-27 04:08:46.684 7156 [WARN] tcmu_glfs_reconfig:714 glfs/block1: device size (1073741824) and backing file size (2147483648) not matching, and ingoring it
```


Log 1. "... not matching, and updating it to kernel" -> "... not matching, updating it to kernel" or "... not matching, hence updating it to kernel"
Log 2. "... and ingoring it" -> "... and ignoring it"


Thanks!

PR #626, opened 2020-05-24
Misc fixes and support to report error stats

mikechristie:
The following patchset fixes several bugs then adds support to track and report errors to handler specific tools.

Note:

I'm not working on tcmu-runner anymore. I'm just posting this with the hope it's useful to the next maintainer. You can give me review comments, but I will not be updating the patchset.


target.c
```diff
@@ -329,8 +330,13 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
 		ret = -ENOMEM;
 		goto done;
 	}
+
+	pthread_attr_init(&attr);
+	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
```
lxbsz, line 335: The `attr` here is defined not used, we should `s/NULL/attr/g` in `pthread_create()`.

target.c
```diff
@@ -329,8 +330,13 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
...
+
 	ret = pthread_create(&tpg->recovery_thread, NULL,
```
lxbsz, line 337: Since the `tpg->recovery_thread` member is not used in any other place after, we can get rid of it by using a local tmp parameter here.

file_optical.c
```diff
@@ -54,6 +54,7 @@
 #include "tcmu-runner.h"
 #include "tcmur_device.h"
 #include "libtcmu.h"
+#include "libtcmu_common.h"
```
lxbsz, line 57: To include the headers wherever needed is always a good idea.

But for this the `libtcmu_common.h` has been already included by `tcmu-runner.h` and `libtcmu.h`, this fixing here actually will do nothing. I think this `undefined symbol` error should be caused you were using the old version of `handler_file_optical.so` ?

alua.c
```diff
@@ -776,6 +776,25 @@ int alua_check_state(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		 * state gets updated.
+		 */
+		list_head_init(&group_list);
+		if (tcmu_get_alua_grps(dev, &group_list))
+			return TCMU_STS_HW_ERR;
+		tcmu_release_alua_grps(&group_list);
+
+		tcmu_update_dev_lock_state(dev);
+	}
+
```
lxbsz, line 787: Will this happen ?

In `dev_added()` it will always do
```
1028         list_head_init(&group_list);        
1029         tcmu_get_alua_grps(dev, &group_list);
1030         tcmu_release_alua_grps(&group_list);
```

ahead of 

```
1040         ret = pthread_create(&rdev->cmdproc_thread, NULL, tcmur_cmdproc_thread,
1041                              dev);
1042         if (ret) {            
1043                 ret = -ret;   
1044                 goto cleanup_lock_cond;             
1045         }
```

Before the `tcmur_cmdproc_thread()` thread run, no any IO command could be received.

tcmur_device.h
```diff
@@ -22,6 +22,7 @@
 #define TCMUR_DEV_FLAG_IS_OPEN		(1 << 2)
 #define TCMUR_DEV_FLAG_STOPPING		(1 << 3)
 #define TCMUR_DEV_FLAG_STOPPED		(1 << 4)
+#define TCMUR_DEV_FLAG_TRY_REAP		(1 << 5)
```
lxbsz, line 25: This flag is not used any where in this PR.

tcmur_cmd_handler.c
```diff
@@ -26,6 +26,7 @@
 #include "tcmu-runner.h"
 #include "tcmu_runner_priv.h"
 #include "tcmur_cmd_handler.h"
+#include "tcmur_work.h"
```
lxbsz, line 29: This should be by mistake in this commit ?

tcmur_device.h
```diff
@@ -49,7 +50,6 @@ struct tcmur_device {
 	uint32_t flags;
 	uint8_t failover_type;
 
-	pthread_t recovery_thread;
```
lxbsz, line 51: For this we may fix it in a separate commit.

tcmur_work.c
```diff
@@ -0,0 +1,119 @@
...
+		goto unlock;
+	}
+
+	pthread_attr_init(&attr);
+	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+
+	work->work_fn = work_fn;
+	work->data = data;
+
+	ret = pthread_create(&work->thread, NULL, tcmur_work_fn, work);
```
lxbsz, line 99: And same here, the `attr` is defined but not used.

alua.c
```diff
@@ -542,21 +543,19 @@ bool lock_is_required(struct tcmu_device *dev)
...
 
-static void *alua_lock_thread_fn(void *arg)
+static void alua_lock_work_fn(void *arg)
 {
 	struct tcmu_device *dev = arg;
 
 	tcmu_set_thread_name("alua-lock", dev);
 
 	/* TODO: set UA based on bgly's patches */
 	tcmu_acquire_dev_lock(dev, -1);
```
lxbsz, line 553: Maybe we could check the rdev's `TCMUR_DEV_FLAG_STOPPING` state just after the `aio_wait_for_empty_queue()` which may take a long time, and then just skip acquiring the lock and returned.

lxbsz:
Closing this PR now, and I have pushed it in https://github.com/open-iscsi/tcmu-runner/pull/630 with some fixings.
Thank you Mike for your work on this.

PR #630, opened 2020-07-21
Misc fixings and add work thread helper support

lxbsz:
> The following patchset fixes several bugs then adds support to track and report errors to handler specific tools.
> 
> Note:
> 
> I'm not working on tcmu-runner anymore. I'm just posting this with the hope it's useful to the next maintainer. You can give me review comments, but I will not be updating the patchset.

This PR is based Mike's previous https://github.com/open-iscsi/tcmu-runner/pull/626 and with some fixings. 
Thank you @mikechristie  for your work for this.

### DIFF
--- a/main.c
+++ b/main.c
@@ -48,6 +48,7 @@
 #include "version.h"
 #include "libtcmu_config.h"
 #include "libtcmu_log.h"
+#include "tcmur_work.h"
 
 #define TCMU_LOCK_FILE   "/run/tcmu.lock"
 
@@ -600,7 +601,7 @@ static void tcmur_stop_device(void *arg)
 	 * The lock thread can fire off the recovery thread, so make sure
 	 * it is done first.
 	 */
-	tcmu_cancel_lock_thread(dev);
+	tcmur_flush_work(rdev->event_work);
 	tcmu_cancel_recovery(dev);
 
 	tcmu_release_dev_lock(dev);
@@ -1031,9 +1032,9 @@ static int dev_added(struct tcmu_device *dev)
 
 	rdev->flags |= TCMUR_DEV_FLAG_IS_OPEN;
 
-	ret = pthread_cond_init(&rdev->lock_cond, NULL);
-	if (ret) {
-		ret = -ret;
+	rdev->event_work = tcmur_create_work();
+	if (!rdev->event_work) {
+		ret = -ENOMEM;
 		goto close_dev;
 	}
 
@@ -1041,13 +1042,13 @@ static int dev_added(struct tcmu_device *dev)
 			     dev);
 	if (ret) {
 		ret = -ret;
-		goto cleanup_lock_cond;
+		goto cleanup_event_work;
 	}
 
 	return 0;
 
-cleanup_lock_cond:
-	pthread_cond_destroy(&rdev->lock_cond);
+cleanup_event_work:
+	tcmur_destroy_work(rdev->event_work);
 close_dev:
 	rhandler->close(dev);
 cleanup_aio_tracking:
@@ -1094,9 +1095,7 @@ static void dev_removed(struct tcmu_device *dev)
 	cleanup_io_work_queue(dev, false);
 	cleanup_aio_tracking(rdev);
 
-	ret = pthread_cond_destroy(&rdev->lock_cond);
-	if (ret != 0)
-		tcmu_err("could not cleanup lock cond %d\n", ret);
+	tcmur_destroy_work(rdev->event_work);
 
 	ret = pthread_mutex_destroy(&rdev->state_lock);
 	if (ret != 0)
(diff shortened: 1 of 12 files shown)


main.c
```diff
@@ -723,6 +725,12 @@ static void check_for_timed_out_cmds(struct tcmu_device *dev)
 		}
 
 		tcmur_cmd->timed_out = true;
+		/*
+		 * These time outs are only currently used for diagnostic
+		 * purpposes right now, so we do not want to escalate the
```
dillaman, line 730:
```suggestion
		 * purposes right now, so we do not want to escalate the
```

target.c
```diff
@@ -329,8 +330,13 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
 		ret = -ENOMEM;
 		goto done;
 	}
-	ret = pthread_create(&tpg->recovery_thread, NULL,
+
+	pthread_attr_init(&attr);
+	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+
+	ret = pthread_create(&thread_id, &attr,
```
dillaman, line 337: If this tread is just running in the background, how is it tracked for joining before the process exits? Why not re-use the new `tcmur_work` thread wrapper and add the work struct to `tgt_port_grp`?

tcmur_device.c
```diff
@@ -489,3 +529,14 @@ void *tcmur_dev_get_private(struct tcmu_device *dev)
 
 	return rdev->hm_private;
 }
+
+void tcmu_notify_cmd_timed_out(struct tcmu_device *dev, bool can_access_dev)
+{
+	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
+
+	rdev->cmd_timed_out_cnt++;
```
dillaman, line 537: Nit: is someone holding the `state_lock` lock on `tcmur_device`?

main.c
lxbsz: Sure, will fix it.

tcmur_work.c
```diff
@@ -0,0 +1,119 @@
...
+	 * itself. Just ignore.
+	 */
+	if (pthread_self() == work->thread)
+		return;
+
+	/*
+	 * Some handlers will crash if we do a cancel so we just wait.
+	 */
+	tcmu_dbg("waiting for %lu to complete\n", work->thread);
+	pthread_cond_wait(&work->cond, &work->lock);
```
lxbsz, line 59: Commit myself:

For the work here we need one refcnt, just think if there has 2 threads queued, and when destroying and flushing the work, the flush could only make sure the first thread is done, the second may still running.

target.c
lxbsz: Yeah, good catch. In linux the detached thread will be terminated when the main process is exiting.  But there still have one problem, if the main process is exiting and it will release all the resources before it terminates all the detached threads, which may still accessing them.

I will convert to the tcmur_work thread wrapper, but the tcmur_work thread wrapper has one bug, I will fix it at the same time.

lxbsz:
@dillaman 
Updated it, for this version I remove the `work->state` and added one `work->refcnt`, with this we can run and queue more than one threads, the old code can only run and queue one thread.

tcmur_work.c
```diff
@@ -0,0 +1,105 @@
...
+	pthread_attr_t attr;
+	int ret;
+
+	pthread_attr_init(&attr);
+	pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+
+	pthread_mutex_lock(&work->lock);
+	work->work_fn = work_fn;
+	work->data = data;
+	ret = pthread_create(&work->thread, &attr, tcmur_work_fn, work);
```
dillaman, line 90: This would stomp on the `work->thread` of a previously running thread, right?  That makes `__tcmur_flush_work` check of the thread suspicious.

tcmur_device.c
```diff
@@ -489,3 +542,20 @@ void *tcmur_dev_get_private(struct tcmu_device *dev)
 
 	return rdev->hm_private;
 }
+
+void tcmu_notify_cmd_timed_out(struct tcmu_device *dev, bool can_access_dev)
```
dillaman, line 546: Nit: is the `can_access_dev` needed? It looks like everyone calls this method w/ `true`

tcmur_work.c
lxbsz: Yeah, right.

I have removed the thread in work and will use the pthread name to identify whether the work thread itself is trying to call the tcmur_flush_work(), which will cause stuck forever.

tcmur_device.c
lxbsz: Removed.

tcmur_work.c
lxbsz: Together with this versoin, I have remove the lock_work and since we can run multiple thread by using one event_work. And also remove the event type when in event reporting helper, since we are always reporting all the event every time.

lxbsz:
Merged, thanks.

PR #632, opened 2020-07-31
worker: do not set thread name in the callback fn

lxbsz:
It will overwrite the event worker thread's "ework-thread" name,
which is very important and we will use it to avoid recursive loop.


lxbsz:
Thanks Jason, merging it.

PR #633, opened 2020-08-01
CMakeList.txt: fix the style and switch tabes to white spaces

lxbsz:
There has mixed using of tabes and white spaces.


lxbsz:
Thanks Jason, will merge it.

PR #634, opened 2020-08-04
rbd: remove all the stale blacklist entries when opening devices 

lxbsz:
When a device is closing, we will add its entiry address into a list
cache. It may have already been blacklisted by another tcmu node
in reopen case.

And all the stale blacklist entris recorded in the list cache will
be removed in any new comming device doing the open.

Fallback to use lock owner API to store the blacklist entry if the
rados_getaddrs() is not supported, though using the acquire lock
APIs we may miss some blacklist entries.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -25,6 +25,7 @@
 
 #include <scsi/scsi.h>
 
+#include "darray.h"
 #include "tcmu-runner.h"
 #include "tcmur_cmd_handler.h"
 #include "libtcmu.h"
@@ -80,6 +81,7 @@ struct tcmu_rbd_state {
 	char *osd_op_timeout;
 	char *conf_path;
 	char *id;
+	char *addrs;
 };
 
 enum rbd_aio_type {
@@ -106,6 +108,9 @@ struct rbd_aio_cb {
 	struct iovec *iov;
 	size_t iov_cnt;
 };
+
+static pthread_mutex_t blacklist_caches_lock = PTHREAD_MUTEX_INITIALIZER;
+static darray(char *) blacklist_caches;
 
 #ifdef LIBRADOS_SUPPORTS_SERVICES
 
@@ -238,6 +243,113 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
 #endif /* RBD_LOCK_ACQUIRE_SUPPORT */
 
 #endif /* LIBRADOS_SUPPORTS_SERVICES */
+
+#if defined LIBRADOS_SUPPORTS_GETADDRS || defined RBD_LOCK_ACQUIRE_SUPPORT
+static void tcmu_rbd_rm_stale_entry_from_blacklist(struct tcmu_device *dev, char *addrs)
+{
+	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
+	const char *p, *q, *end;
+	char *cmd, *addr;
+	int ret;
+
+	/*
+	 * Just skip extra chars before '[' if there has
+	 */
+	p = strchr(addrs, '[');
+	if (!p)
+		p = addrs;
+
+	/*
+	 * The addrs will a string like:
+	 * "[192.168.195.172:0/2203456141,192.168.195.172:0/4908756432]"
+	 * Or
+	 * "192.168.195.172:0/2203456141"
+	 */
+	while (1) {
+		if (p == NULL || *p == ']') {
+			return; /* we are done here */
+		} else if (*p == '[' || *p == ',') {
+			/* Skip "[" and white spaces */
+			while (*p != '\0' && !isalnum(*p)) p++;
+			if (*p == '\0') {
+				tcmu_dev_warn(dev, "Get an invalid address '%s'!\n", addrs);
+				return;
+			}
+
+			end = strchr(p, ',');
+			if (!end)
+				end = strchr(p, ']');
+
+			if (!end) {
+				tcmu_dev_warn(dev, "Get an invalid address '%s'!\n", addrs);
+				return;
+			}
+
+			q = end; /* The *end should be ',' or ']' */
+
+			while (*q != '\0' && !isalnum(*q)) q--;
+			if (*q == '\0') {
+				tcmu_dev_warn(dev, "Get an invalid address '%s'!\n", addrs);
+				return;
+			}
+
+			addr = strndup(p, q - p + 1);
+			p = end;
+		} else {
+			/* In case of "192.168.195.172:0/2203456141" */
+			addr = strdup(p);
+			p = NULL;
+		}
+
+		ret = asprintf(&cmd,
+			       "{\"prefix\": \"osd blacklist\","
+			       "\"blacklistop\": \"rm\","
+			       "\"addr\": \"%s\"}",
+			       addr);
+		free(addr);
+		if (ret < 0) {
+			tcmu_dev_warn(dev, "Could not allocate command. (Err %d)\n",
+				      ret);
+			return;
+		}
+		ret = rados_mon_command(state->cluster, (const char**)&cmd, 1, NULL, 0,
+					NULL, NULL, NULL, NULL);
+		free(cmd);
+		if (ret < 0) {
+			tcmu_dev_err(dev, "Could not rm blacklist entry '%s'. (Err %d)\n",
+				     addr, ret);
+			return;
+		}
+	}
+}
+
+static int tcmu_rbd_rm_stale_entries_from_blacklist(struct tcmu_device *dev)
+{
+	char **entry, *tmp_entry;
+	int ret = 0;
+	int i;
+
+	pthread_mutex_lock(&blacklist_caches_lock);
+	if (darray_empty(blacklist_caches))
+		goto unlock;
+
+	/* Try to remove all the stale blacklist entities */
+	darray_foreach(entry, blacklist_caches) {
+		tcmu_dev_info(dev, "removing addrs: {%s}\n", *entry);
+		tcmu_rbd_rm_stale_entry_from_blacklist(dev, *entry);
+	}
+
+unlock:
+	for (i = darray_size(blacklist_caches) - 1; i >= 0; i--) {
+		tmp_entry = darray_item(blacklist_caches, i);
+		darray_remove(blacklist_caches, i);
+		free(tmp_entry);
+	}
+
+	pthread_mutex_unlock(&blacklist_caches_lock);
+	return ret;
+}
+#endif // LIBRADOS_SUPPORTS_GETADDRS || RBD_LOCK_ACQUIRE_SUPPORT
 
 static char *tcmu_rbd_find_quote(char *string)
 {
@@ -768,6 +880,11 @@ static int tcmu_rbd_unlock(struct tcmu_device *dev)
 static int tcmu_rbd_lock(struct tcmu_device *dev, uint16_t tag)
 {
 	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
+#if !defined LIBRADOS_SUPPORTS_GETADDRS && defined RBD_LOCK_ACQUIRE_SUPPORT
+	rbd_lock_mode_t lock_mode;
+	char *owners1[1], *owners2[1];
+	size_t num_owners1 = 1, num_owners2 = 1;
+#endif
 	int ret;
 
 	ret = tcmu_rbd_has_lock(dev);
@@ -788,6 +905,41 @@ static int tcmu_rbd_lock(struct tcmu_device *dev, uint16_t tag)
 	ret = rbd_lock_acquire(state->image, RBD_LOCK_MODE_EXCLUSIVE);
 	if (ret)
 		goto done;
+
+#if !defined LIBRADOS_SUPPORTS_GETADDRS && defined RBD_LOCK_ACQUIRE_SUPPORT
+	ret = rbd_lock_get_owners(state->image, &lock_mode, owners1,
+				  &num_owners1);
+	if ((!ret && !num_owners1) || ret < 0) {
+		tcmu_dev_warn(dev, "Could not get lock owners to store blacklist entry %d!\n",
+			     ret);
+	} else {
+		int is_owner;
+
+		/* To check whether we are still the lock owner */
+		ret = rbd_is_exclusive_lock_owner(state->image, &is_owner);
+		if (ret) {
+			rbd_lock_get_owners_cleanup(owners1, num_owners1);
+			tcmu_dev_warn(dev, "Could not check lock owners to store blacklist entry %d!\n",
+				      ret);
+			goto no_owner;
+		}
+
+		/* To get the lock owner again */
+		ret = rbd_lock_get_owners(state->image, &lock_mode, owners2,
+				&num_owners2);
+		if ((!ret && !num_owners2) || ret < 0) {
+			tcmu_dev_warn(dev, "Could not get lock owners to store blacklist entry %d!\n",
+					ret);
+		/* Only we didn't lose the lock during the above check will we store the blacklist list */
+		} else if (!strcmp(owners1[0], owners2[0]) && is_owner) {
+			state->addrs = strdup(owners1[0]); // ignore the errors
+		}
+
+		rbd_lock_get_owners_cleanup(owners1, num_owners1);
+		rbd_lock_get_owners_cleanup(owners2, num_owners2);
+	}
+no_owner:
+#endif
 
 set_lock_tag:
 	tcmu_dev_warn(dev, "Acquired exclusive lock.\n");
@@ -837,6 +989,8 @@ static void tcmu_rbd_state_free(struct tcmu_rbd_state *state)
 		free(state->pool_name);
 	if (state->id)
 		free(state->id);
+	if (state->addrs)
+		free(state->addrs);
 	free(state);
 }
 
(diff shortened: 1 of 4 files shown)


rbd.c
```diff
@@ -239,6 +244,54 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
...
+	if (darray_empty(blacklist_caches))
+		goto unlock;
+
+	/* Try to remove all the stale blacklist entities */
+	darray_foreach(entry, blacklist_caches) {
+		tcmu_dev_info(dev, "removing addrs: {%s}\n", *entry);
+		ret = asprintf(&cmd,
+			      "{\"prefix\": \"osd blacklist\","
+			      "\"blacklistop\": \"rm\","
+			      "\"addr\": \"%s\"}", *entry);
```
dillaman, line 265: This probably needs to be tested. That `rados_getaddrs` call might return a address vector (i.e. `[v2:192.168.1.3:40945,v1:192.168.1.3:40946]`) but blacklisting might only act on a single address (i.e. `192.168.1.3:0/<nonce>`).

rbd.c
```diff
@@ -239,6 +244,54 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
...
+			free(cmd);
+			goto unlock;
+		}
+		ret = rados_mon_command(state->cluster, (const char**)&cmd, 1, NULL, 0, NULL, NULL,
+					NULL, NULL);
+		if (ret < 0) {
+			tcmu_dev_err(dev, "Could not rm blacklist entry '%s'. (Err %d)\n",
+				     *entry, ret);
+			free(cmd);
+			goto unlock;
```
dillaman, line 278: Nit: should it still keep trying the other addresses?

rbd.c
```diff
@@ -1484,6 +1563,28 @@ static int tcmu_rbd_reconfig(struct tcmu_device *dev,
...
+	if (darray_empty(blacklist_caches))
+		goto unlock;
+
+	/* Try to remove all the stale blacklist entities */
+	darray_foreach(entry, blacklist_caches) {
+		free(*entry);
+		i++;
+	}
+
+	darray_free(blacklist_caches);
```
dillaman, line 1582: Perhaps you can have an `init` handler to mimic the `destroy` handler to avoid the asymmetric initialization of this array vs its destruction?

rbd.c
lxbsz: Do you mean the `cmd`, right ? 

It will be alloced from asprint() with the size needed every time. BTW, will the `nonce` 
 have different lengths for different sessions some times ? Or will they always have a same string length ? 

Maybe we could just alloc a memory and to reuse it.

rbd.c
lxbsz: Yeah, make sense, I will test this.

rbd.c
```diff
@@ -25,6 +25,7 @@
 
 #include <scsi/scsi.h>
 
+#include "darray.h"
```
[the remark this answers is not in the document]
lxbsz: Will fix it.

rbd.c
```diff
@@ -239,6 +244,54 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
 
 #endif /* LIBRADOS_SUPPORTS_SERVICES */
 
+static int tcmu_rm_stale_entries_from_blacklist(struct tcmu_device *dev)
```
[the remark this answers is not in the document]
lxbsz: Will fix it.

rbd.c
```diff
@@ -1000,6 +1065,20 @@ static void tcmu_rbd_close(struct tcmu_device *dev)
 	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
 
 	tcmu_rbd_image_close(dev);
+
+	/*
+	 * Since we are closing the device, but current device maybe
+	 * already blacklisted by other tcmu nodes. Let's just save
+	 * the entiry addrs into the blacklist_caches, and let's any
```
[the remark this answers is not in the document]
lxbsz: Will fix it.

rbd.c
```diff
@@ -1000,6 +1065,20 @@ static void tcmu_rbd_close(struct tcmu_device *dev)
...
+	 * other new comming device to help remove it.
```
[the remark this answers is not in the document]
lxbsz: Will fix it.

rbd.c
lxbsz: Sound good. I will do it.

tcmu-runner.h
```diff
@@ -194,6 +194,9 @@ struct tcmur_handler {
 	 * Update the logdir called by dynamic config thread.
 	 */
 	bool (*update_logdir)(void);
+
+	/* To destory some global resrouces if needed */
```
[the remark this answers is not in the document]
lxbsz: Will fix it.

rbd.c
lxbsz: Currently in none `DPDK` case, there will only one `processer` in the `Messenger `, and then the `Messenger::my_addrs` will only be single one, like the logs:

```
 124 2020-08-05T14:30:18.410+0800 7f579b7fe700 25 --  >> [v2:10.72.47.137:40500/0,v1:10.72.47.137:40501/0] conn(0x55726e5fb930 msgr2=0x557     26e60bca0 unknown :-1 s=STATE_CONNECTION_ESTABLISHED l=1)._try_send cs.send 72 bytes
 125 2020-08-05T14:30:18.410+0800 7f579bfff700 10 -- 10.72.47.136:0/3287216816 learned_addr had no addrs                                  
 126 2020-08-05T14:30:18.410+0800 7f579bfff700  1 -- 10.72.47.136:0/3287216816 learned_addr learned my addr 10.72.47.136:0/3287216816 (peer_addr_for_me v2:10.72.47.137:0/0)
```

It will be `learned my addr 10.72.47.136:0/3287216816`, not the `peer_addr_for_me v2:10.72.47.137:0/0`

The `[v2:192.168.1.3:40945,v1:192.168.1.3:40946]` you mentioned above you meant the monitor listening addresses in `ceph.conf`, right ? If so, even if the cluster enables msgr1 and msgr2 at the same time, the monclient will only connect to one of them and it will always choose the first one from `[v2:XXXX,v1:XXXX]`.

We won't use the DPDK here, right ?

rbd.c
dillaman: I just mean that it you stop iterating through the `darray` once you hit the first error.

rbd.c
dillaman: > The `[v2:192.168.1.3:40945,v1:192.168.1.3:40946]` you mentioned above you meant the monitor listening addresses in `ceph.conf`, right ? If so, even if the cluster enables msgr1 and msgr2 at the same time, the monclient will only connect to one of them and it will always choose the first one from `[v2:XXXX,v1:XXXX]`.

It's still a `entity_addrvec_t` internally which is converted to a string. I just want to make sure the result of that API call matches the `10.72.47.136:0/3287216816`-style format.
 
> We won't use the DPDK here, right ?

It won't be using DPDK. It would be using the standard POSIX async messenger.

rbd.c
```diff
@@ -983,6 +1043,14 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 				    tcmu_dev_get_block_size(dev), false);
 	tcmu_dev_set_write_cache_enabled(dev, 0);
 
+	tcmu_rbd_rm_stale_entries_from_blacklist(dev);
+
+	/* Get current entry address for the image */
+	ret = rados_getaddrs(state->cluster, &state->addrs);
```
dillaman, line 1049: Nit: need to condition this usage on this API being available.

rbd.c
lxbsz: > It's still a entity_addrvec_t internally which is converted to a string. I just want to make sure the result of that API call matches the 10.72.47.136:0/3287216816-style format.

Yeah, that make sense.

Checked it code, when there has only one address in that vector it will return a string, it will be like `10.72.47.136:0/3287216816`, only when there has more than one addresses, it will be `['10.72.47.136:0/3287216816', '10.72.47.136:0/4587216543']`.

The code:

```
708   friend std::ostream& operator<<(std::ostream& out, const entity_addrvec_t& av) {
709     if (av.v.empty()) {
710       return out;
711     } else if (av.v.size() == 1) {
712       return out << av.v[0];
713     } else {                                                                                                                          
714       return out << av.v;
715     }
716   }
```

But it will be always wise to compatible to this case, and I will fix it.
Thanks

rbd.c
lxbsz: I have raise one PR for this : https://github.com/ceph/ceph/pull/36488

lxbsz:
@dillaman 
Updated it. For this version, it will allows the state->addrs string to be any of the following format:

`"[192.168.195.172:0/2203456141,192.168.195.172:0/4908756432]"`

Or

`"192.168.195.172:0/2203456141"`

lxbsz:
@dillaman 
This version will fallback to use the lock APIs to get the blacklist entries when the `rados_getaddrs()` is not supported. Though it may miss some, that's should be in rare cases.

rbd.c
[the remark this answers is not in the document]
dillaman: Looks like this is still in the commit message -- was the goal `entries` or `entities`?

rbd.c
```diff
@@ -768,6 +875,11 @@ static int tcmu_rbd_unlock(struct tcmu_device *dev)
 static int tcmu_rbd_lock(struct tcmu_device *dev, uint16_t tag)
 {
 	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
+#if !defined LIBRBD_SUPPORTS_GETADDRS && defined RBD_LOCK_ACQUIRE_SUPPORT
```
dillaman, line 878: Nit: I wouldn't worry about adding this alternative approach. We will work to backport the new API to Octopus and Nautilus so we just need the guards on `LIBRADOS_SUPPORTS_GETADDRS` (not `LIBRBD_SUPPORTS_GETADDRS`) to prevent compile errors.

rbd.c
```diff
@@ -239,6 +249,103 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
...
+	 * Or
+	 * "192.168.195.172:0/2203456141"
+	 */
+	while (1) {
+		if (p == NULL || *p == ']') {
+			entry->is_removed = true;
+			return;
+		} else if (*p == '[' || *p == ',') {
+			/* Skip "[ " */
+			while (!isalnum(*p)) p++;
```
dillaman, line 272: Nit: avoid `NULL` as well

rbd.c
```diff
@@ -239,6 +249,103 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
...
+
+	/*
+	 * The entry->addrs will a string like:
+	 * "[192.168.195.172:0/2203456141,192.168.195.172:0/4908756432]"
+	 * Or
+	 * "192.168.195.172:0/2203456141"
+	 */
+	while (1) {
+		if (p == NULL || *p == ']') {
+			entry->is_removed = true;
```
dillaman, line 268: Hmm -- I'm not sure it's worth keeping entries around if you fail to remove them. I was originally just suggesting to try all the cached addresses and don't just stop when you hit the first error (if any)

rbd.c
```diff
@@ -1000,6 +1161,28 @@ static void tcmu_rbd_close(struct tcmu_device *dev)
...
+		if (entry) {
+			entry->addrs = state->addrs;
+			entry->is_removed = false;
+			state->addrs = NULL;
+
+			pthread_mutex_lock(&blacklist_caches_lock);
+			darray_append(blacklist_caches, entry);
+			pthread_mutex_unlock(&blacklist_caches_lock);
+		} else {
+			tcmu_dev_warn(dev, "Failed to add to blacklist caches, it will be removed automatically after timedout!\n");
```
dillaman, line 1182:
```suggestion
			tcmu_dev_warn(dev, "Failed to add to blacklist caches\n");
```

rbd.c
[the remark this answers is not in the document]
lxbsz: It's entity address. Will fix it.

rbd.c
lxbsz: Currently it will try all the cache list entries one by one without considering whether there has failure for some of them.

Yeah, it seems making no sense to try it again, so let's just remove it from the cache list even it fails.

rbd.c
lxbsz: The string is `'\0'` ternimated, so there is no any chance it could be `NULL` in the `else if` case.

rbd.c
lxbsz: Yeah, okay, let's keep it just for the case for some one who are using the old ceph versions without the backporting.

rbd.c
lxbsz: Will fix it.

rbd.c
dillaman: If you have "[\0" as a string, wouldn't this just keep looping past the bounds?

lxbsz: BTW, in which case will this happen ? 

From the ceph code, it should always be '\0' terminated c-string for this address(es), it will be coverted from a vector if there have more than one addresses. And also from my local test the string coverted from a verctor will always be '\0' terminated.

Did I miss something about this ?

lxbsz: Oh, maybe we should switch to use client_lock.is_locked_by_me(), right.

dillaman: No, I am just suggesting you defend against malformed data.

rbd.c
```diff
@@ -789,6 +895,41 @@ static int tcmu_rbd_lock(struct tcmu_device *dev, uint16_t tag)
 	if (ret)
 		goto done;
 
+#if !defined LIBRADOS_SUPPORTS_GETADDRS && defined RBD_LOCK_ACQUIRE_SUPPORT
```
dillaman, line 898: Nit: I am still voting for this code to be removed. The `LIBRADOS_SUPPORTS_GETADDRS` support will be backported through nautilus so for the sake of code-clarity, we shouldn't have to worry about alternative approached.

rbd.c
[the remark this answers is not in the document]
dillaman: Still needs to be addressed

rbd.c
dillaman: The `LIBRADOS_SUPPORTS_GETADDRS` support will be backported through nautilus so for the sake of code-clarity and simplicity, we shouldn't have to worry about alternative approaches (especially since the alternative has a potential race condition).

rbd.c
lxbsz: I think I have lost my last comments about this , and I had fixed it already. Thanks.

rbd.c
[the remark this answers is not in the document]
lxbsz: Sorry, I maybe missed it.

rbd.c
lxbsz: Yeah, that's cool.

For the alternative approach, there should have the potential race condition only in theory when one tcmu-runner node gettting the same nonce twice.

This alternative approach here is a little different from the one we disscussed in the mail before, it will do:

1, after the local tcmu-runner node successfully getting the exclusive lock, it will try to get the lock owner and save it as `lock_owner1`.

2, then check if the local tcmu-runner node is the lock owner. If not, it means local tcmu-runner node had lost the lock, so we cannot trust the `lock_owner1`, then just skip saving the entity addr to local cache arrary and no need to do the following steps.

3, If local node is still the lock owner in Step2, then try to get the lock owner again and save it as `lock_owner2`.

4, only when `lock_owner1` == `lock_owner2`, that means we never lost the lock during Step1. Then we can trust the `lock_owner1` and save it to local cache arrary.

In Step4, if the IP address is foreign tcmu-runner node's, that means during both Step1 and Step3, the lock is acquired by the same foreign tcmu-runner node. and in Step2 the local node acquired it again.  And at the same time in Step1 and Step3 the `nonces` are the same, IMO this should happen only in theory.

Let's just assume the above case won't happen, that means during Step1, Step2 and Step3, the local node never loses the lock.

rbd.c
lxbsz: Okay, will fix it. Thanks.

rbd.c
```diff
@@ -107,6 +109,13 @@ struct rbd_aio_cb {
 	size_t iov_cnt;
 };
 
+struct blacklist_entry {
```
dillaman, line 112: Nit: unused now?

rbd.c
```diff
@@ -239,6 +248,110 @@ static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
...
+				end = strchr(p, ']');
+
+			if (!end) {
+				tcmu_dev_warn(dev, "Get an invalid address '%s'!\n", addrs);
+				return;
+			}
+
+			q = end; /* The *end should be ',' or ']' */
+
+			while (!isalnum(*q)) q--;
```
dillaman, line 295: Nit: same issue here

rbd.c
dillaman: > In Step4 if the IP address is local tcmu-runner nodes's, and it has lost the lock once, then in local node it also got the same nonce twice.

For what it's worth, the nonce will only change if you disconnect from the cluster and reconnect -- which hopefully isn't happening while you are within this method.

rbd.c
lxbsz: I removed the '\0' check here yesterday.

Because in Line#279~#286 the `strchr()` will make sure that the string between pointers `p` and `end` has no '\0'.  If there has, it will return from Line#181 or Line#290 directly.

And the `while()` here is backward decreasing the `end` pointer. And the `p` is must pointing a letter or a number, I just supposed it makes no sense to check the '\0' here.

Since you mentioned it, or maybe there has some potential issues or something that I might miss, but it is harmless to do an extra check of this. So let me add it back.

rbd.c
lxbsz: Good catch, I will remove it.

rbd.c
lxbsz: Yeah, it is. And the Step2 is a barrier.

Since it will reopen/reconnect the image if the local/foreign node will re-acquire the lock. 

If in Step2 it says the local node is the lock owner, it means local node never lost the lock or has re-acquired the lock since Step1, if it is the later case, the reconnection must already be done.

No matter the IP is local or foreign node's in `lock_owner1` and `lock_owner2`. Without considering the case that one node will get a same `nonce` value with previous connection,  if Step4 says `lock_owner1` and `lock_owner2` are the same, we can say that the local node never lost the lock.

I have run the `nonce` generator code for round 1 thousand times, didn't get a same nonce with the existings. WIth this approach we may miss removing some blacklist entries, but it shouldn't matter.

lxbsz:
> lgtm

Thanks @dillaman , merged it.

PR #636, opened 2020-08-21
lock: do not try to acquire the lock for read fops

lxbsz:
After the connection was lost or the lock was broke by other nodes,
and for the read fops just skip acquiring the lock.


tcmur_device.h
```diff
@@ -32,10 +32,40 @@ enum {
...
 	TCMUR_DEV_LOCK_UNLOCKED,
 	TCMUR_DEV_LOCK_LOCKED,
+
+	/*
+	 * After the connection lost and before the first read fop
+	 * coming there was no any write fop for the device, then
+	 * the first read fop will set it and then reopen the device
+	 * without acquiring the lock later.
+	 */
+	TCMUR_DEV_LOCK_READ_REOPENING,
```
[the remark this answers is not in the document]
lxbsz: > And one comment myself here about the current code, we still need to break the lock in read case and could skip acquiring it, because in case:
> 
> If current pathA lost the connection and if it have read fops, it will reopen the image without breaking and acquiring the lock, the pathB still could do read/write. There will be one problem if pathA want to do read-modiry-write things, the same offset maybe already updated by pathB just before pathA writing the modified data.
> 
> So we need to break the lock to blacklist pathB and to make sure the lock has no owner for every read in pathA, that means we must be sure the image is `readonly` temporily.

For the Read-Modify-Write or RMW case, from the linux scsi code, the old reopen code also has the same issue. Because  in kernel space the linux block/scsi code will just retry the cmds if its error == SAM_TS_BUSY. So for AA or shared device use case, if the initiator/apps want to do the write it must use the Revervation for locking or the atomic CAW/Compare-And-Write scsi cmd. 

The VMWare use case has no this issue.

alua.c
```diff
@@ -560,17 +561,57 @@ int alua_implicit_transition(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+	} else if (rdev->lock_state == TCMUR_DEV_LOCK_WRITE_LOCKING) {
+		/* For both read/write cases in this state should return busy */
+		tcmu_dev_dbg(dev, "Write lock acquisition operation is already in process.\n");
 		ret = TCMU_STS_BUSY;
 		goto done;
+	} else if (is_read && !rdev->lock_lost) {
+		/* For read case we will only need to break the lock. */
+		if (rdev->lock_state == TCMUR_DEV_LOCK_READ_LOCKED) {
+			if (!tcmu_lock_has_owner(dev))
+				goto done;
```
lxbsz, line 576: Maybe we do not need to do this ?

tcmur_device.h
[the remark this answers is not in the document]
lxbsz: If this is true, then I will remove the READ_LOCKXXX flags.

lxbsz:
Updated it by removing the break lock stuff for read. It will do:

For read, it will just reopen the device without acquiring the lock. This will be like a read lock. Will use the `TCMUR_DEV_LOCK_READ_LOCKING` and `TCMUR_DEV_LOCK_READ_LOCKED` to just tell the followed reads that the reopen is in process or done.

For write, it will just try to acquire the lock if the previous reads have reopened the device, or it will reopen the device and then acquire the lock. It will use the `TCMUR_DEV_LOCK_WRITE_LOCKING` and `TCMUR_DEV_LOCK_WRITE_LOCKED`  to tell the followed reads/writes, the reopen/acqiure lock is in process or done.

There is one case that, if one write is followed a read imediately, and if the read is trying to reopen the device, but not finished yet, the write will try to reopen it too, then for the write thread it may just upgrade the lock `TCMUR_DEV_LOCK_READ_LOCKING` --> `TCMUR_DEV_LOCK_WRITE_LOCKING`, etc.

tcmur_device.h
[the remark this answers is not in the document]
dillaman: Right. Agree that you need reservations or use the atomic-test-set command.

dillaman:
> If the device reopening and lock breaking for read is still in process, the write comes, we need to upgrade the read lock to write lock, etc. And also for read we only need to break the lock so we need to flag what the lock type is.

The goal was to avoid breaking the lock on read since it leads to path bouncing when the path is probed by ESX.

tcmur_device.c
```diff
@@ -434,6 +437,15 @@ int tcmu_acquire_dev_lock(struct tcmu_device *dev, uint16_t tag)
 		}
 	}
 
+	pthread_mutex_lock(&rdev->state_lock);
+	if (rdev->lock_state == TCMUR_DEV_LOCK_READ_LOCKING) {
+		pthread_mutex_unlock(&rdev->state_lock);
+		no_lock = true;
+		ret = TCMU_STS_OK;
+		goto drop_conn;
```
dillaman, line 445: Nit: 
```suggestion
		goto done;
```

alua.c
```diff
@@ -560,17 +561,49 @@ int alua_implicit_transition(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		/*
+		 * Since we are here, current lock state should be one of:
+		 *   TCMUR_DEV_LOCK_UNLOCKED
+		 *   TCMUR_DEV_LOCK_UNKNOWN
+		 *   TCMUR_DEV_LOCK_READ_LOCKING
+		 *   TCMUR_DEV_LOCK_READ_LOCKED
+		 *
+		 * May will reopen the deivce and Will acquire the lock later.
+		 */
+		rdev->lock_state = TCMUR_DEV_LOCK_WRITE_LOCKING;
```
dillaman, line 605: If you are in `READ_LOCKING` state so it's attempting to open the image and now a write comes in, wouldn't this cause a second thread to start re-opening the image and attempting to acquire the write lock? Should it instead just return `BUSY` in this case to retry acquiring the write lock -- especially because `tcmu_acquire_dev_lock` acquires and drops the lock numerous times so it seems like lots of potential for race conditions.

I know you said it will be a busy loop between the kernel and tcmu-runner when the `STS_BUSY` result is handled, but it's going to be doing that regardless on successive reads while re-opening the image and on successive writes.

lxbsz:
> > If the device reopening and lock breaking for read is still in process, the write comes, we need to upgrade the read lock to write lock, etc. And also for read we only need to break the lock so we need to flag what the lock type is.
> 
> The goal was to avoid breaking the lock on read since it leads to path bouncing when the path is probed by ESX.

Yeah, from my test the most time was cost the lock acquiring. And the newest change of this PR I have removed the breaking only logic for read.

tcmur_device.c
lxbsz: Yeah, make sense. I will fix it.

alua.c
lxbsz: > If you are in `READ_LOCKING` state so it's attempting to open the image and now a write comes in, wouldn't this cause a second thread to start re-opening the image and attempting to acquire the write lock?

It's a good question but the re-opening issue shouldn't happen, but the acquiring the write lock in parallel may will:

There has 3 cases:
Case1:
If the first thread still not finished from `tcmu_reopen_dev()`, it will set `rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY`, so for the second thread, though it may set `reopen = true`, because the first thread may still not set the `rdev->lock_lost = false`. The second thread will just get a `EBUSY` error from `tcmu_reopen_dev()`, then `goto drop_conn` without trying to acquire the lock. But the second have already upgrade the lock state to `TCMUR_DEV_LOCK_WRITE_LOCKING`. So the first thread will try to acquire the lock after it returned from `tcmu_reopen_dev()`.

Case2:
If the first thread has just returned from `tcmu_reopen_dev()` when the second thread begins to run, because the first thread has already set the `rdev->lock_lost = false` in `tcmu_reopen_dev()`, the it won't do the device reopen. But it is possible that both the two threads will see that the lock state is `TCMUR_DEV_LOCK_WRITE_LOCKING` and then do the `rhandler->lock()` in parallel. I think I need to fix this.

The above two cases, if just return `EBUSY` here, it works well. But for the following Case3, it won't. 

Case3:
If the first thread has already finished from `tcmu_reopen_dev()` and is runing the code after `drop_conn:`, that means the first thread didn't see the lock state is `TCMUR_DEV_LOCK_WRITE_LOCKING`, so it didn't acquire the lock. And then we need the second thread to continue to acquire lock only.  This is why I add the `no_lock` in `tcmu_acquire_dev_lock()`.


> Should it instead just return `BUSY` in this case to retry acquiring the write lock -- especially because `tcmu_acquire_dev_lock` acquires and drops the lock numerous times so it seems like lots of potential for race conditions.
> 
> I know you said it will be a busy loop between the kernel and tcmu-runner when the `STS_BUSY` result is handled, but it's going to be doing that regardless on successive reads while re-opening the image and on successive writes.

Yeah, the retry happens in the initiator's scsi driver in kernel space. This logic should be in compliance with the scsi proto.

dillaman: Right, it just seems cleaner to let it retry via `STS_BUSY` instead of kicking off multiple threads.

lxbsz: Yeah, this sounds cool. And I have fixed it.

alua.c
```diff
@@ -560,17 +561,55 @@ int alua_implicit_transition(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
-		tcmu_dev_dbg(dev, "Lock acquisition operation is already in process.\n");
+	} else if (rdev->lock_state == TCMUR_DEV_LOCK_WRITE_LOCKING) {
+		/* For both read/write cases in this state should return busy */
+		tcmu_dev_dbg(dev, "Write lock acquisition operation is already in process.\n");
 		ret = TCMU_STS_BUSY;
 		goto done;
+	} else if (is_read && !rdev->lock_lost) {
+		if (rdev->lock_state == TCMUR_DEV_LOCK_READ_LOCKED) {
+			goto done;
+		} else if (rdev->lock_state == TCMUR_DEV_LOCK_READ_LOCKING) {
```
dillaman, line 575: Nit: can this be bumped up above line 572 and then you could delete the extra check on lines 605-609?

alua.c
```diff
@@ -560,17 +561,55 @@ int alua_implicit_transition(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
+		/* For both read/write cases in this state is good */
 		goto done;
-	} else if (rdev->lock_state == TCMUR_DEV_LOCK_LOCKING) {
-		tcmu_dev_dbg(dev, "Lock acquisition operation is already in process.\n");
+	} else if (rdev->lock_state == TCMUR_DEV_LOCK_WRITE_LOCKING) {
+		/* For both read/write cases in this state should return busy */
+		tcmu_dev_dbg(dev, "Write lock acquisition operation is already in process.\n");
 		ret = TCMU_STS_BUSY;
 		goto done;
+	} else if (is_read && !rdev->lock_lost) {
```
dillaman, line 572: Nit: does this `lock_lost` check matter? It looks like `__tcmu_notify_lock_lost` will set this to `true` and set the state back to `UNLOCKED`.

lxbsz: Hmm, I will remove this.

alua.c
lxbsz: Yeah, makes sense, I will do it.
Thanks.

dillaman: Nit: so now this statement will not be reached so it can be removed

lxbsz: Done.

lxbsz:
> lgtm -- thx

Thanks a lot @dillaman

PR #639, opened 2020-09-23
Add supported vpd pages to zbc vpd pages

ryanattard:
The supported vpd pages page is required by SPC. It should be included in itself.

This resolves an issue where expected sysfs attributes (specifically scsi_device/vpd_pg0 in new kernels)
is not populated on devices emulated with file_zbc.c.

### DIFF
--- a/file_zbc.c
+++ b/file_zbc.c
@@ -952,12 +952,13 @@ static int zbc_evpd_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
 	case 0x00:
 		/* Supported VPD pages */
 		data[3] = 5;
-		data[4] = 0x83;
-		data[5] = 0xb0;
-		data[6] = 0xb1;
-		data[7] = 0xb6;
+		data[4] = 0x0;
+		data[5] = 0x83;
+		data[6] = 0xb0;
+		data[7] = 0xb1;
+		data[8] = 0xb6;
 
-		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, 8);
+		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, 9);
 		break;
 
 	case 0x83:


ryanattard:
Specifically, it states `The supported VPD page list shall contain a list of all VPD page codes (see 7.7) implemented by the logical unit in ascending order beginning with page code 00h.` which makes the current implementation non-compliant.

ryanattard:
See existing code for other plugins: https://github.com/open-iscsi/tcmu-runner/blob/ddc7d14169540fdfe9e75c6e9e31ba09a25c9dd1/scsi.c#L104

ryanattard:
I found a bug in this patch during verification, I'll push out a squashed up commit after I've run additional testing.

ryanattard:
Verified 1fd43f2 by generating an emulated ZBC device and issuing the VPD command.  I ran the libzbc test suite and some IO against the disk just to be double sure. 

```
# Before Update
root@mymachine:~# sg_vpd -p 0x0 /dev/sdf  -rr | xxd
00000000: 1400 0005 83b0 b1b6 4c                   ........L
root@mymachine:~# find /sys/block/sdf/device/vpd*
/sys/block/sdf/device/vpd_pg83

# Post Update
root@mymachine:~# sg_vpd /dev/sdf -rr | xxd
00000000: 1400 0006 0083 b0b1 b600                 ..........
root@mymachine:~# find /sys/block/sdf/device/vpd*
/sys/block/sdf/device/vpd_pg0
/sys/block/sdf/device/vpd_pg83

```

file_zbc.c
```diff
@@ -951,13 +951,14 @@ static int zbc_evpd_inquiry(struct tcmu_device *dev, struct tcmulib_cmd *cmd)
...
 	case 0x00:
 		/* Supported VPD pages */
-		data[3] = 5;
-		data[4] = 0x83;
-		data[5] = 0xb0;
-		data[6] = 0xb1;
-		data[7] = 0xb6;
-
-		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, 8);
+		data[3] = 6;
```
lxbsz, line 954: Shouldn't this be 5 ?

lxbsz: The PAGE LENGTH  should be `n-3` and `n == 8`.

ryanattard: I agree with this, and that explains the `4c` and `00` trailing bytes in both of my prior posted examples. I just incremented all the numbers by 1, so this must be an artifact of the original implementation.

ryanattard: Tested and pushed.

```
root@mymachine:~# sg_vpd -p 0x0 /dev/sdf -rr | xxd
00000000: 1400 0005 0083 b0b1 b6                   .........
```

lxbsz:
LGTM. @ryanattard Thanks.

file_zbc.c
lxbsz: Yeah, the first one, the data[4], should always be `0x00`.

PR #643, opened 2020-11-16
scsi: minor simplification of inquiry if/else logic

ddiss:
The else clauses can be dropped when the if section returns.


lxbsz (approved):
LGTM. Thanks.

PR #644, opened 2021-01-12
tcmur: fail cross-device XCOPY requests

ddiss:
tcmu-runner can't determine whether the device(s) referred to in XCOPY
Copy Source/Copy Destination (CSCD) descriptors should be accessible to
the initiator via transport settings, ACLs, etc. Consequently, fail
XCOPY requests with CSCD descriptors which refer to any device other
than where the XCOPY request is processed.

References: CVE-2020-28374
Fixes: 9c86bd0 ("tcmur: Add emulate XCOPY command support")
Reviewed-by: Lee Duncan <lduncan@suse.com>

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -1349,6 +1349,18 @@ static int xcopy_parse_parameter_list(struct tcmu_device *dev,
 	if (ret != TCMU_STS_OK)
 		goto err;
 
+	/*
+	 * tcmu-runner can't determine whether the device(s) referred to in an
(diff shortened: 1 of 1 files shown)


pkalever (approved):
Looks good to me.

dillaman (approved):
lgtm

ddiss:
JFYI, @msmeissn mentioned that there's a new CVE number to track this separate to the Linux kernel fix:

> For tcmu-runner Mitre suggested that we use a different CVE as its not the same codebase.
> Please use CVE-2021-3139 for tcmu-runner.

PR #651, opened 2021-03-04
rbd: switch to use rbd_aio_write_zeroes for unmap

lxbsz:
The write zeroes API will allow the size not to align to the UNMAP
granularity, this can make sure that the specified area will always
be zeroed or discarded. While the discard API will do nothing when
the option rbd_skip_partial_discards is true.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -1022,8 +1022,9 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 	char *pool, *name, *next_opt;
 	char *config, *dev_cfg_dup;
 	struct tcmu_rbd_state *state;
-	uint32_t max_blocks;
+	uint32_t max_blocks, unmap_gran;
 	int ret;
+	char buf[128];
 
 	state = calloc(1, sizeof(*state));
 	if (!state)
@@ -1133,8 +1134,20 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 	max_blocks = (image_info.obj_size * 4) / tcmu_dev_get_block_size(dev);
 	tcmu_dev_set_opt_xcopy_rw_len(dev, max_blocks);
 	tcmu_dev_set_max_unmap_len(dev, max_blocks);
-	tcmu_dev_set_opt_unmap_gran(dev, image_info.obj_size /
-				    tcmu_dev_get_block_size(dev), false);
+	ret = rados_conf_get(state->cluster, "rbd_discard_granularity_bytes", buf,
+			     sizeof(buf));
+	if (!ret) {
+		tcmu_dev_dbg(dev, "rbd_discard_granularity_bytes: %s\n", buf);
+		unmap_gran = atoi(buf) / tcmu_dev_get_block_size(dev);
+	} else {
+		tcmu_dev_warn(dev,
+			      "Failed to get 'rbd_discard_granularity_bytes', %d\n",
+			      ret);
+		unmap_gran = image_info.obj_size / tcmu_dev_get_block_size(dev);
+	}
+	tcmu_dev_dbg(dev, "unmap_gran: %d\n", unmap_gran);
+	tcmu_dev_set_opt_unmap_gran(dev, unmap_gran, false);
+	tcmu_dev_set_unmap_gran_align(dev, unmap_gran);
 	tcmu_dev_set_write_cache_enabled(dev, 0);
 
 #if defined LIBRADOS_SUPPORTS_GETADDRS || defined RBD_LOCK_ACQUIRE_SUPPORT
--- a/rbd.c
+++ b/rbd.c
@@ -1629,36 +1629,6 @@ out:
 }
 #endif /* RBD_COMPARE_AND_WRITE_SUPPORT */
 
-/*
- * Return scsi status or TCMU_STS_NOT_HANDLED
- */
-static int tcmu_rbd_handle_cmd(struct tcmu_device *dev,
-			       struct tcmur_cmd *tcmur_cmd)
-{
-	struct tcmulib_cmd *cmd = tcmur_cmd->lib_cmd;
-	uint8_t *cdb = cmd->cdb;
-	int ret;
-
-	switch(cdb[0]) {
-#ifdef RBD_WRITE_SAME_SUPPORT
-	case WRITE_SAME:
-	case WRITE_SAME_16:
-		ret = tcmur_handle_writesame(dev, tcmur_cmd,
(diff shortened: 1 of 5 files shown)


rbd.c
```diff
@@ -1456,7 +1456,11 @@ static int tcmu_rbd_unmap(struct tcmu_device *dev, struct tcmur_cmd *tcmur_cmd,
 	if (ret < 0)
 		goto out_free_aio_cb;
 
+#if LIBRBD_SUPPORTS_WRITE_ZEROES
+	ret = rbd_aio_write_zeroes(state->image, off, len, completion, 0, 0);
```
dillaman, line 1460: I would just drop this commit -- the "write_zeroes" call would be more appropriate for the `write_same` SCSI command than the discard.

rbd.c
```diff
@@ -1022,8 +1022,9 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 	char *pool, *name, *next_opt;
```
dillaman, line 1022: Nit: typo in the commit message: `limiation` -> `limitation`

tcmur_cmd_handler.c
```diff
@@ -682,9 +682,19 @@ static int handle_unmap_in_writesame(struct tcmu_device *dev,
 	uint8_t *cdb = cmd->cdb;
```
dillaman, line 680: Nit: typo in commit message: `acquired` -> `required`? Technically I don't think the SCSI spec requires it to be anything more than block-aligned, though.

tcmur_cmd_handler.c
```diff
@@ -682,9 +682,19 @@ static int handle_unmap_in_writesame(struct tcmu_device *dev,
...
 	uint64_t lba = tcmu_cdb_get_lba(cdb);
 	uint64_t nlbas = tcmu_cdb_get_xfer_length(cdb);
+	uint32_t align = tcmu_dev_get_unmap_gran_align(dev);
 	struct unmap_state *state;
 	int ret;
 
+	/* If not aligned then falls back to the writesame without unmap */
+	if (lba % align || nlbas % align) {
+		tcmu_dev_warn(dev,
+			      "Start lba: %"PRIu64" or nlbas: %"PRIu64" not aligns to %"PRIu32"\n",
```
dillaman, line 690: `not aligns` -> `not aligned` (also probably not a warning since it's only an issue because `tcmu` is not treating the WS as a WS). Looking at the kernel's `drivers/scsi/sd.c`, it translates a zero extent call to a WS+UNMAP and doesn't attempt to convert it to an UNMAP. 

Since RBD and Gluster at the only two plugins that implement write-same, maybe it would be cleaner to just expose a direct hook w/ the `unmap` boolean and then let those two plugins attach to the hook directly instead of intercepting it in `XYZ_handle_cmd` and expecting an emulation-like callback from `tcmur_handle_writesame`

rbd.c
lxbsz: Done.

rbd.c
lxbsz: Okay, since that workaround could work well.

tcmur_cmd_handler.c
lxbsz: Yeah, right, check the spec, didn't find it. Will fix the commit comment.

tcmur_cmd_handler.c
lxbsz: > `not aligns` -> `not aligned` (also probably not a warning since it's only an issue because `tcmu` is not treating the WS as a WS). 

Sure, will fix it.

> Looking at the kernel's `drivers/scsi/sd.c`, it translates a zero extent call to a WS+UNMAP and doesn't attempt to convert it to an UNMAP.
> 
Checked it, it just translate the zero request to WA+UNMAP SCSI command, which the hardware could handle. 

The scsi disk hardware controller, which will be equivalent to tcmu-runner/librbd, will decide how to do if it receives a WS command with UNMAP bit set and not aligned. I guess, the scsi disk will just write zero to the segment not aligned. If all are aligned, the UNMAP should be the most efficient way to zero the disk. So internally it will switch to UNMAP I think. This is why the tcmu-runner is trying to switch to UNMAP if WS have explicitly allow it.

This is what the `drivers/target/target_core_sbc.c` will do.

lxbsz: > Since RBD and Gluster at the only two plugins that implement write-same, maybe it would be cleaner to just expose a direct hook w/ the `unmap` boolean and then let those two plugins attach to the hook directly instead of intercepting it in `XYZ_handle_cmd` and expecting an emulation-like callback from `tcmur_handle_writesame`

Yeah, looks better. Let's expose the handler->writesame and handler->caw hooks to clean up the code.

lxbsz:
Test result:

```
[root@client ~]# dd if=/dev/random of=/dev/sdc bs=1K count=1K
dd: warning: partial read (84 bytes); suggest iflag=fullblock
0+1024 records in
0+1024 records out
77652 bytes (78 kB, 76 KiB) copied, 0.138002 s, 563 kB/s

[root@client ~]# dd if=/dev/sdd bs=4K count=16 iflag=direct status=none | hexdump
0000000 1d4a c0fb ef54 32f2 a69b 2014 96e8 f3b4
0000010 e663 dde8 92ba 8bb4 050f e0be 23cb 374a
0000020 a0ec 735b 0223 24de ea30 e938 8285 d47b
0000030 fd9a cf89 3545 7603 12ab d2e0 3e7b 00ee
0000040 af09 be7d fa9a 7db8 a94f 9cc8 874a 5101
0000050 5561 8abb e06d 8d02 b34e ae86 f717 b7be
0000060 34ee 23cd 151a 2ed7 4861 4359 e55a d740
0000070 2231 29a0 7e65 cf44 5c00 cac4 bb09 64ba
0000080 db79 f649 8809 adce 9be0 680b 7e90 e902
0000090 7790 5c1f f633 8f31 e09a 185e 7c2f 56ee
00000a0 9be9 07c5 fd22 c4c5 034f 8452 f9b4 161c
00000b0 f0cf ab0a 19dd 7f3a 3e52 2039 14db a9e9
00000c0 70fd 910f e5c6 8be2 1351 56dd 15f1 79cc
00000d0 2f5a 9a72 3249 b68c c0b3 96d8 ebcc 0337
...

[root@client ~]# sg_write_same --unmap --num=127 /dev/sdb
[root@client ~]# dd if=/dev/sdc bs=1K count=1 iflag=direct status=none | hexdump
0000000 0000 0000 0000 0000 0000 0000 0000 0000
*
0000400
[root@client ~]# 
```

PR #656, opened 2021-05-24
tcmu_read_config(): check read() return value.

maurizio-lombardi:
tcmu-runner doesn't work correctly on Fedora 34 because it
enters an infinite loop while reading the /etc/tcmu/tcmu.conf config
file. The reason is that tcmu_read_config() doesn't check the read()
return value, relying only on errno's value, but the latter should
only be checked when read() returns -1.

Signed-off-by: Maurizio Lombardi <mlombard@redhat.com>


lxbsz (approved):
LGTM. Thanks.

PR #659, opened 2021-06-14
spec: make dist tag optional

ktdreyer:
Technically the `%{dist}` RPM macro is optional, so Fedora's convention is to use a question mark in the Release field.

https://docs.fedoraproject.org/en-US/packaging-guidelines/DistTag/

Practically speaking, `%{dist}` is almost always defined, so the question mark is not necessary, but this silences a warning in one of Red Hat's internal linters.


lxbsz (approved):
LGTM. Thanks.

PR #667, opened 2021-08-26
tcmur_device: add priv lock support

lxbsz:
When the tcmu-runner detect that the lock is lost, it will try to
queue a work event to reopen the image and at the same time queue
a work event to update the service status. While the reopen is not
atomic, and there has a gap between image close and image open,
during which the rbd image's state resource will be released and if
the update status event is fired, we will hit the crash bug.

This commit will add one rdev->priv_lock to protect the private data
in rdev struct. For the service status updating code just skip it
if it's in the reopen gap. And for all the other IOs just return
EBUSY to let the client try it again.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -146,17 +146,14 @@ static int tcmu_rbd_service_status_update(struct tcmu_device *dev,
 
 #endif /* RBD_LOCK_ACQUIRE_SUPPORT */
 
-static int tcmu_rbd_report_event(struct tcmu_device *dev)
+static int tcmu_rbd_report_event(struct tcmu_device *dev, bool has_lock)
 {
-	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
-
 	/*
 	 * We ignore the specific event and report all the current counter
 	 * values, because tools like gwcli/dashboard may not see every
 	 * update, and we do not want one event to overwrite the info.
 	 */
-	return tcmu_rbd_service_status_update(dev,
-			rdev->lock_state == TCMUR_DEV_LOCK_WRITE_LOCKED ? true : false);
+	return tcmu_rbd_service_status_update(dev, has_lock);
 }
 
 static int tcmu_rbd_service_register(struct tcmu_device *dev)
@@ -215,7 +212,7 @@ static int tcmu_rbd_service_register(struct tcmu_device *dev)
 		goto free_meta_buf;
 	}
 
-	ret = tcmu_rbd_report_event(dev);
+	ret = tcmu_rbd_report_event(dev, false);
 
 free_meta_buf:
 	free(metadata_buf);
--- a/tcmur_device.c
+++ b/tcmur_device.c
@@ -86,8 +86,8 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
 		rhandler->close(dev);
 	}
 
-	pthread_mutex_lock(&rdev->rdev_lock);
 	ret = -EIO;
+	pthread_mutex_lock(&rdev->rdev_lock);
 	while (ret != 0 && !(rdev->flags & TCMUR_DEV_FLAG_STOPPING) &&
 	       (retries < 0 || attempt <= retries)) {
 		pthread_mutex_unlock(&rdev->rdev_lock);
@@ -128,7 +128,10 @@ int tcmu_reopen_dev(struct tcmu_device *dev, int retries)
 		pthread_mutex_unlock(&rdev->rdev_lock);
 		return -EBUSY;
 	}
+
 	rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
+	if (rdev->flags & TCMUR_DEV_FLAG_REPORTING_EVENT)
+		pthread_cond_wait(&rdev->report_event_cond, &rdev->rdev_lock);
 	pthread_mutex_unlock(&rdev->rdev_lock);
 
 	return __tcmu_reopen_dev(dev, retries);
@@ -159,6 +162,7 @@ static void __tcmu_report_event(void *data)
 	struct tcmu_device *dev = data;
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
 	struct tcmur_handler *rhandler = tcmu_get_runner_handler(dev);
+	bool has_lock;
 	int ret;
 
 	/*
@@ -168,9 +172,29 @@ static void __tcmu_report_event(void *data)
 	sleep(1);
 
 	pthread_mutex_lock(&rdev->rdev_lock);
-	ret = rhandler->report_event(dev);
+	/*
+	 * If the device is in recovery, will skip reporting the event
+	 * this time because the event should be report when the device
+	 * is reopening.
+	 */
+	if (rdev->flags & (TCMUR_DEV_FLAG_IN_RECOVERY |
+			   TCMUR_DEV_FLAG_STOPPING |
+			   TCMUR_DEV_FLAG_STOPPED)) {
+		pthread_mutex_unlock(&rdev->rdev_lock);
+		return;
+	}
+
+	rdev->flags |= TCMUR_DEV_FLAG_REPORTING_EVENT;
+	has_lock = rdev->lock_state == TCMUR_DEV_LOCK_WRITE_LOCKED;
+	pthread_mutex_unlock(&rdev->rdev_lock);
+
+	ret = rhandler->report_event(dev, has_lock);
 	if (ret)
 		tcmu_dev_err(dev, "Could not report events. Error %d.\n", ret);
+
+	pthread_mutex_lock(&rdev->rdev_lock);
+	rdev->flags &= ~TCMUR_DEV_FLAG_REPORTING_EVENT;
+	pthread_cond_signal(&rdev->report_event_cond);
 	pthread_mutex_unlock(&rdev->rdev_lock);
 }
 
--- a/tcmur_device.h
+++ b/tcmur_device.h
@@ -22,6 +22,7 @@
 #define TCMUR_DEV_FLAG_IS_OPEN		(1 << 2)
 #define TCMUR_DEV_FLAG_STOPPING		(1 << 3)
 #define TCMUR_DEV_FLAG_STOPPED		(1 << 4)
+#define TCMUR_DEV_FLAG_REPORTING_EVENT	(1 << 5)
 
 #define TCMUR_UA_DEV_SIZE_CHANGED	0
 
@@ -82,6 +83,8 @@ struct tcmur_device {
 	pthread_mutex_t format_lock; /* for atomic format operations */
 
 	int cmd_time_out;
+
+	pthread_cond_t report_event_cond;
 
 	pthread_spinlock_t cmds_list_lock; /* protects cmds_list */
 	struct list_head cmds_list;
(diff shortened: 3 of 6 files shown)


rbd.c
```diff
@@ -115,11 +115,12 @@ static darray(char *) blacklist_caches;
 #ifdef LIBRADOS_SUPPORTS_SERVICES
 
 #ifdef RBD_LOCK_ACQUIRE_SUPPORT
+/* rdev->priv_lock is held_*/
```
idryomov, line 118:
```suggestion
/* rdev->priv_lock is held */
```
here and everywhere else

rbd.c
```diff
@@ -115,11 +115,12 @@ static darray(char *) blacklist_caches;
...
 static int tcmu_rbd_service_status_update(struct tcmu_device *dev,
 					  bool has_lock)
 {
-	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
 	struct tcmur_device *rdev = tcmu_dev_get_private(dev);
+	struct tcmu_rbd_state *state = tcmur_dev_get_private(dev);
```
idryomov, line 123: This is a purely cosmetic change, right?

rbd.c
```diff
@@ -238,9 +251,10 @@ static int tcmu_rbd_service_register(struct tcmu_device *dev)
 }
 
 #ifdef RBD_LOCK_ACQUIRE_SUPPORT
-static void tcmu_rbd_service_status_update(struct tcmu_device *dev,
+static int tcmu_rbd_service_status_update(struct tcmu_device *dev,
 					   bool has_lock)
 {
+	return 0;
```
idryomov, line 257: Squash "rbd: remove possible warning" commit into an earlier commit that made the void -> int change.

idryomov (commented):
I'm confused about the scope of the issue.  The description says that the problem is that the status update handler could encounter a NULL state but the PR goes on to change all handlers, including I/O handers such as read and write.  Would tcmu-runner core really initiate I/O on a closed image?

Could we instead look at something like `TCMUR_DEV_FLAG_IS_OPEN`?  Wrapping all handlers with `rdev->priv_lock` seems too heavyweight to me.

idryomov:
I would suggest splitting unrelated bug fixes ("rbd: fix use-after-free of addr", "rbd: fix memory leak when fails to get the address" and "rbd: fix and add more debug logs") into a separate PR.

rbd.c
lxbsz: This maybe introduced by my previous changes and there should be some other changes in this func but removed again, this could be removed and makes no sense.

rbd.c
lxbsz: Will fix it.

rbd.c
lxbsz: I have split this into a new PR in https://github.com/open-iscsi/tcmu-runner/pull/673.

And also have split some other commits into separate PRs to be more easy to review: https://github.com/open-iscsi/tcmu-runner/pull/672, https://github.com/open-iscsi/tcmu-runner/pull/671 and https://github.com/open-iscsi/tcmu-runner/pull/670.

Please take a look, thanks.

lxbsz:
> I'm confused about the scope of the issue. The description says that the problem is that the status update handler could encounter a NULL state but the PR goes on to change all handlers, including I/O handers such as read and write. Would tcmu-runner core really initiate I/O on a closed image?
> 

Yeah, it's possible.

> Could we instead look at something like `TCMUR_DEV_FLAG_IS_OPEN`? Wrapping all handlers with `rdev->priv_lock` seems too heavyweight to me.

Let me check it more about this and have a try.


> I would suggest splitting unrelated bug fixes ("rbd: fix use-after-free of addr", "rbd: fix memory leak when fails to get the address" and "rbd: fix and add more debug logs") into a separate PR.

Done.

lxbsz:
@idryomov Please take a look, thanks.

tcmur_device.c
```diff
@@ -167,11 +170,30 @@ static void __tcmu_report_event(void *data)
...
+	 */
+	if (rdev->flags & (TCMUR_DEV_FLAG_IN_RECOVERY |
+			   TCMUR_DEV_FLAG_STOPPING |
+			   TCMUR_DEV_FLAG_STOPPED)) {
+		pthread_mutex_unlock(&rdev->rdev_lock);
+		return;
+	}
+
+	rdev->flags |= TCMUR_DEV_FLAG_REPORTING_EVENT;
+	pthread_mutex_unlock(&rdev->rdev_lock);
```
idryomov, line 187: Why is `rdev_lock` released here?  `->report_event()` used to be called with it held.

tcmur_device.h
```diff
@@ -75,13 +77,16 @@ struct tcmur_device {
...
 	uint32_t format_progress;
 	pthread_mutex_t format_lock; /* for atomic format operations */
 
 	int cmd_time_out;
+
+	pthread_cond_t report_event_cond;
+
+	pthread_spinlock_t cmds_list_lock; /* protects cmds_list */
 	struct list_head cmds_list;
 };
```
idryomov, line 92: You seem to have checked in a `kmod-devel-25-16.el8.x86_64.rpm` binary by mistake.

idryomov:
I think I'm still missing something.  I'm going to ignore the renames for the moment and speak in terms of what is currently in master.

From my reading of the description, the problem is that `->report_event()` and `__tcmu_reopen_dev()` can race and `->report_event()` might crash on a momentarily closed device.  Now `->report_event()` is currently called under `state_lock` and `__tcmu_reopen_dev()` clears `TCMUR_DEV_FLAG_IS_OPEN` under `state_lock` before proceeding with closing.  Wouldn't it be sufficient to just add a `TCMUR_DEV_FLAG_IS_OPEN` check to `__tcmu_report_event()`?  Why `TCMUR_DEV_FLAG_REPORTING_EVENT` and a new condition variable are necessary?

tcmur_device.c
```diff
@@ -167,11 +170,30 @@ static void __tcmu_report_event(void *data)
...
+
 	ret = rhandler->report_event(dev);
 	if (ret)
 		tcmu_dev_err(dev, "Could not report events. Error %d.\n", ret);
-	pthread_mutex_unlock(&rdev->state_lock);
+
+	pthread_mutex_unlock(&rdev->rdev_lock);
```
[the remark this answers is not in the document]
lxbsz: > Was this code path tested?


Yeah, tested, checked it this has been fixed but wasn't amended to it. 

I found in my setups one node has fixed this, another didn't.

For local test I didn't hit any issue.

tcmur_device.c
lxbsz: The `rdev_lock` should always be released when calling the handler's hooks. I think we need to pass a `has_lock` boolean parameter.

lxbsz:
> I think I'm still missing something. I'm going to ignore the renames for the moment and speak in terms of what is currently in master.
> 
> From my reading of the description, the problem is that `->report_event()` and `__tcmu_reopen_dev()` can race and `->report_event()` might crash on a momentarily closed device. Now `->report_event()` is currently called under `state_lock` and `__tcmu_reopen_dev()` clears `TCMUR_DEV_FLAG_IS_OPEN` under `state_lock` before proceeding with closing. Wouldn't it be sufficient to just add a `TCMUR_DEV_FLAG_IS_OPEN` check to `__tcmu_report_event()`? Why `TCMUR_DEV_FLAG_REPORTING_EVENT` and a new condition variable are necessary?

The reopen and event report will be run in two different threads.

The reopen will be split into `handler->close()` and handler->open()` without holding the `state_lock`, that means the reopen is none atomic.

In case that just after the `__tcmu_report_event()` checked that the `TCMUR_DEV_FLAG_IS_OPEN` flag is set and then it will try to access to the `rbd_state` private data to report the events to ceph cluster. While at the same time the reopen thread could be fired, which will clear the `TCMUR_DEV_FLAG_IS_OPEN` flag and release the `rbd_state` private data in `handler->close()`.

The use-after-free bug still exists...

We need to let the reopen thread wait a bit to be sure that the event report thread has finished.

lxbsz:
Run the following test for 2 hours, worked fine for me.

```
[root@client ~]# lsblk 
NAME        MAJ:MIN RM  SIZE RO TYPE  MOUNTPOINT
sda           8:0    0   25G  0 disk  
├─sda1        8:1    0    1G  0 part  /boot
└─sda2        8:2    0   24G  0 part  
  ├─cl-root 253:0    0 21.9G  0 lvm   /
  └─cl-swap 253:1    0  2.1G  0 lvm   [SWAP]
sdb           8:16   0    1M  0 disk  
└─mpathd    253:2    0    1M  0 mpath 
sdc           8:32   0    1M  0 disk  
└─mpathd    253:2    0    1M  0 mpath 
sr0          11:0    1  8.6G  0 rom   

# while [ 1 ]; do dd if=/dev/zero of=/dev/sdb bs=1K count=1024; sleep 1; dd if=/dev/zero of=/dev/sdc bs=1K count=1024; done
```

tcmur_device.h
lxbsz: Yeah, I have removed it.

idryomov:
> The reopen and event report will be run in two different threads.
> 
> The reopen will be split into `handler->close()` and `handler->open()` without holding the `state_lock`, that means the reopen is none atomic.
> 
> In case that just after the `__tcmu_report_event()` checked that the `TCMUR_DEV_FLAG_IS_OPEN` flag is set and then it will try to access to the `rbd_state` private data to report the events to ceph cluster. While at the same time the reopen thread could be fired, which will clear the `TCMUR_DEV_FLAG_IS_OPEN` flag and release the `rbd_state` private data in `handler->close()`.

In master, `state_lock`  is held across `->report_event()` so if while reopen is indeed not atomic, event report is.  `__tcmu_reopen_dev()` would not be able to clear `TCMUR_DEV_FLAG_IS_OPEN` and close the image while `->report_event()` is executing.

You are removing `state_lock` protection from `->report_event()` though, saying that it (now renamed to `rdev_lock`) "should always be released when calling the handler's hooks".  Can you elaborate on why it is not OK to hold it across `->report_event()` as it is currently done?

lxbsz:
> > The reopen and event report will be run in two different threads.
> > The reopen will be split into `handler->close()` and `handler->open()` without holding the `state_lock`, that means the reopen is none atomic.
> > In case that just after the `__tcmu_report_event()` checked that the `TCMUR_DEV_FLAG_IS_OPEN` flag is set and then it will try to access to the `rbd_state` private data to report the events to ceph cluster. While at the same time the reopen thread could be fired, which will clear the `TCMUR_DEV_FLAG_IS_OPEN` flag and release the `rbd_state` private data in `handler->close()`.
> 
> In master, `state_lock` is held across `->report_event()` so if while reopen is indeed not atomic, event report is. `__tcmu_reopen_dev()` would not be able to clear `TCMUR_DEV_FLAG_IS_OPEN` and close the image while `->report_event()` is executing.
> 
> You are removing `state_lock` protection from `->report_event()` though, saying that it (now renamed to `rdev_lock`) "should always be released when calling the handler's hooks". Can you elaborate on why it is not OK to hold it across `->report_event()` as it is currently done?

As I remembered long time ago as discussed, the rule is that the `state_lock` should be only used in the libtcmu, and shouldn't be used in the handlers. And when calling any handler hook we should release the `state_lock`, because there may have third part handlers could call libtcmu's helpers, which will be possibly acquire the `state_lock` again, potentially will introduce dead lock bug.

The current code in master is buggy when begin to support event report feature.

PR #669, opened 2021-09-06
main: fix g_object_unref: assertion 'G_IS_OBJECT (object)' failed

lxbsz:
If the 'manager' is not initialized just skip it.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1398,7 +1398,7 @@ int main(int argc, char **argv)
 	    g_unix_signal_add(SIGTERM, handle_sig, loop) <= 0 ||
 	    g_unix_signal_add(SIGHUP, handle_sighup, loop) <= 0) {
 		tcmu_err("couldn't setup signal handlers\n");
-		goto unwatch_cfg;
+		goto loop_unref;
 	}
 
 	/* Set up event for libtcmu */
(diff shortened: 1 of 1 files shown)


idryomov:
So this is because `G_IS_OBJECT(NULL)` returns false and `g_object_unref()` does `g_return_if_fail (G_IS_OBJECT (object))` at the top:
```
void
g_object_unref (gpointer _object)
{
  GObject *object = _object;
  gint old_ref;

  g_return_if_fail (G_IS_OBJECT (object));
```

idryomov (approved):
LGTM.  It might be worth a note in the commit message that you are also fixing a potential `loop` reference leak in case adding signal handlers fails.

lxbsz:
> So this is because `G_IS_OBJECT(NULL)` returns false and `g_object_unref()` does `g_return_if_fail (G_IS_OBJECT (object))` at the top:
> 
> ```
> void
> g_object_unref (gpointer _object)
> {
>   GObject *object = _object;
>   gint old_ref;
> 
>   g_return_if_fail (G_IS_OBJECT (object));
> ```

Yeah, it is.

> LGTM. It might be worth a note in the commit message that you are also fixing a potential `loop` reference leak in case adding signal handlers fails.

Sure, will fix it.

lxbsz:
Only changed the commit comments, merge it. Thanks @idryomov

PR #670, opened 2021-09-06
rbd: fix use-after-free of addr

lxbsz:


idryomov (approved):
I would probably do it with a label but this is fine too.

lxbsz:
> I would probably do it with a label but this is fine too.

Okay, thanks @idryomov

PR #672, opened 2021-09-06
rbd: fix memory leak when fails to get the address

lxbsz:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required (VERSION 2.8 FATAL_ERROR)
 project (tcmu-runner C)
-set(VERSION 1.5.1)
+set(VERSION 1.5.2)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror -Wall -Wdeclaration-after-statement -std=c99")
 
 include(GNUInstallDirs)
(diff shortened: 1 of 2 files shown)


rbd.c
```diff
@@ -1159,9 +1159,11 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 #ifdef LIBRADOS_SUPPORTS_GETADDRS
 	/* Get current entry address for the image */
 	ret = rados_getaddrs(state->cluster, &state->addrs);
-	tcmu_dev_info(dev, "address: {%s}\n", state->addrs);
-	if (ret < 0)
-		return ret;
+	if (ret < 0) {
+		tcmu_dev_err(dev, "failed getting the address\n");
```
idryomov, line 1163: Nit: for consistency with other error messages in this function
```suggestion
		tcmu_dev_err(dev, "Could not get address.\n");
```

rbd.c
```diff
@@ -1159,9 +1159,11 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
...
+		goto stop_image;
+	}
+	tcmu_dev_info(dev, "opened, the address: {%s}\n", state->addrs);
```
idryomov, line 1166: Nit: keep the old string as the image is opened unconditionally whereas this message is printed only if `rados_getaddrs()` is supported (nautilus and later).
```suggestion
	tcmu_dev_info(dev, "address: {%s}\n", state->addrs);
```

rbd.c
lxbsz: Will fix it.

rbd.c
lxbsz: Will fix it.

lxbsz:
Closed this one by accident, and have update it in a new PR https://github.com/open-iscsi/tcmu-runner/pull/675.

Thanks @idryomov

PR #674, opened 2021-09-13
Revert "Create codeql-analysis.yml"

lxbsz:
This reverts commit f756a5549026b975289826c48d38646e59865f65.


lxbsz:
Revert this for now, it won't work well.

PR #676, opened 2021-09-14
rbd: switch strtok to strtok_r

lxbsz:
The strtok is not thread safe, there could be several threads will
call the tcmu_rbd_open() at the same time.

### DIFF
--- a/rbd.c
+++ b/rbd.c
@@ -1026,8 +1026,9 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 	char *config, *dev_cfg_dup;
 	struct tcmu_rbd_state *state;
 	uint32_t max_blocks, unmap_gran;
-	int ret;
 	char buf[128];
+	char *savedptr = NULL;
+	int ret;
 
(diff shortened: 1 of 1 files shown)


rbd.c
```diff
@@ -1026,8 +1026,9 @@ static int tcmu_rbd_open(struct tcmu_device *dev, bool reopen)
 	char *config, *dev_cfg_dup;
 	struct tcmu_rbd_state *state;
 	uint32_t max_blocks, unmap_gran;
-	int ret;
 	char buf[128];
+	char *savedptr;
```
idryomov, line 1030: I know it doesn't matter for `strtok_r` but it's a good idea to initialize anyway as some compilers and static checkers would complain.
```suggestion
	char *saveptr = NULL;
```

idryomov (commented):
Might be good to include an example log snippet in the commit message.

rbd.c
lxbsz: Done.

idryomov (approved):

lxbsz:
Thanks @idryomov , will merge it.

Issue #654, opened 2021-04-09
Add support for mark as write protected


liulanzheng:
I need to mark a device write protected or not depending on what file is opened. But i didn't find an api to set the write protected bit in libtcmu.

lxbsz:
Yeah, it doesn't implement the SWP bit in the mode sense control page.

You can implement it in `handle_control_page()` if you want, which is similar to `dev->write_cache_enabled`.

liulanzheng:
@lxbsz Thanks for you advice, I implemented it and worked. But I am not quite familiar with iSCSI protocol and 'mode sense control page', I didn't find how to implement it in handle_control_page(). At last, I set  the SWP bit in the 'DEVICE-SPECIFIC PARAMETER' field in the 'mode parameter header'.  I am not sure whether it is correct even thong it worked.

lxbsz:
Yeah, set the `wp` bit in `DEVICE-SPECIFIC PARAMETER` should be enough.
For the spec, you can refer to https://origin-www.seagate.com/files/staticfiles/support/docs/manual/Interface%20manuals/100293068j.pdf

PR #680, opened 2022-02-18
libtcmu: add support for write protect

liulanzheng:
As discussed in #654, add an api to set write-protect for dev, and set the corresponding parameter when mode sense.

Signed-off-by: liulanzheng <lanzheng.liulz@alibaba-inc.com>

### DIFF
--- a/libtcmu.c
+++ b/libtcmu.c
@@ -1021,6 +1021,16 @@ bool tcmu_dev_get_unmap_enabled(struct tcmu_device *dev)
 	return dev->unmap_enabled;
 }
 
+void tcmu_dev_set_write_protect_enabled(struct tcmu_device *dev, bool enabled)
+{
+	dev->write_protect_enabled = enabled;
+}
+
+bool tcmu_dev_get_write_protect_enabled(struct tcmu_device *dev)
+{
+	return dev->write_protect_enabled;
+}
+
 int tcmu_dev_get_fd(struct tcmu_device *dev)
 {
 	return dev->fd;
@@ -1060,7 +1070,7 @@ tcmu_dev_get_memory_info(struct tcmu_device *dev, void **base,
 			/* get length of map from file */
 			ssize_t size;
 			char *size_name;
-			
+
 			if (asprintf(&size_name, sizefmt, dev->dev_name) == -1) {
 				tcmu_err("cannot construct device map size filename\n");
 				goto err_free;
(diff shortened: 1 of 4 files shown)


lxbsz:
Do you have any use case for this ?

liulanzheng:
@lxbsz  in [overlaybd](https://github.com/containerd/overlaybd), we output container images as block device through libtcmu for data acceleration. Images, like overlaybd containerd images, and vm images, also iso images, are readonly, it is better to provided a method to mark the device as readonly fundamentally, which is significantly better than executing blockdev --setro command on client. 
At first, we use tgtd as the backstore and it has readonly option for target configuration. So i think adding an option for write-protect/readonly is a real need.

lxbsz:
BTW, which spc or sbc you are referring to for `orig_buf[3]` and `orig_buf[2]` ? From the sbc for MODE_SENSE_10 the first 8 bytes are reserved only for the `NUMBER OF LOGICAL BLOCKS`, and for MODE_SENSE there are 4 bytes.

lxbsz:
> BTW, which spc or sbc you are referring to for `orig_buf[3]` and `orig_buf[2]` ? From the sbc for MODE_SENSE_10 the first 8 bytes are reserved only for the `NUMBER OF LOGICAL BLOCKS`, and for MODE_SENSE there are 4 bytes.

Okay, I misread the docs and I see it.

libtcmu.c
```diff
@@ -1021,6 +1021,16 @@ bool tcmu_dev_get_unmap_enabled(struct tcmu_device *dev)
 	return dev->unmap_enabled;
 }
 
+bool tcmu_dev_set_write_protect_enabled(struct tcmu_device *dev, bool enabled)
```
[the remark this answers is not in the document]
liulanzheng: sorry, i remember i fixed it, but ...   it's fixed now.

libtcmu_priv.h
```diff
@@ -55,6 +55,7 @@ struct tcmu_device {
 	unsigned int write_cache_enabled:1;
 	unsigned int solid_state_media:1;
 	unsigned int unmap_enabled:1;
+	unsigned int write_protect_enabled: 1;
```
lxbsz, line 58: Please remove the white space before `1`.

liulanzheng: done.

lxbsz (approved):
LGTM. Thanks.

PR #682, opened 2022-06-30
Use get_unaligned_be*() to avoid UB

dmitryash:
In most cases using strict aliasing is undefined behaviour.
Obviously cdb is supposed to be an array of bytes and accessing
any CDB field using strict aliasing leads to misaligned read which
may cause even a crash.

See [here](https://en.cppreference.com/w/c/language/object#Strict_aliasing)

### DIFF
--- a/api.c
+++ b/api.c
@@ -67,14 +67,14 @@ uint64_t tcmu_cdb_get_lba(uint8_t *cdb)
 
 	switch (tcmu_cdb_get_length(cdb)) {
 	case 6:
-		val = be16toh(*((uint16_t *)&cdb[2]));
+		val = get_unaligned_be16(&cdb[2]);
 		return ((cdb[1] & 0x1f) << 16) | val;
 	case 10:
-		return be32toh(*((u_int32_t *)&cdb[2]));
+		return get_unaligned_be32(&cdb[2]);
 	case 12:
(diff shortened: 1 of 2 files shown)


lxbsz:
@dmitryash 

Could you add the `signed-of-by:` and add some comments in commit ? And what the `UB` means here ?
Thanks!

dmitryash:
@lxbsz Hi, I added the requested information, review please.

api.c
```diff
@@ -87,11 +87,11 @@ uint32_t tcmu_cdb_get_xfer_length(uint8_t *cdb)
...
 		return cdb[4];
 	case 10:
-		return be16toh(*((uint16_t *)&cdb[7]));
+		return get_unaligned_be16(&cdb[7]));
 	case 12:
-		return be32toh(*((u_int32_t *)&cdb[6]));
+		return get_unaligned_be32(&cdb[6]));
 	case 16:
-		return be32toh(*((u_int32_t *)&cdb[10]));
+		return get_unaligned_be32(&cdb[10]));
```
lxbsz, line 94: Did you ever compile this change ? Please remove the last `)`.

dmitryash: Now it compiles. Checked compilation.

lxbsz:
LGTM. Thanks!

PR #691, opened 2023-01-18
Match exactly against cfgstring

geordieintheshellcode:
Two bug fixes:
1. When searching for a suitable handler in `find_handler` ensure that `handler->subtype` and the config string match exactly. The previous code would erroneously match config strings which were a prefix match of `handler->subtype`. For example, the config string "foo" would erroneously match against the subtype "foobar".
2. This is slightly more paranoid. Ensure that when matching the config string against the expected prefix ("tcm-user") the strings match exactly. The previous code would erroneously match the config string "tcm-userXXX" against "tcm-user".


geordieintheshellcode:
Wondering if someone is able to take a look at this PR please?

lxbsz:
And also for this could you fix the commit comments as in https://github.com/open-iscsi/tcmu-runner/pull/693 ?

Thanks,

libtcmu.c
```diff
@@ -438,9 +440,10 @@ static bool device_parse_cfg(struct tcmu_device *dev,
 	/* Check valid cfgstring */
 	oldptr = cfgstring;
 	ptr = strchr(oldptr, '/');
+	cfgstrlen = ptr-oldptr;
```
lxbsz, line 443: Why not just use the `len` instead of adding a new `cfgstrlen` ?

geordieintheshellcode: Done

geordieintheshellcode:
Comments addressed and commit comment fixed.

Issue #678, opened 2021-11-20
Feature request and offer to PR: provide a "no netlink" mode for libtcmu


geordieintheshellcode:
Some folks might be using libtcmu in their own applications which don't hook into TCMU runner, much like the `consumer.c` example. In those cases, it can be useful to avoid using netlink altogether. As I'm sure you are aware, it's almost impossible to have more than one user of TCMU when netlink is involved. This is because the kernel broadcasts device events to all TCMU listeners over netlink and simply accepts the first response it receives. This can lead to the following scenario:

1. Application A creates a TCMU backstore with handler=app_a
2. The kernel (target_core_user) broadcasts the netlink message to all instances of libtcmu
3. Application B receives the message first. Application B doesn't have a handler for app_a, so it replies with an error code to the kernel.
4. The kernel receives the error and aborts backstore creation.
5. Application A wonders why its backstore creation failed.

Users can avoid using netlink altogether by setting `nl_reply_supported=-1` at backing store creation time. This is great as multiple applications using TCMU can now coexist on the same machine. The problem is we don't have a nice way to utilise libtcmu in this model. We need a way to notify libtcmu that a device has been created, reconfigured, removed so the relevant handlers can be called. Currently, this is done via netlink.

Would you accept a PR that allows users to utilise all the great features of libtcmu, but without using netlink? The PR would be roughly as follows:
* modify `tcmulib_initialize` to take a boolean parameter which informs libtcmu whether or not it should set up netlink. If `true` we do. If `false` we don't.
* expose new methods `tcmulib_notify_device_added`, `tcmulib_notify_device_removed` and `tcmulib_notify_device_reconfiged`, which application code should call to inform libtcmu that a device has been added, removed, or reconfigured. These API's take the `dev_name` and pass through to the existing `device_add`, `device_remove` and `device_reconfig` functions in libtcmu.c. `tcmulib_notify_device_reconfiged` will need additional parameters to specify the new config.
* updates the documentation to reflect these changes and notes that `tcmulib_master_fd_ready` and `tcmulib_get_master_fd` should not be called if you chose to call `tcmulib_initialize` in "don't use netlink mode".

This change benefits all users of libtcmu who are managing their own devices and who would like to be able to run multiple instances of their app on one machine and/or coexist with other users of libtcmu/tcmu-runner. 

Great work with the library and let me know if this feature sounds interesting!

lxbsz:
@geordieintheshellcode 

This sounds reasonable to me.

[...]

PR #694, opened 2023-01-19
New feature - option to use libtcmu without netlink

geordieintheshellcode:
Some folks might be using libtcmu in their own applications which don't hook into TCMU runner, much like the `consumer.c` example. In those cases, it can be useful to avoid using netlink altogether. As I'm sure you are aware, it's almost impossible to have more than one user of TCMU when netlink is involved. This is because the kernel broadcasts device events to all TCMU listeners over netlink and simply accepts the first response it receives. This can lead to the following scenario:

1. Application A creates a TCMU backstore with handler=app_a
2. The kernel (target_core_user) broadcasts the netlink message to all instances of libtcmu
3. Application B receives the message first. Application B doesn't have a handler for app_a, so it replies with an error code to the kernel.
4. The kernel receives the error and aborts backstore creation.
5. Application A wonders why its backstore creation failed.

Users can avoid using netlink altogether by setting `nl_reply_supported=-1` at backing store creation time. This is great as multiple applications using TCMU can now coexist on the same machine. The problem is we don't have a nice way to utilise libtcmu in this model. We need a way to notify libtcmu that a device has been created, reconfigured, removed so the relevant handlers can be called. Currently, this is done via netlink.

This PR allows users to utilise all the great features of libtcmu, but without using netlink. The PR makes the following changes:
* modify `tcmulib_initialize` to take a boolean parameter which informs libtcmu whether or not it should set up netlink. If `true` we do. If `false` we don't.
* expose new methods `tcmulib_notify_device_added`, `tcmulib_notify_device_removed` and `tcmulib_notify_device_reconfiged`, which application code should call to inform libtcmu that a device has been added, removed, or reconfigured. These API's take the `dev_name` and pass through to the existing `device_add`, `device_remove` and `device_reconfig` functions in libtcmu.c. `tcmulib_notify_device_reconfiged` will need additional parameters to specify the new config.
* updates the documentation to reflect these changes and notes that `tcmulib_master_fd_ready` and `tcmulib_get_master_fd` should not be called if you chose to call `tcmulib_initialize` in "don't use netlink mode".
* adds an example application `consumer_no_netlink` which shows how to create a standalone application using libtcmu without netlink.

Initial discussion of this feature: https://github.com/open-iscsi/tcmu-runner/issues/678


geordieintheshellcode:
Relates to https://github.com/open-iscsi/tcmu-runner/pull/519,

lxbsz:
@geordieintheshellcode 

Could you add the detail comments and signed-off-by in the commit ?

main.c
```diff
@@ -1380,7 +1381,7 @@ int main(int argc, char **argv)
 		darray_append(handlers, tmp_handler);
 	}
 
-	tcmulib_context = tcmulib_initialize(handlers.item, handlers.size);
+	tcmulib_context = tcmulib_initialize(handlers.item, handlers.size, use_netlink);
```
lxbsz, line 1384: You can just discard the `use_netlink` and set it to `true` here.

tcmu-synthesizer.c
```diff
@@ -197,7 +198,7 @@ int main(int argc, char **argv)
 		exit(1);
 	}
 
-	ctx = tcmulib_initialize(&syn_handler, 1);
+	ctx = tcmulib_initialize(&syn_handler, 1, use_netlink);
```
lxbsz, line 201: Same here.

consumer_no_netlink.c
```diff
@@ -0,0 +1,299 @@
...
+
+	if (tcmu_setup_log(LOG_DIR)) {
+		fprintf(stderr, "Could not setup tcmu logger.\n");
+		exit(1);
+	}
+
+	/* If any TCMU devices that exist that match subtype,
+	   handler->added() will now be called from within
+	   tcmulib_initialize(). */
+	tcmulib_ctx = tcmulib_initialize(&foo_handler, 1, use_netlink);
```
lxbsz, line 223: Same here.

consumer_no_netlink.c
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	fgets(buffer, size, f);
+	pclose(f);
+	return true;
+}
+
+static bool create_backstore()
+{
+	int ret = system("./create_backstore.sh");
```
lxbsz, line 105: This seems incorrect approach.
Why not comply to the current `consumer.c`, which will create the devices via targetcli or ceph-iscsi ?

main.c
geordieintheshellcode: Done

consumer_no_netlink.c
geordieintheshellcode: Done

consumer_no_netlink.c
geordieintheshellcode: For one, I don't believe `targetcli` has the ability to specify `nl_reply_supported=-1` when creating the backstore. This is vital if you want to use TCMU without netlink. Secondly, I think there's merit in having a completely stand-alone example which doesn't require the user to have external tools installed.

geordieintheshellcode:
> @geordieintheshellcode
> 
> Could you add the detail comments and signed-off-by in the commit ?

Done.

geordieintheshellcode:
Fixed commit comments

gorbak25:
+1 For my use-case i need for multiple libtcmu instances in different containers to coexist. @geordieintheshellcode thank you for your contribution. I will port your PR to my fork for now https://github.com/hocus-dev/photon-libtcmu

PR #695, opened 2023-03-20
tcmur_device: fix racy between reopening devices and reporting events

lxbsz:
To make sure all the in-flight IOs have been finished before flushing the event_work. Or just after we flush the event_work a new timedout IO callback could come and then it will fire a new event work, which may access the device while we are closing the device later.

### DIFF
--- a/target.c
+++ b/target.c
@@ -339,6 +339,7 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
 
 add_to_list:
 	list_add(&tpg->devs, &rdev->recovery_entry);
+	rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
 done:
 	tcmu_release_alua_grps(&alua_list);
 	pthread_mutex_unlock(&tpg_recovery_lock);
--- a/tcmur_device.c
+++ b/tcmur_device.c
@@ -47,23 +47,11 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
 	bool needs_close = false;
 
 	pthread_mutex_lock(&rdev->rdev_lock);
+	assert(rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY);
 	if (rdev->flags & TCMUR_DEV_FLAG_STOPPING) {
 		ret = 0;
 		goto done;
 	}
-	pthread_mutex_unlock(&rdev->rdev_lock);
-
-	/*
-	 * There are no SCSI commands running but there may be
-	 * async lock requests in progress that might be accessing
-	 * the device.
-	 */
-	tcmur_flush_work(rdev->event_work);
-
-	pthread_mutex_lock(&rdev->rdev_lock);
-	if (rdev->flags & TCMUR_DEV_FLAG_IS_OPEN)
-		needs_close = true;
-	rdev->flags &= ~TCMUR_DEV_FLAG_IS_OPEN;
 	pthread_mutex_unlock(&rdev->rdev_lock);
 
 	if (pthread_self() != rdev->cmdproc_thread)
@@ -74,12 +62,27 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
 		 */
 		tcmu_dev_flush_ring(dev);
 
+	/*
+	 * To make sure all the in-flight IOs have been finished before
+	 * flushing the event_work. Or just after we flush the event_work
+	 * a new timedout IO callback could come and then it will fire a
+	 * new event work, which may access the device while we are closing
+	 * the device later.
+	 */
 	tcmu_dev_dbg(dev, "Waiting for outstanding commands to complete\n");
 	ret = aio_wait_for_empty_queue(rdev);
 	if (ret) {
 		pthread_mutex_lock(&rdev->rdev_lock);
 		goto done;
 	}
+
+	tcmur_flush_work(rdev->event_work);
+
+	pthread_mutex_lock(&rdev->rdev_lock);
+	if (rdev->flags & TCMUR_DEV_FLAG_IS_OPEN)
+		needs_close = true;
+	rdev->flags &= ~TCMUR_DEV_FLAG_IS_OPEN;
+	pthread_mutex_unlock(&rdev->rdev_lock);
 
 	if (needs_close) {
 		tcmu_dev_dbg(dev, "Closing device.\n");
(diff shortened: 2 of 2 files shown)


tcmur_device.c
```diff
@@ -74,13 +61,28 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
...
 		goto done;
 	}
 
+	tcmur_flush_work(rdev->event_work);
+
+	pthread_mutex_lock(&rdev->rdev_lock);
+	if (rdev->flags & TCMUR_DEV_FLAG_IS_OPEN)
+		needs_close = true;
+	rdev->flags &= ~TCMUR_DEV_FLAG_IS_OPEN;
+	pthread_mutex_unlock(&rdev->rdev_lock);
```
pkalever, line 84: @lxbsz 
Can we close the handler under this lock, so that there are zero windows for any IO to happen?

idryomov: The lock isn't held across IO itself, so holding it for handler's `close` method would seem pointless.  The important thing is to ensure that nothing new can start executing in the interim and that appears to be achieved via flags.

lxbsz: Yeah, agree with Ilya.

tcmur_device.c
```diff
@@ -74,13 +61,28 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
...
 
+	/*
+	 * To make sure all the in-flight IOs have been finished before
+	 * flushing the event_work. Or just after we flush the event_work
+	 * a new timedout IO callback could come and then it will fire a
+	 * new event work, which may access the device while we are closing
+	 * the device later.
+	 */
 	tcmu_dev_dbg(dev, "Waiting for outstanding commands to complete\n");
 	ret = aio_wait_for_empty_queue(rdev);
```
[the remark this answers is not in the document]
lxbsz: There are two ways to do that: 
1, `TCMUR_DEV_FLAG_IN_RECOVERY`: then all the cmds in the ring buffer will be cleaned by just returning `-EBUSY` without not doing the real IOs to handlers.

So just after `aio_wait_for_empty_queue()`, we are safe to say that all the in-flight IOs have finished.

2, There is one interface to tell the kernel LIO/TCMU do not queue new cmds to the ring buffer and then in userland we can just clean the ring buffer. And then call `aio_wait_for_empty_queue()`.


Here we will use the first one.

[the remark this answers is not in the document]
idryomov: > So just after `aio_wait_for_empty_queue()`, we are safe to say that all the in-flight IOs have finished.

I did some grepping to confirm that this is the case.  It appears to be but I have two suggestions:
- set `TCMUR_DEV_FLAG_IN_RECOVERY` in `tcmu_add_dev_to_recovery_list` where the device is added to recovery list instead of conditionally in `__tcmu_notify_conn_lost`:
```
add_to_list:
	list_add(&tpg->devs, &rdev->recovery_entry);
	rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
```
- replace "TCMUR_DEV_FLAG_IN_RECOVERY must be set before calling" comment with an assert as this is critical requirement -- crashing on a NULL dereference somewhere in librados is much worse:
```
	bool needs_close = false;

	pthread_mutex_lock(&rdev->rdev_lock);
	assert(rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY);
	if (rdev->flags & TCMUR_DEV_FLAG_STOPPING) {
		ret = 0;
		goto done;
```
What do you think?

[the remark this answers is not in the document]
lxbsz: > > So just after `aio_wait_for_empty_queue()`, we are safe to say that all the in-flight IOs have finished.
> 
> I did some grepping to confirm that this is the case. It appears to be but I have two suggestions:
> 
> * set `TCMUR_DEV_FLAG_IN_RECOVERY` in `tcmu_add_dev_to_recovery_list` where the device is added to the recovery list instead of conditionally in `__tcmu_notify_conn_lost`:
> 
> ```
> add_to_list:
> 	list_add(&tpg->devs, &rdev->recovery_entry);
> 	rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
> ```
> 

Yeah, this looks okay for me.


> * replace "TCMUR_DEV_FLAG_IN_RECOVERY must be set before calling" comment with an assert as this is critical requirement -- crashing on a NULL dereference somewhere in librados is much worse:
> 
> ```
> 	bool needs_close = false;
> 
> 	pthread_mutex_lock(&rdev->rdev_lock);
> 	assert(rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY);
> 	if (rdev->flags & TCMUR_DEV_FLAG_STOPPING) {
> 		ret = 0;
> 		goto done;
> ```
> 
> What do you think?

This also makes sense to me.

I will revise this. thanks.

tcmur_device.c
```diff
@@ -46,26 +46,15 @@ int __tcmu_reopen_dev(struct tcmu_device *dev, int retries)
 	int ret, attempt = 0;
 	bool needs_close = false;
 
+	assert(rdev->flags & TCMUR_DEV_FLAG_IN_RECOVERY);
```
idryomov, line 49: Shouldn't this be done with `rdev->rdev_lock` held (i.e. on line 50)?

target.c
```diff
@@ -336,11 +336,14 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
 		goto done;
 	}
 	list_add(&tpg_recovery_list, &tpg->recovery_entry);
+	rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
```
idryomov, line 339: I think this should be done after adding `rdev` to `tpg->devs` list:
```
add_to_list: 
        list_add(&tpg->devs, &rdev->recovery_entry);
        rdev->flags |= TCMUR_DEV_FLAG_IN_RECOVERY;
```
Otherwise `TCMUR_DEV_FLAG_IN_RECOVERY` wouldn't get set in the "another device already kicked off recovery" case (or, if it somehow would, it's really not obvious).

target.c
```diff
@@ -336,11 +336,14 @@ int tcmu_add_dev_to_recovery_list(struct tcmu_device *dev)
...
+	rdev->conn_lost_cnt++;
```
idryomov, line 340: The increment should stay in `__tcmu_notify_conn_lost` IMO.  The point of my suggestion was to keep related things close and not mix them:
- set `TCMUR_DEV_FLAG_IN_RECOVERY` in the function that adds `rdev` to the respective list (using `rdev->recovery_entry`)
- increment `rdev->conn_lost_cnt` in the function that handles a lost connection

tcmur_device.c
lxbsz: This is not that critical IMO. But I will fix it.

target.c
lxbsz: Okay.

idryomov (approved):

PR #696, opened 2023-04-28
tcmur_device: skip reporting events if the device is closed

lxbsz:
After the rbd device being blocklisted it will lost the connection and at the same time will timedout all the IOs. While the tcmu-runner will report a 'lock_lost_cnt' event to ceph cluster. And if during this the device is reopened it will cause the use-after-free bug.

When reopening the device it will hold the 'rdev->rdev_lock' lock and then clear the 'TCMUR_DEV_FLAG_IS_OPEN' flag. And we can check this flag and just skip reporting events if the device is closed.

### DIFF
--- a/tcmur_device.c
+++ b/tcmur_device.c
@@ -171,9 +171,11 @@ static void __tcmu_report_event(void *data)
 	sleep(1);
 
 	pthread_mutex_lock(&rdev->rdev_lock);
-	ret = rhandler->report_event(dev);
-	if (ret)
-		tcmu_dev_err(dev, "Could not report events. Error %d.\n", ret);
+	if (rdev->flags & TCMUR_DEV_FLAG_IS_OPEN) {
+		ret = rhandler->report_event(dev);
+		if (ret)
+			tcmu_dev_err(dev, "Could not report events. Error %d.\n", ret);
+	}
 	pthread_mutex_unlock(&rdev->rdev_lock);
 }
 


tcmur_device.c
```diff
@@ -171,9 +171,11 @@ static void __tcmu_report_event(void *data)
 	sleep(1);
 
 	pthread_mutex_lock(&rdev->rdev_lock);
-	ret = rhandler->report_event(dev);
-	if (ret)
-		tcmu_dev_err(dev, "Could not report events. Error %d.\n", ret);
+	if (rdev->flags & TCMUR_DEV_FLAG_IS_OPEN) {
```
idryomov, line 174: > After the rbd device being blocklisted it will lost the connection and at the same time will timedout all the IOs. While the tcmu-runner will report a 'lock_lost_cnt' event to ceph cluster. And if during this the device is reopened it will cause the use-after-free bug.

Hrm, the events are flushed before the device is closed:
https://github.com/open-iscsi/tcmu-runner/blob/2f20fd95eccc5dc766f24aab007ce837b3e3635b/tcmur_device.c#L65-L90
So this race seems to be already accounted for, according to the comment at least.  Does this mean that something can queue a new event after `tcmur_flush_work()` returns or am I confused?

lxbsz: For example, if the last inflight IO failed due to losing the rbd exclusive lock the `rbd` handler will be triggerred with:

```
      IO cbk thread:                                                           report event thread:                               IO thread:
rbd_finish_aio_generic()                                                      
    --> tcmu_rbd_handle_blocklisted_cmd()
          --> tcmu_notify_lock_lost()
               --> __tcmu_notify_lock_lost()
                    --> tcmu_report_event()
                         --> tcmur_run_work()
                         // it will wake up event_work               __tcmu_report_event()
                         // to fire a dedicated report                     --> sleep(1)
                         // event thread
    --> tcmur_cmd_complete()                                                                                                             tcmur_cmd_handler()
                                                                                                                                                                    -->alua_check_state()
                                                                                                                                                                         -->alua_implicit_transition()
                                                                                                                                                                             --> alua_event_work_fn()
                                                                                                                                                                                 --> tcmu_acquire_dev_lock()
                                                                                                                                                                                     --> __tcmu_reopen_dev()
                                                                                                                                                                                          --> rhandler->close()                                                                                         
                                                                                                --> rhandler->report_event(dev);
                                                                                                                                                                                          --> rhandler->open()

Comment (idryomov) [reply tcmur_device.c:174]: > This is a different case with previous one we fixed, it's racy between deferred report event thread and new IOs.

I understand that but I assumed that
```
	tcmur_flush_work(rdev->event_work);
```
in `__tcmu_reopen_dev()` would block until report event work finishes.  However, looking at what `tcmur_flush_work()` does, I see this:
```
static void __tcmur_flush_work(struct tcmur_work *work)
{
	/*
	 * The event work thread may need to do a handler reopen
	 * call and try to flush itself. Just ignore.
	 */
	if (__tcmu_is_ework_thread)
		return;
```
and `__tcmu_is_ework_thread` is set for _all_ works since:
- the only way to schedule a work is via `tcmur_run_work()`
- `tcmur_run_work()` wraps `work_fn` into `tcmur_work_fn()`
- `tcmur_work_fn()` sets `__tcmu_is_ework_thread` unconditionally

which makes it seem like `tcmur_flush_work()` is basically a no-op (in all three places it's called).  That can't be right?

lxbsz: No, not all cases which call the `tcmu_reopen_dev()` are no-op for `__tcmur_flush_work()`. Such as from `handle_stpg()` --> ... --> `tcmu_reopen_dev()`.

idryomov: > No, not all cases which call the `tcmu_reopen_dev()` are no-op for `__tcmur_flush_work()`. Such as from `handle_stpg()` --> ... --> `tcmu_reopen_dev()`.

I see.  `handle_stpg()` and `handle_rtpg()` are executed by cmdproc thread and through a rather deep chain end up calling `__tcmu_reopen_dev()` such that it's not struck by `__tcmu_is_ework_thread` business.

With that in mind the change makes sense but I have to say that this is _really_ confusing...

idryomov (approved):

tcmur_device.c
lxbsz: Yeah, it is. Thanks @idryomov

PR #701, opened 2024-05-27
tcmur_cmd_handler: To fix the issue on a CentOS 7.9 client where the …

zlxzfg:
…tcmu-runner ver=1.5.4 service faild after using mkfs.xfs to format a LUN device

### DIFF
--- a/tcmur_cmd_handler.c
+++ b/tcmur_cmd_handler.c
@@ -685,7 +685,7 @@ static int handle_unmap_in_writesame(struct tcmu_device *dev,
 	int ret;
 
 	/* If not aligned then falls back to the writesame without unmap */
-	if (lba % align || nlbas % align) {
+	if (align == 0 || lba % align || nlbas % align) {
 		tcmu_dev_dbg(dev,
 			     "Start lba: %"PRIu64" or nlbas: %"PRIu64" not aligned to %"PRIu32"\n",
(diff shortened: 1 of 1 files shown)


zlxzfg:
when use  user:glfs  type create  a lun on tcmu=1.5.4， then  use  mkfs.xfs /dev/sdb format lun device on centos7 (iscsi-initiator-utils-iscsiuio-6.2.0.874-19.el7.x86_64),   result cause  tcmu-runner service core (failed)

tcmur_cmd_handler.c
```diff
@@ -685,7 +685,7 @@ static int handle_unmap_in_writesame(struct tcmu_device *dev,
 	int ret;
 
 	/* If not aligned then falls back to the writesame without unmap */
-	if (lba % align || nlbas % align) {
+	if (align == 0 || lba % align || nlbas % align) {
```
xiubli, line 688: Do you know in which case will the `align` be `0` ? Could you enable the debug logs and reproduce it and then upload the logs ?

Thanks

xiubli: Okay, please ignore this. In the later comment you said you are using the `glfs`, not rbd.

lxbsz:
@zlxzfg Could you test and check whether should we fix it in `align_and_split_unmap()` ?

PR #711, opened 2025-08-08
libtcmu: Fix memory accounting

krnowak:
The accounting for used memory in `data` was off by 4, as it seemed to be thought as relative to `&data[4]`, but then it used to calculate how many bytes in `data` are still available, and the result was passed to `snprintf`. Glibc with some fortification flags enabled has its own way of figuring out how many bytes are really left in the buffer and bails out with a buffer overflow error if its calculation result ends up lower than the value passed to `snprintf`.

Helps with https://github.com/containerd/overlaybd/issues/370, if https://github.com/data-accelerator/photon-libtcmu picks it up too.


liulanzheng:
@lxbsz please take a look

